PR #151, opened 2023-09-29
feat: Exhaustive greedy rewrite strategy

aborgna-q:
This is a straight combination of the greedy and exhaustive strategies. Instead of applying each strategy separately, apply each one once but also include all the following strategies that do not modify the same nodes.

This is a draft, since for some reason combining multiple strategies at once is invalidating the hugr.
They may not be correctly reporting their modified nodes ?


aborgna-q:
This requires https://github.com/CQCL/hugr/pull/600 to check whether the rewrites are non-overlapping.

We get _really_ nice results.

On my laptop, 1 thread, `barenco_tof_10`, 60s timeout, queue size 500:

Base cost: 192 CX gates, 326 gates.

- Using HEAD:
  - 
  
- With this branch:
  - Min cost: (120, 260)

lmondada (approved):
Excellent!

PR #174, opened 2023-10-06
feat: queue size argument for TASO

aborgna-q:
Adds a `-q`/`--queue-size` parameter, that defaults to 10_000.

Drive-by: Slight reformatting of the log output.


lmondada (approved):
Sorry it took me so long to get to your PRs, I've been distracted left and right all day and weekend. I'll look at the other ones now!

PR #175, opened 2023-10-06
feat: TASO worker sharding

### LINKED ISSUE #128: TASO multithreading idea: Sharding
@doug-q proposed an idea that may help with #118;

- Give each worker a local priority queue.
- Distribute circuit jobs based on their `hash % n_threads`. Each worker has a MPSC channel so we can send the work directly.

We'd probably have a separate channel for reporting new best circuit sizes to the main thread.

aborgna-q:
Closes #128.

Currently the performance is quite bad (We get 100% CPU utilisation, so it's not a contention problem).


src/optimiser/taso/log.rs
```diff
@@ -47,36 +49,44 @@ impl<'w> TasoLogger<'w> {
...
     pub fn log_progress(
         &mut self,
-        circ_cnt: usize,
+        circuits_processed: usize,
         workqueue_len: Option<usize>,
         seen_hashes: usize,
     ) {
-        if circ_cnt % 1000 == 0 {
-            self.progress(format!("{circ_cnt} circuits..."));
+        if circuits_processed > self.last_circ_processed && circuits_processed % 1000 == 0 {
```
lmondada, line 82: Can't you simplify this to `if circuits_processed > self.last_circ_processed + 1000`? Then we might also get away with less `log_counts` calls.

lmondada (commented):
Clean!

From what you've told me about performance, we could even consider doing pattern matching just in the main thread, and then sending those to be processed in parallel? Not quite sure, haven't thought it through.

And also still don't have a clear understanding of what exactly is taking long. You probably know better.

doug-q:
I did some profiling, building tosa-optimiser with in release with debuginfo, collecting samples with
```
 perf record  --call-graph=lbr -o n$n.data -F99 target/debug/taso-optimiser --input test_files/barenco_tof_10.json  --eccs test_files/h_rz_cxcomplete_ECC_set.json --n-threads=$n
```
n1
n10

PR #180, opened 2023-10-10
feat: Add Python TASO pass

lmondada:
This is a useful Python binding to create the pytket-hyper extension on top of.

### DIFF
--- a/pyrs/src/pass.rs
+++ b/pyrs/src/pass.rs
@@ -1,0 +1,146 @@
+use std::{cmp::min, convert::TryInto, fs, num::NonZeroUsize, path::PathBuf};
+
+use pyo3::{prelude::*, types::IntoPyDict};
+use tket2::{json::TKETDecode, op_matches, passes::apply_greedy_commutation, Circuit, T2Op};
+use tket_json_rs::circuit_json::SerialCircuit;
+
+use crate::{
+    circuit::{try_update_hugr, try_with_hugr},
+    optimiser::PyDefaultTasoOptimiser,
+};
+
+#[pyfunction]
+fn greedy_depth_reduce(py_c: PyObject) -> PyResult<(PyObject, u32)> {
+    try_with_hugr(py_c, |mut h| {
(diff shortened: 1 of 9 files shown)


pyrs/src/pass.rs
```diff
@@ -0,0 +1,147 @@
...
+/// Will use at most `max_threads` threads (plus a constant) and take at most
+/// `timeout` seconds (plus a constant). Default to the number of cpus and
+/// 30s respectively.
+///
+/// Log files will be written to the directory `log_dir` if specified.
+///
+/// This requires a `nam_6_3.rwr` file in the current directory. The location
+/// can alternatively be specified using the `rewriter_dir` argument.
+#[pyfunction]
+fn nam_taso_optimise(
```
aborgna-q, line 79: This has almost the same doc as `taso_optimise`. The fact that one rebases and the other doesn't should be clear there.

Perhaps just have a `rebase` optional parameter that defaults to `true`?

pyrs/src/pass.rs
```diff
@@ -0,0 +1,147 @@
...
+        ),
+        n if n >= 4 => (
+            vec![n, 2, 1],
+            vec![timeout / 2, timeout / 10 * 3, timeout / 10 * 2],
+        ),
+        n if n > 1 => (vec![n, 1], vec![timeout / 2, timeout / 2]),
+        1 => (vec![1], vec![timeout]),
+        _ => unreachable!(),
+    };
+    let optimiser = PyDefaultTasoOptimiser::load_precompiled(rewrite_dir.join("nam_6_3.rwr"));
```
aborgna-q, line 105: The hardcoded file is sketchy... add a TODO to remove it for now.

pyrs/src/pass.rs
lmondada: Yes, I've removed the `nam_` variant (I had originially thought I would write some of this in Python, but chose to have it all in Rust in the end).

PR #182, opened 2023-10-10
feat: Multithreading improvements

aborgna-q:
The main feature of this PR is reducing the memory usage of multithreading with a priority channel by adding a shared `Arc<RwLock<Option<P>>>` with the maximum circuit cost in the queue, so the workers don't fill the channels with useless data.

The main improvement of this is limiting the memory usage when multiple threads are involved. Now I can run `barenco_tof_10` on 10 threads with ~4GB of RAM, where before it exceeded 10GB (and caused a OOM).

I also included multiple small changes extracted from #175;

- `CircuitCost::as_usize` instead of proxy operations, so we can share cost limits as atomics and run other primitive operations directly.
- Print the actual number of circuits processed on the log (not only the seen count, which includes elements in the queue).
- Replace `eprintln!`s with log calls.
- Print the max queue capacity at the beginning of the execution.

### DIFF
--- a/src/optimiser/taso/hugr_pchannel.rs
+++ b/src/optimiser/taso/hugr_pchannel.rs
@@ -14,7 +14,7 @@ use super::hugr_pqueue::{Entry, HugrPQ};
 
 /// A unit of work for a worker, consisting of a circuit to process, along its
 /// hash and cost.
-pub type Work<P> = (P, u64, Hugr);
+pub type Work<P> = Entry<Hugr, P, u64>;
 
 /// A priority channel for HUGRs.
 ///
@@ -88,7 +88,10 @@ impl<P: CircuitCost> PriorityChannelCommunication<P> {
         //
         match self.max_cost() {
             Some(max_cost) => {
-                let filtered = work.into_iter().filter(|(c, _, _)| c < &max_cost).collect();
+                let filtered = work
+                    .into_iter()
+                    .filter(|Work { cost, .. }| cost < &max_cost)
+                    .collect();
                 self.push.send(filtered)?;
             }
             _ => self.push.send(work)?,
@@ -219,7 +222,7 @@ where
                             }
                             self.enqueue_circs(new_circs);
                         }
-                        send(self.pop, {let Entry {cost, hash, circ} = self.pq.pop().unwrap(); (cost, hash, circ)}) -> result => {
+                        send(self.pop, self.pq.pop().unwrap()) -> result => {
                             if result.is_err() {
                                 // Something went wrong.
                                 break 'main;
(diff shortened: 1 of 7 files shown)


src/optimiser/taso/hugr_pchannel.rs
```diff
@@ -188,19 +254,35 @@ where
...
-            }
+        // This is the result from processing a circuit. Add it to the count.
+        self.circ_cnt += 1;
+        self.log
+            .send(PriorityChannelLog::CircuitCount {
+                processed_count: self.circ_cnt,
+                seen_count: self.seen_hashes.len(),
+                queue_length: self.pq.len(),
+            })
+            .unwrap();
```
lmondada, line 269: Is it on purpose that you are sending a log at every `circ_cnt` increment?

src/optimiser/taso/hugr_pqueue.rs
```diff
@@ -10,23 +10,22 @@ use crate::circuit::CircuitHash;
...
-pub(super) struct Entry<C, P, H> {
-    pub(super) circ: C,
-    pub(super) cost: P,
-    #[allow(unused)] // TODO remove?
-    pub(super) hash: H,
+pub struct Entry<C, P, H> {
+    pub circ: C,
+    pub cost: P,
+    pub hash: H,
 }
```
lmondada, line 24: Could/should this type be the same as `type Work` in `hugr_pchannel`?

src/optimiser/taso/log.rs
```diff
@@ -47,36 +49,44 @@ impl<'w> TasoLogger<'w> {
...
     pub fn log_progress(
         &mut self,
-        circ_cnt: usize,
+        circuits_processed: usize,
         workqueue_len: Option<usize>,
         seen_hashes: usize,
     ) {
-        if circ_cnt % 1000 == 0 {
-            self.progress(format!("{circ_cnt} circuits..."));
+        if circuits_processed > self.last_circ_processed && circuits_processed % 1000 == 0 {
```
lmondada, line 82: I think this can be `if circuits_processed >= self.last_circ_processed + 1000`

src/optimiser/taso/worker.rs
```diff
@@ -4,64 +4,93 @@ use std::thread::{self, JoinHandle};
...
+use crate::circuit::cost::CircuitCost;
 use crate::circuit::CircuitHash;
 use crate::rewrite::strategy::RewriteStrategy;
 use crate::rewrite::Rewriter;
 
+use super::hugr_pchannel::PriorityChannelCommunication;
+
+/// A unit of work for a worker, consisting of a circuit to process, along its
+/// hash and cost.
+pub type Work<P> = (P, u64, Hugr);
```
lmondada, line 16: You've defined this twice.

lmondada (approved):

PR #184, opened 2023-10-10
feat: Use cost deltas instead of recalculating the full cost

aborgna-q:
Avoids doing full circuit traversals for each candidate.

### DIFF
--- a/src/rewrite/strategy.rs
+++ b/src/rewrite/strategy.rs
@@ -127,7 +127,7 @@ pub trait ExhaustiveThresholdStrategy {
     /// Returns true if the rewrite is allowed, based on the cost of the pattern and target.
     #[inline]
     fn under_threshold(&self, pattern_cost: &Self::OpCost, target_cost: &Self::OpCost) -> bool {
-        target_cost.sub_cost(pattern_cost) <= 0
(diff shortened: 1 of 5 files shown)


src/rewrite/strategy.rs
```diff
@@ -53,6 +53,50 @@ pub trait RewriteStrategy {
...
 
+/// The result of a rewrite strategy.
+///
+/// Returned by [`RewriteStrategy::apply_rewrites`].
+pub struct RewriteResult<Cost: CircuitCost> {
+    /// The rewritten circuits.
+    pub circs: Vec<Hugr>,
+    /// The cost delta of each rewritten circuit.
+    pub cost_deltas: Vec<Cost::CostDelta>,
+}
```
lmondada, line 64: What's the advantage of this over `type RewriteResult<Cost: CircuitCost> = Vec<(Hugr, Cost::CostDelta)>`?

You'd get all of the methods below for free. Otherwise keep it this way and use `delegate!`.

lmondada: I guess it makes testing a bit easier than having to play with indices.

aborgna-q: Yeah, specially since the fields are public it's nicer if you can name them.

PR #187, opened 2023-10-12
chore: Support any pytket >= 1.0

lmondada:
I don't think there is a reason to only support pytket 1.0. I was getting some
conflicts with pytket extensions based on this.


lmondada:
Sorry, I'm stupid, this doesn't change anything obviously.

PR #193, opened 2023-10-19
feat: drop tkcxx

### LINKED ISSUE #189: Drop `tkcxx` feature.
Currently we don't use `tket_rs` at any point in the code, but the `tkcxx` feature still compiles the library when enabled.

This is problematic, since to successfully use the feature we require a correctly setup conan configuration; following all the instructions in the `tket-rs` repo, installing an old version of gcc (11)...

Even if we do not use the feature, `rust-analyser` will try to compile with it during local development (and it will panic if something's not fully setup).

I say we should drop the feature altogether, and potentially include it as a separate package later on.

aborgna-q:
It wasn't used in the code, but required the dev environment to have conan set up.

drive-by: Unify the `tket-json-rs` dependency in al the Cargo.tomls.

Closes #189


ss2165 (approved):
🔥

PR #197, opened 2023-10-20
feat: Do not require &mut ConvexCheckers

aborgna-q:
The latest hugr update dropped the mut requirement in `SiblingSubgraph`.


acl-cqc (approved):
Yes! Thanks :-).

Feels like we should allow `clippy` to be a reviewer here :)

PR #198, opened 2023-10-24
perf: Use jemallocator by default for taso

aborgna-q:
Also enables thin LTO on release builds.

Tested on `barenco_tof_10`, 10s timeout, single threaded, on my laptop.

Before / after:

### DIFF
--- a/taso-optimiser/Cargo.toml
+++ b/taso-optimiser/Cargo.toml
@@ -17,7 +17,7 @@ tracing-subscriber = "0.3.17"
 tracing-appender = "0.2.2"
 peak_alloc = { version = "0.2.0", optional = true }
 
-[target.'cfg(all(not(target_env = "msvc"), not(feature = "peak_alloc"))'.dependencies]
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
 tikv-jemallocator = "0.5"
 
 [features]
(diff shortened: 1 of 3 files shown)


taso-optimiser/Cargo.toml
```diff
@@ -17,6 +17,9 @@ tracing-subscriber = "0.3.17"
 tracing-appender = "0.2.2"
 peak_alloc = { version = "0.2.0", optional = true }
 
+[target.'cfg(not(target_env = "msvc"))'.dependencies]
```
doug-q, line 20:
```suggestion
[target.'cfg(all(not(target_env = "msvc"), not(feature = "peak_alloc"))'.dependencies]
```

doug-q (approved):
Nice. Should we also offer a configuration which uses the default global allocator?

taso-optimiser/Cargo.toml
aborgna-q:
```
> warning: /home/aborgna/src/tket2/taso-optimiser/Cargo.toml:
  Found `feature = ...` in `target.'cfg(...)'.dependencies`.
  This key is not supported for selecting dependencies and will not work as expected.
  Use the [features] section instead:
  https://doc.rust-lang.org/cargo/reference/features.html
```

aborgna-q:
I think defaulting to `jemalloc` when possible (so, outside windows) is generally better.
Note that this is only used for the `taso-optimiser` binary target, crate users are free to chose their own allocator.

PR #202, opened 2023-10-27
feat: Simplify the circ hash code. Allow nodes with children.

aborgna-q:
Before this change, we stored the hash for each port in a map. We could then drop it when consumed if the port was linear.
This wasn't really necessary; the memory usage reduction vs just storing all the node hashes is quite small in comparison with the circuit size, and the extra logic made things slower.

This PR replaces that logic with a straight "compute each node hash and store it". This is more robust in cases that have non-local edges.

I also added a recursive call to support hashing dataflow container nodes.

### DIFF
--- a/src/circuit/hash.rs
+++ b/src/circuit/hash.rs
@@ -1,12 +1,11 @@
 //! Circuit hashing.
 
-use core::panic;
 use std::hash::{Hash, Hasher};
 
 use fxhash::{FxHashMap, FxHasher64};
-use hugr::ops::{LeafOp, OpName, OpTag, OpTrait, OpType};
-use hugr::types::TypeBound;
-use hugr::{HugrView, Node, Port};
+use hugr::hugr::views::{HierarchyView, SiblingGraph};
+use hugr::ops::{LeafOp, OpName, OpType};
+use hugr::{HugrView, Node};
 use petgraph::visit::{self as pg, Walker};
 
 use super::Circuit;
@@ -38,10 +37,12 @@ where
             .filter(|&n| n != self.root())
         {
             let hash = hash_node(self, node, &mut hash_state);
-            hash_state.set_node(self, node, hash);
+            hash_state.set_hash(node, hash);
         }
 
-        hash_state.get_nonlinear(self.output())
+        hash_state
+            .node_hash(self.output())
+            .expect("Output hash has not been set")
     }
 }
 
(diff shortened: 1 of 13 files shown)


src/circuit/hash.rs
```diff
@@ -50,61 +51,25 @@ where
...
-                    self.linear_hashes.insert((node, port), hash);
-                }
-            }
-        }
-        if any_nonlinear || optype.tag() <= OpTag::Output {
-            self.nonlinear_hashes.insert(node, hash);
-        }
+    #[inline]
+    fn set_hash(&mut self, node: Node, hash: u64) {
+        self.hashes.insert(node, hash);
```
acl-cqc, line 68: Might want to check the value returned from `insert` was `None`

src/circuit/hash.rs
```diff
@@ -130,9 +95,14 @@ fn hash_node(circ: &impl HugrView, node: Node, state: &mut HashState) -> u64 {
     let op = circ.get_optype(node);
     let mut hasher = FxHasher64::default();
 
+    // Hash the node children
     if circ.children(node).count() > 0 {
-        panic!("Cannot hash container node {node:?}.");
+        let container: SiblingGraph =
+            SiblingGraph::try_new(circ, node).expect("Cannot hash container node");
```
acl-cqc, line 101: AFAICS SiblingGraph does not insist its root node is a container; if its just a NodeHandle for a non-container, you'll get back a SiblingGraph with just one node (the root). So IIUC the only case where `try_new` will return a failure here is if `node` is bad, so you might tweak this message (or just unwrap!)

src/circuit/hash.rs
```diff
@@ -130,9 +95,14 @@ fn hash_node(circ: &impl HugrView, node: Node, state: &mut HashState) -> u64 {
...
+        container.circuit_hash().hash(&mut hasher);
```
acl-cqc, line 102: Just in passing but I have to say - wtf, the method is `u64.hash(&mut hasher)` not `FxHasher64.hash(u64)` ?!

src/circuit/hash.rs
```diff
@@ -38,10 +37,12 @@ where
             .filter(|&n| n != self.root())
         {
             let hash = hash_node(self, node, &mut hash_state);
-            hash_state.set_node(self, node, hash);
+            hash_state.set_hash(node, hash);
```
acl-cqc, line 40: Can I make a suggestion here? Just rename `hash_state` to `node_hashes`. Clarify that there is no state there *apart from* the node hashes, which is critical in ensuring the behaviour of the hash function is not affected by e.g. the particular topsort order that we get

acl-cqc (approved):
Nice! :) Hashing the container is easy, and I totally agree that the linear/non-linear is not carrying its (substantial!) weight

PR #206, opened 2023-11-01
chore: Fix devenv for Linux

cqc-alec:


ss2165 (approved):
thanks

cqc-alec:
> thanks
> 
> though maybe change the commit prefix to `infra`, we don't want this in the changelog

Don't think `infra` is allowed; could use `ci` or `chore`.

PR #210, opened 2023-11-02
refactor: Move tket2 code to a workspace member

aborgna-q:
Moves the `tket2` crate code to `./tket2/`, and adds it to the workspace. This is mostly for cleaning up, as part of the python module cleanup task.

As a drive-by, makes sure that all the `Cargo.toml`s shale the workspace metadata.


tket2/src/optimiser/taso/qtz_circuit.rs
```diff
@@ -145,7 +145,7 @@ mod tests {
     #[cfg_attr(miri, ignore)] // Opening files is not supported in (isolated) miri
     fn test_read_rep() {
         let rep_map: HashMap<String, Circuit> =
-            load_representative_set("test_files/h_rz_cxrepresentative_set.json");
+            load_representative_set("../test_files/h_rz_cxrepresentative_set.json");
```
ss2165, line 148: can we set some kind of workspace environment variable or something to make this less flaky?

tket2/tests/taso_termination.rs
```diff
@@ -0,0 +1,62 @@
+#![cfg(feature = "portmatching")]
```
ss2165, line 1: for some reason github didn't recognize this file move

ss2165 (approved):

tket2/src/optimiser/taso/qtz_circuit.rs
aborgna-q: That will be a bit noisy, so I'll open an issue for that.

PR #212, opened 2023-11-03
fix: Make native py modules behave like python's

### LINKED ISSUE #209: `from pyrs.pyrs.optimiser import TasoOptimiser` doesn't work.
It requires importing optimiser first.

aborgna-q:
Cleans up the `tket2-py` structure and makes the modules behave like normal python ones.

With this we can do
```python
from tket2.circuit import T2Circuit
```

Minor changes:

- Moves some code from `tket2-py/src/lib.rs` to a new `pattern.rs`.
- Deletes some commented-out code.

Closes #209


ss2165 (approved):
unfortunately noisy diff but looks good

PR #219, opened 2023-11-07
chore: update hugr dependency

aborgna-q:
This required a partial move from using `Port` everywhere to the more specific `OutgoingPort` and `IncomingPort`.

I'll open a separate issue for the more pervasive changes like updating the `Units` iterator and the `portmatching` module.
For now, we just cast and `unwrap` in those cases.

Requires https://github.com/CQCL/hugr/pull/647

### DIFF
--- a/tket2/src/circuit.rs
+++ b/tket2/src/circuit.rs
@@ -6,8 +6,8 @@ mod hash;
 pub mod units;
 
 pub use command::{Command, CommandIterator};
-use either::Either::{Left, Right};
 pub use hash::CircuitHash;
+use itertools::Either::{Left, Right};
 
 use derive_more::From;
 use hugr::hugr::hugrmut::HugrMut;
--- a/tket2/src/circuit.rs
+++ b/tket2/src/circuit.rs
@@ -6,6 +6,7 @@ mod hash;
 pub mod units;
 
 pub use command::{Command, CommandIterator};
+use either::Either::{Left, Right};
 pub use hash::CircuitHash;
 
 use derive_more::From;
@@ -174,7 +175,7 @@ pub(crate) fn remove_empty_wire(
     if input_port >= circ.num_outputs(inp) {
         return Err(CircuitMutError::InvalidPortOffset(input_port));
     }
-    let input_port = Port::new_outgoing(input_port);
+    let input_port = Port::new(Direction::Outgoing, input_port);
     let link = circ
         .linked_ports(inp, input_port)
         .at_most_one()
@@ -234,9 +235,15 @@ fn shift_ports<C: HugrMut + ?Sized>(
         }
         for (other_n, other_p) in links {
             // TODO: simplify when CQCL-DEV/hugr#565 is resolved
-            match dir {
-                Direction::Incoming => circ.connect(other_n, other_p, node, free_port),
-                Direction::Outgoing => circ.connect(node, free_port, other_n, other_p),
+            match other_p.as_directed() {
+                Right(other_p) => {
+                    let dst_port = free_port.as_incoming().unwrap();
+                    circ.connect(other_n, other_p, node, dst_port)
+                }
+                Left(other_p) => {
+                    let src_port = free_port.as_outgoing().unwrap();
+                    circ.connect(node, src_port, other_n, other_p)
+                }
             }?;
         }
         free_port = port;
--- a/tket2/src/circuit.rs
+++ b/tket2/src/circuit.rs
@@ -234,7 +234,6 @@ fn shift_ports<C: HugrMut + ?Sized>(
             circ.disconnect(node, port)?;
         }
         for (other_n, other_p) in links {
-            // TODO: simplify when CQCL-DEV/hugr#565 is resolved
             match other_p.as_directed() {
                 Right(other_p) => {
                     let dst_port = free_port.as_incoming().unwrap();
--- a/tket2/src/passes/chunks.rs
+++ b/tket2/src/passes/chunks.rs
@@ -223,13 +223,11 @@ struct ChunkInsertResult {
 }
 
 /// The target of a chunk connection in a reassembled circuit.
-#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, From)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 enum ConnectionTarget {
     /// The target is a chunk node's input.
-    #[from]
     InsertedInput(Node, IncomingPort),
     /// The target is a chunk node's output.
-    #[from]
     InsertedOutput(Node, OutgoingPort),
     /// The link goes directly to the opposite boundary, without an intermediary
     /// node.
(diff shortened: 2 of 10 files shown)


tket2/src/passes/chunks.rs
```diff
@@ -155,7 +155,7 @@ impl Chunk {
                 ConnectionTarget::TransitiveConnection(input_connection)
             } else {
                 // Translate the original chunk node into the inserted node.
-                (*node_map.get(&node).unwrap(), port).into()
+                ConnectionTarget::InsertedOutput(*node_map.get(&node).unwrap(), port)
```
acl-cqc, line 158: You could leave this line unchanged, right, as you are `#[derive(...,From)]` on ConnectionTarget and `#[from]` on Inserted(In,Out)put? I'd either use the From derivations or not derive them

tket2/src/passes/chunks.rs
```diff
@@ -225,9 +225,12 @@ struct ChunkInsertResult {
 /// The target of a chunk connection in a reassembled circuit.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, From)]
 enum ConnectionTarget {
-    /// The target is a single node and port.
+    /// The target is a chunk node's input.
```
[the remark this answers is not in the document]
acl-cqc: (later you can change `incoming_connections` and `outgoing_connections` to have specific direction, by parameterization, avoiding the panic, but that can be another PR, sure...)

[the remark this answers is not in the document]
acl-cqc: Hmmmm, given that, are we sure that it doesn't make more sense to keep this as `Port` and then do `as_incoming().unwrap()`/`as_outgoing().unwrap()` in the two places instead of the `panic`s you've added?

tket2/src/portmatching/pattern.rs
```diff
@@ -42,17 +43,17 @@ impl CircuitPattern {
             let op = cmd.optype().clone();
             pattern.require(cmd.node().into(), op.try_into().unwrap());
             for in_offset in 0..cmd.input_count() {
-                let in_offset = Port::new_incoming(in_offset);
-                let edge_prop =
-                    PEdge::try_from_port(cmd.node(), in_offset, circuit).expect("Invalid HUGR");
+                let in_offset: IncomingPort = in_offset.into();
+                let edge_prop = PEdge::try_from_port(cmd.node(), in_offset.into(), circuit)
```
acl-cqc, line 47: I think you can do this more directly (avoiding `in_offset`) via `Port::new_incoming`

tket2/src/portmatching/pattern.rs
```diff
@@ -235,8 +236,14 @@ mod tests {
             edges,
             [
                 (cx_gate, h_gate),
-                (cx_gate, NodeID::CopyNode(inp, Port::new_outgoing(0))),
-                (cx_gate, NodeID::CopyNode(inp, Port::new_outgoing(1))),
+                (
+                    cx_gate,
+                    NodeID::CopyNode(inp, Port::new(Direction::Outgoing, 0))
```
acl-cqc, line 241:
```suggestion
                    NodeID::CopyNode(inp, Port::new_outgoing(0))
```

tket2/src/portmatching/pattern.rs
```diff
@@ -235,8 +236,14 @@ mod tests {
...
+                ),
+                (
+                    cx_gate,
+                    NodeID::CopyNode(inp, Port::new(Direction::Outgoing, 1))
```
acl-cqc, line 245:
```suggestion
                    NodeID::CopyNode(inp, Port::new_outgoing(1))
```

tket2/src/portmatching/pattern.rs
```diff
@@ -292,7 +299,7 @@ mod tests {
             assert!(edges.iter().any(|e| {
                 e.reverse().is_none()
                     && e.source.unwrap() == rx_n.into()
-                    && e.target.unwrap() == NodeID::CopyNode(inp, Port::new_outgoing(1))
+                    && e.target.unwrap() == NodeID::CopyNode(inp, Port::new(Direction::Outgoing, 1))
```
acl-cqc, line 302: similarly

tket2/src/portmatching/pattern.rs
aborgna-q: That doesn't exist anymore.

tket2/src/circuit.rs
```diff
@@ -234,9 +235,15 @@ fn shift_ports<C: HugrMut + ?Sized>(
         }
         for (other_n, other_p) in links {
             // TODO: simplify when CQCL-DEV/hugr#565 is resolved
-            match dir {
-                Direction::Incoming => circ.connect(other_n, other_p, node, free_port),
-                Direction::Outgoing => circ.connect(node, free_port, other_n, other_p),
+            match other_p.as_directed() {
```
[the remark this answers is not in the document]
aborgna-q: We resolved to keep the output, then input order so this comment can be dropped.
(that implemented the first version of incoming/outgoing ports).

tket2/src/passes/chunks.rs
aborgna-q: dropped the From, the explicit version is more clear

tket2/src/passes/chunks.rs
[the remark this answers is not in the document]
aborgna-q: Having the explicit variants makes it so any direction bug is detected at construction time, rather than when unwrapping. The panic will only be raised for bugs in the intermediate code, so it's a bit more scoped.

Hopefully this will be cleaned out in #220.

tket2/src/portmatching/matcher.rs
```diff
@@ -129,12 +129,16 @@ impl PatternMatch {
         let inputs = pattern_ref
             .inputs
             .iter()
-            .map(|p| p.iter().map(|(n, p)| (map[n], *p)).collect_vec())
+            .map(|ps| {
+                ps.iter()
+                    .map(|(n, p)| (map[n], p.as_incoming().unwrap()))
```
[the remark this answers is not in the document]
aborgna-q: That's a can of worms (I tried :P). Outgoing/Incoming ports don't implement Serialize, so it's not a drop-in replacement.

That's one of the items in #220.

tket2/src/portmatching/pattern.rs
aborgna-q: We use `in_offset` below for `linked_outputs`, so we need the Incoming port.

tket2/src/portmatching/pattern.rs
acl-cqc: Yes, I just found that myself, sorry. `OutgoingPort::from(0).into()`, hmmm. Add `NodeID::new_copy(Node, impl Into<Port>)` ?

tket2/src/passes/chunks.rs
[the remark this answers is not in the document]
acl-cqc: Yah ok. As an intermediate step this is indeed slightly better I think. I was only thinking that we'll be kind-of reverting this in #220 but of course, it depends how long it takes us to get round to that P-low. So, fine :)

tket2/src/portmatching/matcher.rs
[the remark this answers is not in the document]
acl-cqc: Hah, boo. 😢 but ok

acl-cqc (approved):
LGTM, thanks @aborgna-q !

PR #228, opened 2023-11-08
refactor: s/taso/badger/

### LINKED ISSUE #224: Rename TASO -> Badger
I am finally taking the time to read the original "TASO" paper.

https://dl.acm.org/doi/10.1145/3341301.3359630

Turns out TASO stands for "Tensor Algebra SuperOptimiser" and doesn't refer to the optimiser, but the whole idea of generating rewrite rules automatically, in the context of NN computational graphs... The optimiser itself is refered to as a "cost-based backtracking search", which is a pretty descriptive term that we could add to the documentation.

So it's probably a good idea to get rid of the TASO name. I'd like to suggest the more fun name of "badger" optimiser. Coz in Italian "tasso" means a badger.

aborgna-q:
Closes #224


lmondada (approved):
Long live Badger the Backtracker!

tket2/src/optimiser/badger.rs
```diff
@@ -32,15 +32,15 @@ use hugr::Hugr;
...
+use crate::optimiser::badger::hugr_pqueue::{Entry, HugrPQ};
+use crate::optimiser::badger::worker::BadgerWorker;
 use crate::passes::CircuitChunks;
 use crate::rewrite::strategy::RewriteStrategy;
 use crate::rewrite::Rewriter;
 use crate::Circuit;
 
-/// The TASO optimiser.
+/// The Badger optimiser.
 ///
```
lmondada, line 44:
```suggestion
///
/// A cost-based backtracking search for rewrite-base circuit optimisation.
///
```

PR #231, opened 2023-11-09
feat: add angle type to tket2 extension 

### LINKED ISSUE #217: Port `Angle` types from hugr repo

ss2165:
moved from hugr https://github.com/CQCL/hugr/pull/670

Closes #217

### DIFF
--- a/tket2/src/extension/angle.rs
+++ b/tket2/src/extension/angle.rs
@@ -1,10 +1,11 @@
 use std::{cmp::max, num::NonZeroU64};
 
 use hugr::{
-    extension::{prelude::ERROR_TYPE, SignatureError},
+    extension::{prelude::ERROR_TYPE, ExtensionRegistry, SignatureError, TypeDef, PRELUDE},
     types::{
         type_param::{TypeArgError, TypeParam},
-        ConstTypeError, CustomCheckFailure, CustomType, FunctionType, Type, TypeArg, TypeBound,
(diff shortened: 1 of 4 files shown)


cqc-alec (approved):

tket2/src/extension/angle.rs
```diff
@@ -136,49 +126,51 @@ impl CustomConst for ConstAngle {
...
 
-fn aunop_sig(arg_values: &[TypeArg]) -> Result<FunctionType, SignatureError> {
-    let [arg] = collect_array(arg_values);
-    Ok(FunctionType::new_linear(vec![angle_type(arg.clone())]))
+fn aunop_sig(extension: &Extension) -> Result<FunctionType, SignatureError> {
+    let angle = type_var(0, extension)?;
+    Ok(FunctionType::new_linear(vec![angle]))
+}
+
+fn angle_def(extension: &Extension) -> &TypeDef {
```
cqc-alec, line 172: Maybe this is a wider question about the extensions API, but it seems strange that this function (and some others -- `type_var`, `angle_custom_type`, `atrunc_sig` etc) need to take their own extension as an argument. I suppose it is because `angle_custom_type` is exposed at the module level? But it feels like it should be unnecessary.

cqc-alec: OK, maybe something to ponder but this LGTM.

ss2165: I have an issue to raise already about needing a temporary register containing types in order to be able to create validated type schemes, perhaps we can rethink the definition API as part of that.

Issue #204, opened 2023-10-27
Tracking issue: Python bindings UX


aborgna-q:
This is a tracking issue for improvements to the python API.

## Documentation & repo structure

- [x] CQCL/tket2#207
- [x] CQCL/tket2#169
- [x] CQCL/tket2#229

## Issues

- [x] CQCL/tket2#208
  E.g. `from pyrs.pyrs import optimiser`.
- [x] CQCL/tket2#209
  It requires importing `optimiser` first.

### DAG queries

- [ ] CQCL/tket2#226
- [ ] CQCL/tket2#225
- [x] CQCL/tket2#227
- [x] CQCL/tket2#178

### Matching and rewriting
- [ ] Expose rewrite primitives (SimpleRewrite, Inline*)
- [ ] Rewrite combinators / rewrite strategies.
- [x] ~TASO~Badger as a `CustomPass`. CQCL/tket2#185 

### Construction

- [x] CQCL/hugr#946 
- [x] CQCL/tket2#303 
- [ ] Control flow builder

Issue #178, opened 2023-10-09
Use tket1 circuit and Hugr interchangeably 


ss2165:
Having this is nice for avoiding unnecessary roundtrips, but it'd be nice to be able to use this interchangeably with TKET1 circs. Open an issue to investigate options?

_Originally posted by @aborgna-q in https://github.com/CQCL-DEV/tket2/pull/177#discussion_r1350269565_

PR #235, opened 2023-11-10
refactor: Move python bindings from `tket2` to `tket2-py`

### LINKED ISSUE #148: Move src/portmatching/pyo3.rs to pyrs
There is no reason this should live within `tket2`.

aborgna-q:
Adds transparent wrappers on `tket2-py` that derive pyclass themselves.
This has multiple goals:

- Cleans up the code. Python stuff goes into the python crate.
  Now `tket2:pyo3` and `hugr:pyo3` are only used for the python error implementations.
  (We could potentially move this too, and avoid us the `pyo3` dependency headache).
  
- Let's us use the `tket2-py` conversion methods for tket1 and tket2 circuits. This is required for #178, which is in turn required for all the python bindings in #204 that require stable indices.

Closes #148 

---

The code is mostly the same, asides from:
- replacing the custom implementation of `PyPatternMatch` with a transparent wrapper. (There was a TODO for this).
- updating to tket-json-rs 0.3 (cleans up some calls that use `Python::with_gil`)


ss2165 (approved):
nice :)

PR #237, opened 2023-11-10
test: Add small parallel badger test

aborgna-q:
Timeout=0, so runtime impact is minimal, and it's kept deterministic; the loop quits immediately and no rewrites are applied.


ss2165 (approved):
Looks like the test is a copy of the one above with a couple of parameters changed so could be better with either a loop or even better rstest parametrisation...

PR #243, opened 2023-11-14
feat: Use tket1 and tket2 circuits interchangeably everywhere

### LINKED ISSUE #178: Use tket1 circuit and Hugr interchangeably

aborgna-q:
Adds a `tket1`/`tket2` to the `with_hugr` helpers, so we always know what format to output afterwards.

The more noisy part of this commit changing all the GIL-independent types to capturing references, so we don't have to manually lock multiple times per call.

Closes #178.

### DIFF
--- a/tket2-py/src/circuit.rs
+++ b/tket2-py/src/circuit.rs
@@ -13,20 +13,18 @@ use tket2::json::TKETDecode;
 use tket2::rewrite::CircuitRewrite;
 use tket_json_rs::circuit_json::SerialCircuit;
 
-pub use self::convert::{try_update_hugr, try_with_hugr, update_hugr, with_hugr, T2Circuit};
+pub use self::convert::{try_update_hugr, try_with_hugr, update_hugr, with_hugr, Tk2Circuit};
 
 /// The module definition
 pub fn module(py: Python) -> PyResult<&PyModule> {
     let m = PyModule::new(py, "_circuit")?;
-    m.add_class::<T2Circuit>()?;
+    m.add_class::<Tk2Circuit>()?;
     m.add_class::<PyNode>()?;
     m.add_class::<tket2::T2Op>()?;
     m.add_class::<tket2::Pauli>()?;
 
     m.add_function(wrap_pyfunction!(validate_hugr, m)?)?;
     m.add_function(wrap_pyfunction!(to_hugr_dot, m)?)?;
-    m.add_function(wrap_pyfunction!(tket1_to_tket2, m)?)?;
-    m.add_function(wrap_pyfunction!(tket2_to_tket1, m)?)?;
 
     m.add("HugrError", py.get_type::<hugr::hugr::PyHugrError>())?;
     m.add("BuildError", py.get_type::<hugr::builder::PyBuildError>())?;
@@ -56,18 +54,6 @@ pub fn validate_hugr(c: &PyAny) -> PyResult<()> {
 #[pyfunction]
 pub fn to_hugr_dot(c: &PyAny) -> PyResult<String> {
     with_hugr(c, |hugr, _| hugr.dot_string())
-}
-
-/// Cast a python tket1 circuit to a [`T2Circuit`].
-#[pyfunction]
-pub fn tket1_to_tket2(c: &PyAny) -> PyResult<T2Circuit> {
-    T2Circuit::from_circuit(c)
-}
-
-/// Cast a [`T2Circuit`] to a python tket1 circuit.
-#[pyfunction]
-pub fn tket2_to_tket1(py: Python, c: T2Circuit) -> PyResult<&PyAny> {
-    c.finish(py)
 }
 
 /// A [`hugr::Node`] wrapper for Python.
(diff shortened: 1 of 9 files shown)


tket2-py/src/circuit.rs
```diff
@@ -47,20 +48,26 @@ pub fn module(py: Python) -> PyResult<&PyModule> {
...
 /// Return a Graphviz DOT string representation of the circuit.
 #[pyfunction]
-pub fn to_hugr_dot(c: Py<PyAny>) -> PyResult<String> {
-    with_hugr(c, |hugr| hugr.dot_string())
+pub fn to_hugr_dot(c: &PyAny) -> PyResult<String> {
+    with_hugr(c, |hugr, _| hugr.dot_string())
 }
 
-/// Downcast a python object to a [`Hugr`].
+/// Cast a python tket1 circuit to a [`T2Circuit`].
```
ss2165, line 61: "cast" implies "no copy" to me..?

tket2-py/src/circuit/convert.rs
```diff
@@ -21,71 +23,130 @@ pub struct T2Circuit {
...
 
-    fn apply_match(&mut self, rw: PyCircuitRewrite) {
+    /// Apply a rewrite on the circuit.
+    pub fn apply_match(&mut self, rw: PyCircuitRewrite) {
         rw.rewrite.apply(&mut self.hugr).expect("Apply error.");
     }
+
+    /// Encode the circuit as a HUGR json string.
+    pub fn to_hugr_json(&self) -> PyResult<String> {
+        Ok(serde_json::to_string(&self.hugr).unwrap())
```
ss2165, line 46: a note that previously we have used messagepack (and the behaviour between json and messagepack is not always identical), but I'm ok avoiding the extra python dependency here for now - at least until the serialized format gets a bit more stable (hopefully soon)

tket2-py/src/circuit/convert.rs
```diff
@@ -21,71 +23,130 @@ pub struct T2Circuit {
...
+    /// Converts a `Hugr` into the format indicated by the flag.
+    pub fn convert(self, py: Python, hugr: Hugr) -> PyResult<&PyAny> {
+        match self {
+            CircuitType::Tket1 => SerialCircuit::encode(&hugr)?.to_tket1(py),
+            CircuitType::Tket2 => Ok(Py::new(py, T2Circuit { hugr })?.into_ref(py)),
+        }
     }
 }
 
 /// Apply a fallible function expecting a hugr on a pytket circuit.
```
ss2165, line 103: docstring is out of date?

tket2-py/src/circuit/convert.rs
```diff
@@ -21,71 +23,130 @@ pub struct T2Circuit {
...
+pub fn with_hugr<T, F>(circ: &PyAny, f: F) -> PyResult<T>
 where
-    F: FnOnce(Hugr) -> T,
+    F: FnOnce(Hugr, CircuitType) -> T,
 {
-    try_with_hugr(circ, |hugr| Ok::<T, PyErr>((f)(hugr)))
+    try_with_hugr(circ, |hugr, typ| Ok::<T, PyErr>((f)(hugr, typ)))
 }
 
 /// Apply a hugr-to-hugr function on a pytket circuit, and return the modified circuit.
```
ss2165, line 129: docstring

tket2-py/src/circuit.rs
```diff
@@ -47,20 +48,26 @@ pub fn module(py: Python) -> PyResult<&PyModule> {
...
 #[pyfunction]
-pub fn to_hugr(c: Py<PyAny>) -> PyResult<T2Circuit> {
-    with_hugr(c, |hugr| hugr.into())
+pub fn tket1_to_tket2(c: &PyAny) -> PyResult<T2Circuit> {
```
ss2165, line 63: these function names are very easy to confuse, in python I think `obj.to_tket1()` and `obj.to_tket2()` might be nicer?

aborgna-q: We cant add a `tket1.Circuit.to_tket2`, but I'll drop these and leave
```rust
Tk2Circuit(&tket1) -> Self
Tk2Circuit.to_tket1(&self) -> tket1
```

tket2-py/src/circuit/convert.rs
aborgna-q: Added a TODO

tket2-py/src/circuit.rs
ss2165: well...it's python so you probably can add it with some horrible hacking. But i'm happier with this.

tket2-py/src/circuit/convert.rs
```diff
@@ -14,78 +16,149 @@ use crate::pattern::rewrite::PyCircuitRewrite;
...
-pub struct T2Circuit {
+pub struct Tk2Circuit {
     /// Rust representation of the circuit.
     pub hugr: Hugr,
 }
 
 #[pymethods]
-impl T2Circuit {
+impl Tk2Circuit {
+    /// Cast a tket1 circuit to a [`Tk2Circuit`].
```
ss2165, line 26: still a couple mentions of "cast"?

ss2165 (approved):

PR #244, opened 2023-11-15
docs: Port text deleted in https://github.com/CQCL/hugr/pull/694 .

cqc-alec:
There was no obvious place to put this text so I created a `docs` directory.


ss2165 (changes requested):
Worth adding a reference + link to the main hugr spec, doesn't make much sense without it.

ss2165 (approved):
thanks 👍

PR #247, opened 2023-11-17
chore: update deps

aborgna-q:
- Updates the API changes introduced in https://github.com/CQCL/hugr/pull/680
- Uses the latest published portmatching instead of a git dependency
- Moves the lint configuration to the workspace Cargo.toml (new to rust [1.74](https://blog.rust-lang.org/2023/11/16/Rust-1.74.0.html))
  This throws warnings in older rust versions due to the unrecognised fields, but works alright.

### DIFF
--- a/tket2-py/src/lib.rs
+++ b/tket2-py/src/lib.rs
@@ -1,6 +1,4 @@
 //! Python bindings for TKET2.
-#![warn(missing_docs)]
-
 pub mod circuit;
 pub mod optimiser;
 pub mod passes;
--- a/tket2/src/json/op.rs
+++ b/tket2/src/json/op.rs
@@ -245,13 +245,15 @@ impl TryFrom<&OpType> for JsonOp {
         let mut num_qubits = 0;
         let mut num_bits = 0;
         let mut num_params = 0;
-        for ty in op.signature().input.iter() {
-            if ty == &QB_T {
-                num_qubits += 1
-            } else if *ty == *LINEAR_BIT {
-                num_bits += 1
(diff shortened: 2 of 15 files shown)


tket2-py/src/lib.rs
```diff
@@ -1,6 +1,4 @@
 //! Python bindings for TKET2.
-#![warn(missing_docs)]
```
ss2165, line 2: is this inherited now?

tket2/src/json/op.rs
```diff
@@ -245,13 +245,15 @@ impl TryFrom<&OpType> for JsonOp {
...
         let mut num_bits = 0;
         let mut num_params = 0;
-        for ty in op.signature().input.iter() {
-            if ty == &QB_T {
-                num_qubits += 1
-            } else if *ty == *LINEAR_BIT {
-                num_bits += 1
-            } else if ty == &FLOAT64_TYPE {
-                num_params += 1
+        if let Some(sig) = op.dataflow_signature() {
```
ss2165, line 248: not for this PR but we might be able to use `op: & impl DataflowOpTrait` in this function instead

ss2165 (approved):
Maybe we need a `HugrView::get_dataflow_op(Node) -> Box<dyn DataflowOpTrait>` thingemy

tket2-py/src/lib.rs
aborgna-q: Yes, that's the
```toml
[lints]
workspace = true
```
in Cargo.toml (since rust 1.74)

PR #252, opened 2023-11-20
feat: bindings for circuit cost and hash

aborgna-q:
- Adds `Tk2Circuit::hash` to the python bindings.
- Adds bindings for computing the cost of a circuit.
- Adds tests
  - I split `test_bindings.py` into `test_circuit` and `test_pass`


tket2-py/src/circuit/cost.rs
```diff
@@ -0,0 +1,168 @@
...
+        })
+    }
+}
+
+impl Sum for PyCircuitCost {
+    fn sum<I: Iterator<Item = Self>>(iter: I) -> Self {
+        Python::with_gil(|py| {
+            let mut acc = None;
+            for c in iter {
+                match &mut acc {
```
acl-cqc, line 79: Did you consider `fold` ?

tket2-py/src/circuit/cost.rs
```diff
@@ -0,0 +1,168 @@
...
+                    None => acc = Some(c.cost),
+                    Some(cost) => {
+                        *cost = cost
+                            .call_method1(py, "__add__", (c.cost,))
+                            .expect("Could not add circuit cost objects.")
+                    }
+                }
+            }
+            PyCircuitCost {
+                cost: acc.unwrap_or_else(|| py.None()),
```
acl-cqc, line 89:
```suggestion
                cost: acc.unwrap_or_else(py.None),
```

acl-cqc: Might be making things less clear, feel free to drop

tket2-py/src/circuit/cost.rs
```diff
@@ -0,0 +1,168 @@
...
+    fn default() -> Self {
+        Python::with_gil(|py| PyCircuitCost { cost: py.None() })
+    }
+}
+
+impl Add for PyCircuitCost {
+    type Output = PyCircuitCost;
+
+    fn add(self, rhs: PyCircuitCost) -> Self::Output {
+        Python::with_gil(|py| {
```
acl-cqc, line 38: I am reading all of these `impl`s with all of these `with_gil` and thinking - this would be so much easier if we had a wrapper on the python side and did the summing there. Indeed, probably also the mapping there. I guess that's not the current setup but if we are likely to head that way then it might save us a lot of code to just do that now....

acl-cqc (commented):
Ok, I didn't realize "bindings" meant python, which in hindsight I should have. Definitely not feeling very expert here so maybe it's best that you've reassigned this one, but a couple of thoughts....

tket2-py/src/circuit/convert.rs
```diff
@@ -1,18 +1,22 @@
...
 use tket2::json::TKETDecode;
 use tket2::passes::CircuitChunks;
+use tket2::{Circuit, T2Op};
 use tket_json_rs::circuit_json::SerialCircuit;
 
 use crate::rewrite::PyCircuitRewrite;
 
+use super::PyCircuitCost;
+
 /// A manager for tket 2 operations on a tket 1 Circuit.
```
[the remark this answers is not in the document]
aborgna-q: Outdated comment. I replaced with a better explanation and a short example.

tket2-py/src/circuit/convert.rs
```diff
@@ -73,6 +77,46 @@ impl Tk2Circuit {
...
     }
+
+    /// Compute the cost of the circuit based on a per-operation cost function.
+    ///
+    /// :param cost_fn: A function that takes a `T2Op` and returns an arbitrary cost.
+    ///     The cost must implement `__add__`.
+    /// :returns: The sum of all operation costs.
+    pub fn circuit_cost<'py>(&self, cost_fn: &'py PyAny) -> PyResult<&'py PyAny> {
+        let py = cost_fn.py();
+        let circ_cost = self.hugr.circuit_cost(|op| {
```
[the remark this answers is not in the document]
aborgna-q: Cleaned it up a bit. Hopefully is better now.

tket2-py/src/circuit/cost.rs
aborgna-q: It's annoying, but we're limited by the design. Once we have bindings for the hugr Walker we could revisit this.

tket2-py/src/circuit/cost.rs
aborgna-q: `py` is a variable, so it wouldn't work

tket2-py/src/circuit/cost.rs
```diff
@@ -0,0 +1,165 @@
...
+    fn cmp(&self, other: &Self) -> Ordering {
+        Python::with_gil(|py| {
+            let res = self
+                .cost
+                .call_method1(py, "__lt__", (&other.cost,))
+                .expect("Could not compare circuit cost objects.");
+            match res.is_true(py) {
+                Ok(true) => Ordering::Less,
+                _ => Ordering::Greater,
+            }
```
ss2165, line 123: 1. what if they are equal?
2. Why return Greater on `is_true` error?

tket2-py/src/circuit/cost.rs
acl-cqc: Ah! Sorry :-!

tket2-py/src/circuit/convert.rs
```diff
@@ -73,6 +97,48 @@ impl Tk2Circuit {
...
+                    "Could not convert circuit operation to a `Tk2Op`: {e}"
+                ))
+            })?;
+            let cost = cost_fn.call1((tk2_op,))?;
+            Ok(PyCircuitCost {
+                cost: cost.to_object(py),
+            })
+        };
+        let circ_cost = self.hugr.circuit_cost(cost_fn)?;
+        Ok(circ_cost.cost.into_ref(py))
```
acl-cqc, line 120: 👍 much clearer now, thanks!

ss2165 (approved):
👍

Issue #239, opened 2023-11-13
Cyclic circuit produced during Badger optimisation


aborgna-q:
#200 is now fixed, but more bugs remain...

Edit: See Luca's clearer explanation below: https://github.com/CQCL/tket2/issues/239#issuecomment-1809959354

Translating this qasm file into a hugr and applying the badger optimisation generates a cyclic graph. (Which causes the hashing code to panic).

```qasm
OPENQASM 2.0;
include "qelib1.inc";

qreg q[5];
cx q[4],q[1];
cx q[3],q[4];
cx q[1],q[2];
cx q[4],q[0];
u3(0.5*pi,0.0*pi,0.5*pi) q[1];
cx q[0],q[2];
cx q[3],q[1];
cx q[0],q[2];
```

The failing example is here:
https://github.com/CQCL/tket2/blob/9f752895c649d6e22f6f190efe77d1187046b7c6/tket2/examples/non_dag.rs

Checkout `debug/non-dag-example` and `cargo run --features portmatching --example non_dag`.

aborgna-q:
The problem comes from an erroneous invalidation set definition for the rewrite composition.

It's simpler to see it in this simpler (failing) circuits:
```qasm
OPENQASM 2.0;
include "qelib1.inc";

qreg q[4];
cx q[0],q[1]; # a
cx q[2],q[3]; # b
h q[1];
h q[3];
cx q[2],q[1]; # c
cx q[0],q[3]; # d
h q[0];
h q[1];
h q[2];
h q[3];
```
(hadamards added for padding, so the immediate neighbours do not overlap).

In that circuit, two possible rewrites are:
- swap `a` and `d`
- swap `b` and `c`
on their own they are valid CNOT swaps.

But if both are applied simultaneously, they create a loop in the circuit.

Original circuit:
![graphviz_diagram](https://github.com/CQCL/tket2/assets/121866228/deeac728-f574-4be4-811c-dddcfbacceef)

After swapping `b` and `c`:
![graphviz_diagram (1)](https://github.com/CQCL/tket2/assets/121866228/04754cf0-9837-47b6-aff5-e9a978d71831)

After swapping `a` and `d`:
![graphviz_diagram (2)](https://github.com/CQCL/tket2/assets/121866228/6b0e6788-d0e4-401a-b3e3-f62d6666ddac)

(Notice the loop in nodes $7 \to 3 \to 6 \to 13 \to 14 \to 4 \to 7$)

[...]

PR #255, opened 2023-11-21
fix: Check for rewrite composition in badger

aborgna-q:
This is a short-term quick-but-incomplete fix for #239.

We now check right after the composition whether we have invalidated the circuit by making a loop.
We leverage that circuit hashing already does a toposort traversal, adding an error when a loop is found and catching it.

Still, it is theoretically possible for an invalid chain of rewrites to end up producing a valid (but not equivalent) circuit. There is a discussion open for discussing more general solutions to this problem #242.


cqc-alec (approved):
Clever. Would it be difficult to add a test case to exercise this failure mode (which we'd of course hope to remove in due course)?

aborgna-q:
I added a test, but left it ignored-by-default since loading the rewrite set takes a while (#257).

PR #267, opened 2023-12-14
feat: rewrite tracing

### LINKED ISSUE #216: Rewrite tracing via metadata
(Optionally?) store the sequence of rewrites applied on each circuit in TASO and other passes, so we can get a better insight and stats on which rules are the most impactful.

aborgna-q:
Optionally store metadata about the rewrites applied to a circuit.

If the `rewrite-tracing` feature is disabled, this has no effect on the runtime.

If the feature is enabled, circuits must opt-in to the tracing by calling `RewriteTracer::enable_rewrite_tracing`.

Closes #216

### DIFF
--- a/tket2/src/rewrite/trace.rs
+++ b/tket2/src/rewrite/trace.rs
@@ -1,0 +1,126 @@
+//! Utilities for tracing the rewrites applied to a circuit.
+//!
+//! This is only tracked if the `rewrite-tracing` feature is enabled.
+
+use hugr::hugr::hugrmut::HugrMut;
+use hugr::hugr::NodeMetadata;
+use itertools::Itertools;
+
+use crate::Circuit;
+
+use super::CircuitRewrite;
+
+/// Metadata key for the circuit rewrite trace.
+pub const METADATA_REWRITES: &str = "TKET2.rewrites";
+
+/// Global read-only flag for enabling rewrite tracing.
+/// Enable it by setting the `rewrite-tracing` feature.
+///
+/// Note that circuits must be explicitly enabled for rewrite tracing by calling
+/// [`RewriteTracer::enable_rewrite_tracing`].
+pub const REWRITE_TRACING_ENABLED: bool = cfg!(feature = "rewrite-tracing");
+
+/// The trace of a rewrite applied to a circuit.
+///
+/// Traces are only enabled if the `rewrite-tracing` feature is enabled and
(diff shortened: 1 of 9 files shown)


tket2/src/rewrite/trace.rs
```diff
@@ -0,0 +1,126 @@
...
+
+/// Metadata key for the circuit rewrite trace.
+pub const METADATA_REWRITES: &str = "TKET2.rewrites";
+
+/// Global read-only flag for enabling rewrite tracing.
+/// Enable it by setting the `rewrite-tracing` feature.
+///
+/// Note that circuits must be explicitly enabled for rewrite tracing by calling
+/// [`RewriteTracer::enable_rewrite_tracing`].
+pub const REWRITE_TRACING_ENABLED: bool = cfg!(feature = "rewrite-tracing");
```
ss2165, line 21: is rust analyzer etc. good enough to ignore code guarded by this constant?

tket2/src/rewrite/trace.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///
+    /// Returns `None` if rewrite tracing is not enabled for this circuit.
+    #[inline]
+    fn rewrite_trace(&self) -> Option<Vec<RewriteTrace>> {
+        if !REWRITE_TRACING_ENABLED {
+            return None;
+        }
+        let meta = self.get_metadata(self.root(), METADATA_REWRITES)?;
+        let rewrites = meta.as_array()?;
+        Some(rewrites.iter().map_into().collect_vec())
```
ss2165, line 122: could return a `&[RewriteTrace]`?

ss2165 (approved):

ss2165:
There doesn't seem to be any key or namespace reservation happening in the metadata to denote a section as being for rewrites?

tket2/src/rewrite/trace.rs
aborgna-q: rust analyser checks the default features unless configured differently, bit this should work as any const flag.

tket2/src/rewrite/trace.rs
aborgna-q: Not possible with the `.map_into`, but we could return an `impl Iterator` once RPITIT` is here...

aborgna-q:
There is no structure for the metadata namespace yet. There's a discussion thread for that: https://github.com/CQCL/hugr/discussions/268

PR #270, opened 2023-12-14
feat!: Add lexicographic cost

lmondada:


tket2/src/circuit/cost.rs
```diff
@@ -41,14 +42,29 @@ pub trait CostDelta:
...
+    fn from((major, minor): (T, T)) -> Self {
+        Self([major, minor])
+    }
+}
+
+/// A cost that is ordered lexicographically.
+///
+/// An array of cost functions, where the first one is infinitely more important
+/// than the second, which is infinitely more important than the third, etc.
+#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, From)]
```
aborgna-q, line 57: If you just want to allow any `Into<[T; N]>`,
```suggestion
impl<const N: usize, V, T> From<V> for LexicographicCost<T, N>
where
    V: Into<[T; N]>,
{
    fn from(v: V) -> Self {
        Self(v.into())
    }
}

/// A cost that is ordered lexicographically.
///
/// An array of cost functions, where the first one is infinitely more important
/// than the second, which is infinitely more important than the third, etc.
#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
```

aborgna-q (approved):
:+1:

PR #275, opened 2023-12-15
feat: Return rewrite strategies as a generator

### LINKED ISSUE #269: feat: Return an iterator in `RewriteStrategy::apply_rewrites`
This would allow us to easily pick only a subset of pattern matches, sorted by their cost improvements.

This should probably wait for RPITIT to land with rust 1.75.

aborgna-q:
...instead of collecting a vector and returning it.

This lets us generate rewrites on-demand, so we avoid unnecessary work.

Closes #269


lmondada (approved):
Thanks!

PR #276, opened 2023-12-15
feat: Move pre/post rewrite cost to the RewriteStrategy API

aborgna-q:
Expose the rewrite cost delta functions


lmondada (approved):
Thanks! I had written the same code on my branch 😅

PR #277, opened 2023-12-18
feat!: Lexicographic cost fn

lmondada:
The major+minor circuit cost was generalised to an arbitrary lexicographic cost in #270. The old code now corresponds to the special case N=2.

This PR modifies the cost functions related to it to reflect that generalisation. I've also taken the opportunity to change the language a bit. I hope naming is now clearer and more consistent.


tket2/src/rewrite/strategy.rs
```diff
@@ -336,49 +338,51 @@ pub trait StrategyCost {
...
 }
 
-impl NonIncreasingGateCountCost<fn(&OpType) -> usize, fn(&OpType) -> usize> {
+impl LexicographicCostFunction<fn(&OpType) -> usize, 2> {
     /// Non-increasing rewrite strategy based on CX count.
     ///
-    /// The minor cost to break ties between equal CX counts is the number of
-    /// quantum gates.
+    /// A fine-grained cost function given by the total number of quantum gates
+    /// is used rank circuits with equal CX count.
```
ss2165, line 379:
```suggestion
    /// is used to rank circuits with equal CX count.
```

ss2165 (approved):
Just a typo

lmondada:
Thanks!

PR #288, opened 2024-01-17
chore: devenv and poetry update

ss2165:


devenv.nix
```diff
@@ -42,14 +42,18 @@ in
...
   };
 
   languages.python = {
     enable = true;
     poetry = {
       enable = true;
+      activate.enable = true;
+      # normal package is 1.71.0 which seems to break on arm macos
+      # this installs 1.4.2 instead
+      package = pkgs-stable.poetry;
```
aborgna-q, line 56: Is that the last working version? It looks old.

aborgna-q (approved):

devenv.nix
ss2165: https://github.com/NixOS/nixpkgs/pull/281397

it was fixed 11 hours ago so added that as the source

ss2165: (now v1.7.1)

aborgna-q (approved):

PR #293, opened 2024-02-21
feat: Support any ops in portmatching

aborgna-q:
Use the encoded operation as a hacky way to compare different instantiations (since we need it to implement `Hash + Eq + Ord`).

We can avoid the encoding when we know the operation has no parameters.


lmondada (approved):
It's great, thanks Agustin!

PR #296, opened 2024-02-27
feat: Use `IncomingPort` and `OutgoingPort` instead of `Port` where possible.

### LINKED ISSUE #220: Replace `Port` with `OutgoingPort`/`IncomingPort` where possible
A partial update was done in #219.

The two big targets of this issue are:

- `Units` iterator and filters. These return a generic `Port`, where it would be lot safer to have directed variants.

- `portmatching`. Here the move should be easy, except for the fact that the directed port variants do not currently implement `Serialize`, so they cannot be stored in the pattern structs.

aborgna-q:
Mostly a refactor. The `Units` struct required some API changes so the iterator can return the kind of port matching its direction.

Closes #220

### DIFF
--- a/tket2/src/circuit/units.rs
+++ b/tket2/src/circuit/units.rs
@@ -13,17 +13,15 @@
 //! [`Command`]: super::command::Command
 
 pub mod filter;
-pub use filter::FilteredUnits;
 
 use std::iter::FusedIterator;
+use std::marker::PhantomData;
 
 use hugr::types::{EdgeKind, Type, TypeRow};
-use hugr::CircuitUnit;
+use hugr::{CircuitUnit, IncomingPort, OutgoingPort};
 use hugr::{Direction, Node, Port, Wire};
 
 use crate::utils::type_is_linear;
-
-use self::filter::UnitFilter;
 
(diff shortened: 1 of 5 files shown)


tket2/src/circuit/units.rs
```diff
@@ -77,52 +73,69 @@ pub struct Units<UL = DefaultUnitLabeller> {
...
-    /// Apply a [`UnitFilter`] to the iterator's output, possibly unwrapping the
-    /// [`CircuitUnit`] into either a linear unit or a wire.
-    pub fn filter_units<F: UnitFilter>(self) -> FilteredUnits<F, UL> {
-        self.filter_map(F::accept)
+    /// Create a new reversed iterator over the units of a node.
+    ///
+    /// This iterator will yield all units reaching the given node,
+    /// and return the corresponding incoming ports.
+    #[inline]
+    pub(super) fn new_reversed(circuit: &impl Circuit, node: Node, unit_labeller: UL) -> Self {
```
croyzor, line 114: What's 'reversed' about this?

croyzor (approved):

PR #298, opened 2024-03-01
docs: Expand the main module and README docs

aborgna-q:
Adds a small usage example to the landing pages (the README and the main module doc).

drive-by: Re-export `Hugr`, so users can type the returned circuits without including the full crate.


README.md
```diff
@@ -1,15 +1,60 @@
-# tket2
+# tket2: The Hardware Agnostic Quantum Compiler
```
acl-cqc, line 1: I feel the hardware agnosticism is not new to TKET2 but is inherited from TKET1, and so in that respect we really are "Version 2 of the TKET compiler" (as you remove), or the sequel of, spawn of, etc. (Now *Hypertket* really is *spawn of* TKET1 for its multithreaded/distributed capabilities, but there...)

I'd say the big emphasis of TKET2 is on composable rewrites, rather than largely-monolithic "passes". That and the hypertket aspects/characteristics. YMMV though....

README.md
```diff
@@ -1,15 +1,60 @@
...
 ![msrv][]
 [![codecov][]](https://codecov.io/gh/CQCL/tket2)
 
-Version 2 of the TKET compiler.
-
   [build_status]: https://github.com/CQCL-DEV/hugr/workflows/Continuous%20integration/badge.svg?branch=main
   [msrv]: https://img.shields.io/badge/rust-1.75.0%2B-blue.svg
   [codecov]: https://img.shields.io/codecov/c/gh/CQCL/tket2?logo=codecov
 
+TKET2 is an open source quantum compiler developed by Quantinuum. Central to
```
acl-cqc, line 11: It is a bit of a shame to repeat these words exactly from the rust crate docs...

aborgna-q: These two are read from different places. The crate docs only really appear in `docs.rs`, whereas the readme is for github/crates.io.

We could make the crate docs more technical at some point, this is just so we have _something_ useful here.

README.md
aborgna-q: I think the "version 2" is already implicit from the name, and the previous subtitle didn't explain what this lib is really about.

This is a short one-liner, the actual crate content should be explained below. I'm happy to change the subtitle to something better :)

tket2/src/lib.rs
```diff
@@ -4,6 +4,43 @@
...
+//!
+//! assert_eq!(circ.qubit_count(), 9);
+//! assert_eq!(circ.num_gates(), 170);
+//!
+//! // Traverse the circuit and print the gates.
+//! for command in circ.commands() {
+//!     println!("{:?}", command.optype());
+//! }
+//!
+//! # // TODO: Waiting for mermaid support in hugr.
```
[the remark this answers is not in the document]
aborgna-q: Re-enabled this now that we've updated to hugr 0.3

README.md
```diff
@@ -1,15 +1,60 @@
...
+TKET2's design is its hardware agnosticism which allows researchers and
+quantum software developers to take advantage of its state of the art
+compilation for many different quantum architectures.
+
+TKET2 circuits are represented using the HUGR IR defined in the
+[quantinuum-hugr] crate. The [`Circuit`] trait provides a high level interface
+for working with HUGRs representing quantum circuits, and defines a HUGR
```
acl-cqc, line 18: Phrasing/grammar: this `and` suggests that the subject on the left, i.e. `The [```Circuit```] trait`, defines the HUGR extension ;-). How about:
```
...[quantinuum-hugr] crate. TKET2 augments Hugr with
* The [`Circuit`] trait, providing a high-level interface for working with HUGRs representing quantum circuits
* a HUGR extension with quantum operations
* A composable pass system for optimising circuits
* A number of built-in passes for common optimisations
```

acl-cqc (approved):
Ok, I am probably not the best person to define the role of TKET1 vs TKET2, so I'll go quiet on that one ;-). One suggested change, otherwise looks good to me, thanks

README.md
acl-cqc: Yes, I mean, it's a shame not to be able to link between them / copy text (they seem to be aimed at very similar audiences), but I admit I don't see a good way to do that, so ok nvm

acl-cqc: Oh, should possibly be `Quantinuum Ltd`

PR #299, opened 2024-03-01
feat: Add `PatternMatch::nodes` and `subcircuit` + matching example

### LINKED ISSUE #297: Expand the pattern matching API
Currently there is no direct way to query the nodes or the root of a `PatternMatch`.

(One needs to `.to_rewrite().subgraph().nodes()` instead)

aborgna-q:
In addition to the getter methods for `PatternMatch`, adds an example in the `portmatching` module docs using it.

Closes #297 .


PabloAndresCQ (approved):
Looks good to me!

PR #307, opened 2024-04-03
feat: expose Tk2Op name

Cobord:
a TODO was written for exposing the name as a public function on Tk2Op to avoid intos at caller

also update the comment on the impl of Rewrite of PullForward to reflect change in hugr


aborgna-q (approved):
Sorry for the delay. Thanks!

PR #310, opened 2024-04-12
fix: Track input linear units in `Command`

### LINKED ISSUE #309: Querying the linear inputs of a `QFree` command throws an error.
Test case: 
```rust
let qb_row = vec![QB_T; 1];
let mut h = DFGBuilder::new(FunctionType::new(qb_row.clone(), vec![]))?;
let [q_in] = h.input_wires_arr();
h.add_dataflow_op(Tk2Op::QFree, [q_in])?;
let circ = h.finish_hugr_with_outputs([], &REGISTRY)?;

let mut cmds = circ.commands();
let free_cmd = cmds.next().unwrap();
assert_eq!(
    free_cmd.inputs().map(|(unit, _, _)| unit).collect_vec(),
    [CircuitUnit::Linear(0)]
);
```

Error: 
```
thread 'circuit::command::test::alloc_free' panicked at tket2/src/circuit/command.rs:169:13:
Could not assign a linear unit to port Port(Incoming, 0) of node Node(3)
```

aborgna-q:
`Command` only stored the linear units assigned to its output ports, so querying the input linear unit to a `QFree` operation panicked.

This pr tracks the `input_linear_units` in addition to the output ones.

Fixes #309.

### DIFF
--- a/tket2/src/circuit/command.rs
+++ b/tket2/src/circuit/command.rs
@@ -10,6 +10,7 @@ use hugr::hugr::NodeType;
 use hugr::ops::{OpTag, OpTrait};
 use hugr::{IncomingPort, OutgoingPort};
 use itertools::Either::{self, Left, Right};
+use itertools::{EitherOrBoth, Itertools};
 use petgraph::visit as pv;
 
 use super::units::{filter, DefaultUnitLabeller, LinearUnit, UnitLabeller, Units};
@@ -25,11 +26,10 @@ pub struct Command<'circ, Circ> {
     circ: &'circ Circ,
     /// The operation node.
     node: Node,
-    /// An assignment of linear units to the node's ports.
-    //
-    // We'll need something more complex if `follow_linear_port` stops being a
-    // direct map from input to output.
-    linear_units: Vec<LinearUnit>,
+    /// An assignment of linear units to the node's input ports.
+    input_linear_units: Vec<LinearUnit>,
+    /// An assignment of linear units to the node's output ports.
+    output_linear_units: Vec<LinearUnit>,
 }
 
 impl<'circ, Circ: Circuit> Command<'circ, Circ> {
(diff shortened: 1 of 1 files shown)


tket2/src/circuit/command.rs
```diff
@@ -373,56 +369,61 @@ where
...
+                EitherOrBoth::Right((_, out_port, _)) => {
+                    // Add a new linear unit for this output port.
+                    let new_wire = Wire::new(node, out_port);
+                    let linear_id = self.wire_unit.len();
+                    self.wire_unit.insert(new_wire, linear_id);
+                    output_linear_units.push(LinearUnit::new(linear_id));
+                }
+                EitherOrBoth::Left((_, in_port, _)) => {
+                    // Terminate the input linear unit.
+                    let Some(linear_id) = self.circ.linked_outputs(node, in_port).next().and_then(
```
ss2165, line 399: can use `single_linked_input` given its linear?

tket2/src/circuit/command.rs
```diff
@@ -373,56 +369,61 @@ where
...
+                            self.wire_unit.remove(&Wire::new(wire_node, wire_port))
+                        },
+                    ) else {
+                        continue;
+                    };
+                    let linear_unit = LinearUnit::new(linear_id);
+                    let new_wire = Wire::new(node, out_port);
+                    input_linear_units.push(linear_unit);
+                    output_linear_units.push(linear_unit);
+                    self.wire_unit.insert(new_wire, linear_id);
```
ss2165, line 421: would be good to extract the logic in the previous cases in to functions and re-use them here

ss2165 (commented):
IIUC, now all linear wires will be assigned a linearunit index by the commands iterator (regardless of the actual linear type?). Meaning now the indices used to build using CircuitBuilder won't necessarily match the indices coming out of the commands iterator right? Is this what we want? Seems confusing.

aborgna-q:
We don't currently have a way to define matching inputs and outputs in the opdef.
The behaviour here is to list the linear inputs and the linear outputs sequentially and match them in order.

I think that should work for the majority of the cases. It's true that an opt-in high-grained configuration of it may be required in the future, but I'd leave it as a new issue.

ss2165 (approved):

PR #328, opened 2024-05-02
ci(pr-title): Escape variables used in bash scripts

aborgna-q:
Don't insert github expressions directly in the bash script.
- It breaks if there are [backticks] or [quotes] inside the variable.
- It's a security issue, it allows arbitrary code injections from the PR description.

The solution is to put them into env variables first.


aborgna-q:
Failing job due to the ` in my description :P
https://github.com/CQCL/tket2/actions/runs/8923278983/job/24507186630?pr=328

cqc-alec (approved):

PR #333, opened 2024-05-03
feat: pauli propagation use case example

ss2165:
Add bindings to emulate python builder - intent is to replace this with a pure python builder.

Add "CircuitBuilder" like utilities to assist appending operations to qubits.

Use the above to demonstrate pauli propagation through a circuit using exhaustive rewrite rule application.

+ Add some type annotations for circuit bindings, should ideally be replicated for other modules


tket2-py/tket2/tket2/_circuit.pyi
```diff
@@ -0,0 +1,106 @@
...
+    def input_node(self) -> Node:
+        """The input node of the circuit."""
+    def output_node(self) -> Node:
+        """The output node of the circuit."""
+
+class Tk2Op(Enum):
+    """A Tket2 built-in operation."""
+
+    H = 1
+    CX = 2
```
aborgna-q, line 33: Add all the others? At least the ones that appear in `build.py`.
```suggestion
    CX = auto()
    T = auto()
    S = auto()
    X = auto()
    Y = auto()
    Z = auto()
    Tdg = auto()
    Sdg = auto()
    ZZMax = auto()
    Measure = auto()
    RzF64 = auto()
    RxF64 = auto()
    PhasedX = auto()
    ZZPhase = auto()
    AngleAdd = auto()
    CZ = auto()
    TK1 = auto()
    QAlloc = auto()
    QFree = auto()
    Reset = auto()
```
It's annoying that the values are required. They do not exist on the rust side.

tket2-py/tket2/circuit/build.py
```diff
@@ -0,0 +1,141 @@
...
+
+@dataclass(frozen=True)
+class H(Command):
+    qubit: int
+    gate_name = "H"
+    n_qb = 1
+
+    def qubits(self) -> list[int]:
+        return [self.qubit]
+
```
aborgna-q, line 101: There's a bit of a mix here of Operations with no corresponding command, and commands with no first-level operation definition.

No need to actually add them here since this is a prototype, but maybe add a comment mentioning the difference?

aborgna-q (approved):
Awesome.

We'll rework this interface soon™, but this is a cool proof of concept for the API.

PR #341, opened 2024-05-13
chore: bump hugr to 0.4.0-alpha.1

doug-q:


aborgna-q (approved):

aborgna-q:
We haven't published tket2 yet, so I think depending on an alpha with the serialisation fixes should be alright.

Issue #350, opened 2024-05-21
Add a python-defined layer over the bindings.


aborgna-q:
Move the bindings to a `tket2._tket2` submodule and re-define all structures from python, backed by their binded counterparts.

See [Quantomatic/quizx](https://github.com/Quantomatic/quizx/tree/master/pybindings) for an example.

This gives us more flexibility on the python interface, where we can define methods that better reflect the python conventions.
Additionally, typing information is simpler and less error-prone; we only need the hand-written .pyi for the reduced binded methods and missing values will not affect the external API.

ss2165:
Summary of in-person discussion:
Following on from #465, we intend to use hugr-py `Node, Port, Wire` types in the surface python layer. For ops and types, things like `HugrType` or `TK2Ops` should implement the hugr-py `Op` and `Type` protocols for compatibility. They should also compare as equal to their hugr-py counterparts.

aborgna-q:
Closing this, as it will be included in any new tket2 bindings support / replaced by hugr-py uses.

PR #351, opened 2024-05-21
refactor: Move rust bindings to `tket2._tket2`

aborgna-q:
First step towards #350.

Deinterlaces the rust-binded modules from the pure python ones. And puts them in `_tket`
For now, we still re-export everything directly, I'll add the python wrappers and missing typing info in a following PR.


aborgna-q:
Removing the breaking change flag, since this only affects internal (`_*`) modules

ss2165 (approved):
nice

PR #352, opened 2024-05-22
feat: Add missing typing hints

aborgna-q:
Adds typing info for the currently binded classes and methods.
Also binds a couple classes that where missing.

Fixes #342.

PR #355, opened 2024-05-23
feat: drop pyo3 core dep

aborgna-q:
Adds a python-side definition of `Tk2Op` and `Pauli`, so we can drop the dependency on `pyo3` from the main rust package.

In addition of being necessary for #350, having only one package depending on `pyo3` removes building headaches due to it linking to python multiple times.

Blocked by #352


tket2-py/src/ops.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// Python equivalent of [`Tk2Op`].
+///
+/// [`Tk2Op`]: tket2::ops::Tk2Op
+#[pyclass]
+#[pyo3(name = "Tk2Op")]
+#[derive(Debug, Clone, From)]
+#[repr(transparent)]
+pub struct PyTk2Op {
+    /// Rust representation of the circuit chunks.
```
cqc-alec, line 28: "circuit chunks"?

tket2-py/src/ops.rs
```diff
@@ -0,0 +1,202 @@
...
+///
+/// Python equivalent of [`Pauli`].
+///
+/// [`Pauli`]: tket2::ops::Pauli
+#[pyclass]
+#[pyo3(name = "Pauli")]
+#[derive(Debug, Clone, From)]
+#[repr(transparent)]
+pub struct PyPauli {
+    /// Rust representation of the circuit chunks.
```
cqc-alec, line 110: ?

cqc-alec (approved):
LGTM

PR #360, opened 2024-05-24
ci: Use hugrbot in automation

aborgna-q:
Uses @hugrbot's token in CI, so messages and tokens are created with its username.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -19,5 +19,5 @@ jobs:
       - uses: googleapis/release-please-action@v4
         with:
             # Using a personal access token so releases created by this workflow can trigger the deployment workflow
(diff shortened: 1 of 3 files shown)


.github/workflows/release-please.yml
```diff
@@ -19,5 +19,5 @@ jobs:
       - uses: googleapis/release-please-action@v4
         with:
             # Using a personal access token so releases created by this workflow can trigger the deployment workflow
-            token: ${{ secrets.RELEASE_PLEASE_PAT }}
```
ss2165, line 22: have these secrets been removed?

ss2165 (approved):

.github/workflows/release-please.yml
aborgna-q: That token was never in this repo, I added the workflow yesterday and was waiting for the bot to configure it.

PR #368, opened 2024-05-29
feat!: init tket2-hseries

doug-q:
BREAKING CHANGE: require `hugr-0.5.0`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -739,7 +739,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a20246e5f1a0aae160b80b71bc4d5c1dcccc5605dbe6458ac6f9af73137339c3"
 dependencies = [
- "hugr-core",
+ "hugr-core 0.2.0",
  "hugr-passes",
 ]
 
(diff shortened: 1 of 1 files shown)


tket2-hseries/README.md
aborgna-q, line 1: Expand this a bit more.
See the `hugr-cli` [README](https://github.com/CQCL/hugr/blob/main/hugr-cli/README.md) for a template.

tket2-hseries/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[features]
+default = ["cli"]
+cli = ["dep:hugr-cli", "dep:clap"]
+
+[dependencies]
+hugr.workspace = true
+hugr-cli = { workspace = true, optional = true }
+clap = { workspace = true, optional = true, features = ["derive"] }
+tket2.workspace = true
```
aborgna-q, line 24: `tket2` is currently a workspace dependency, but I think it works better with `release-plz` if you define explicitly here;
```suggestion
tket2 = { path = "../tket2", version = "0.1.0-alpha.1" }
```
That version will get bumped in the release PR.

Can you also update tket2-py, and remove it from the workspace?

aborgna-q (commented):
Looks good 👍 

Please add this to the end of `release-plz.toml`:
```toml

[[package]]
name = "tket2-hseries"
release = true

# Disabled until the first version is manually published
publish = false
git_tag_enable = false
git_release_enable = false
```

aborgna-q:
And yes. Normally bumping a dependency is not a braking change, but here we re-export `Hugr` so a breaking change in `hugr` is also a breaking change here.

aborgna-q (approved):
👍 

We should add the `--help` output to the README (or an example) once there's something there.

PR #370, opened 2024-05-31
refactor!: Replace Circuit trait with a struct

### LINKED ISSUE #112: Restrict `Circuit` to HUGRs with valid roots
Instead of a blanket implementation on `T: HugrView`.
This will probably need some dark type magic if we want to keep a clean API.

aborgna-q:
Replaces the `trait Circuit : HugrView` with a struct containing a `T: HugrView` and a node id indicating the container node for the circuit in the hugr.

There are a few design going into this definition:

- #### Struct with checked parent node type.
  The previous `Circuit` trait asked the user to only use it for dataflow-based hugrs. This of course lead to functions assuming things about the hugr structure and panicking with random errors when passed an otherwise valid hugr.
  
  Bar checking the root node at each method call, this can only be solved with an opaque struct that checks its preconditions at construction time.
  
  The code here will throw a user-friendly error when trying to use incompatible hugrs:
  ```
  Node(0) cannot be used as a circuit parent. A Module is not a dataflow container.
  ```

- #### Generic `T` defaulting to `Hugr`
  Most uses in this library manipulate circuits as owned structures (e.g. passing circuits around in badger).
  However, sometimes we don't own the hugr so we cannot create a `Circuit<Hugr>` from it. With the generics definition we can use `Circuit<&Hugr>` or `Circuit<&mut Hugr>` to respectively view or modify non-owned structures.
  
- #### Parent node pointer
  Circuits represented as hugrs are not necessarily a flat dataflow structure.
  For example, a guppy-defined circuit may include calls to other methods from the module.
  By including a pointer to the entry point of the circuit we can keep all this structure, and rewrite it latter as necessary.
  This is also useful for non-owned hugrs; we can have a `Circuit<&Hugr>` pointing to a region in the hugr without needing to modify anything else.

Closes #112.

BREAKING CHANGE: Replaced the `Circuit` trait with a wrapper struct.


tket2/src/circuit.rs
```diff
@@ -28,54 +28,138 @@ pub use hugr::{Node, Port, Wire};
...
+    }
+
+    /// Unwrap the HUGR containing the circuit.
+    pub fn into_hugr(self) -> T {
+        self.hugr
+    }
+
+    /// Get a mutable reference to the HUGR containing the circuit.
+    ///
+    /// Mutation to the hugr MUST NOT invalidate the parent node,
```
ss2165, line 89:
```suggestion
    /// Mutation of the hugr MUST NOT invalidate the parent node,
```

tket2/src/json.rs
```diff
@@ -37,23 +36,23 @@ const METADATA_Q_REGISTERS: &str = "TKET1_JSON.qubit_registers";
 /// Explicit names for the input bit registers.
 const METADATA_B_REGISTERS: &str = "TKET1_JSON.bit_registers";
 
-/// A JSON-serialized circuit that can be converted to a [`Hugr`].
+/// A JSON-serialized circuit that can be converted to a [`Circuit`].
```
ss2165, line 39: specify tket1 format

tket2/src/rewrite/trace.rs
```diff
@@ -67,18 +67,19 @@ impl From<RewriteTrace> for serde_json::Value {
     }
 }
 
-/// Extension trait for circuits that can trace rewrites applied to them.
+/// Implementation for rewrite tracing in circuits.
 ///
 /// This is only tracked if the `rewrite-tracing` feature is enabled and
 /// `enable_rewrite_tracing` is called on the circuit before.
-pub trait RewriteTracer: Circuit + HugrMut + Sized {
+impl<T: HugrMut> Circuit<T> {
```
ss2165, line 74: nice

ss2165 (approved):
huge, thanks

tket2/src/circuit.rs
```diff
@@ -28,54 +28,138 @@ pub use hugr::{Node, Port, Wire};
...
-// - Depth
-pub trait Circuit: HugrView {
+/// A quantum circuit, represented as a function in a HUGR.
+#[derive(Debug, Clone, PartialEq)]
+pub struct Circuit<T = Hugr> {
+    /// The HUGR containing the circuit.
+    hugr: T,
+    /// The parent node of the circuit.
+    ///
+    /// This is checked at runtime to ensure that the node is a DFG node.
```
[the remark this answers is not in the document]
aborgna-q: A priori we want to support any hugr as long as the top-level node can be interpreted as a runnable circuit.

If anything, we may want to relax this later to accept `FuncDecl` later?

PR #375, opened 2024-06-03
chore: Fix mypy lints

aborgna-q:
Re-enables mypy on the pre-commit checks now that we have typing info on the bindings, and fixes all errors.

We are not checking this in CI yet. I'll put that in a separate PR with other CI changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,8 +112,7 @@ jobs:
       - name: Build pyo3 bindings
         run: |
           poetry install
-          poetry run maturin build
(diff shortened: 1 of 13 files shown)


aborgna-q:
Looks like our CI job had some issues with cache.
Even though I updated `passes.py:badger_pass`, it was still using and older cached version :/ https://github.com/CQCL/tket2/actions/runs/9350614107/job/25734463175?pr=375#step:8:61

Swapping `maturin build` for `maturin develop` (so that it gets automatically added to the venv) fixed the problem.

cqc-alec (approved):

PR #381, opened 2024-06-04
fix: Commands iterator ignoring the hierarchy.

### LINKED ISSUE #42: Test nested circuit regions & command iterator.
Make sure that nested HUGRs are supported by the Circuit interface.

Update the command iterator once everything is checked.
https://github.com/CQCL-DEV/tket2/pull/37#discussion_r1269154445

aborgna-q:
The commands iterator did a toposort on the whole Hugr, ignoring the hierarchy.
This generated problems with #370, since circuits may now be nested somewhere in the hugr.
It would also have caused problems with circuit boxes, but we don't support that yet here.

We use a `SiblingGraph` now, to ensure that we only explore the top-level region of the circuit.
Subcircuits will be returned as a single command.

Adds a `build_module_with_circuit` helper to build circuits inside modules.

Closes #42.


ss2165 (approved):
good spot

PR #391, opened 2024-06-07
fix: Support non-DFG circuits

### LINKED ISSUE #385: Test `FuncDefn` circuits instead of `DFG`s
We currently parse tket1 circuits into a `DFG` node, and store their name in a `name` metadata entry.
Circuits defined in guppy use a `FuncDefn` instead, and store their name in the function op definition.

We should drop the metadata entry, and rely on function definitions. `DFG`-defined circuits will not define a name.

FuncDefns require defining the extension deltas in their definitions. This seems to cause issues with rewrite operatios, as they currently do not allow for modifying the set of extensions.

### LINKED ISSUE #389: `CircuitRewrite` does only support Dfg-rooted circuits
`Subcircuit::create_rewrite` and `CircuitRewrite::try_new` convert the target circuit to work with the requirements of `hugr::SimpleReplacement`.

See https://github.com/CQCL/hugr/issues/1169 https://github.com/CQCL/hugr/issues/1171

aborgna-q:
Followup to #390.
Fixes #389.
Closes #385.
~~Blocked by https://github.com/CQCL/hugr/issues/1175.~~

Fixes support for non-`Dfg` circuits and circuits with a non-root parent:

Adds a `Circuit::extract_dfg(&self)` function that extracts the circuit into a new hugr with a DFG operation at the root.
In some cases, like in a CFG DataflowBlock node, this requires some changes to the definition to eliminate the output sum type.
Here I only implemented it for the kind of blocks produced by guppy. We could replace the manual implementation once https://github.com/CQCL/hugr/issues/818 gets implemented.

With this we can now fix #389 by extracting the circuit before using it as a replacement in `SimpleReplacement::create_simple_replacement`.

Replaces `DfgBuilder` with `FunctionBuilder` where possible, so we can use named circuits in the tests.
(This failed before due to the bug in CircuitRewrite).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -179,9 +179,9 @@ checksum = "37b2a672a2cb129a2e41c10b1224bb368f9f37a2b16b612598138befd7b37eb5"
 
 [[package]]
 name = "cc"
-version = "1.0.99"
+version = "1.0.98"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96c51067fd44124faa7f870b4b1c969379ad32b2ba805aa959430ceaa384f695"
+checksum = "41c270e7540d725e65ac7f1b212ac8ce349719624d7bcff99f8e2e488e8cf03f"
 
 [[package]]
 name = "cesu8"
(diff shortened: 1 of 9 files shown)


ss2165 (approved):
LGTM

the rewrite from cfg block to dfg without tag/unit type is indeed a bit hacky... even a simple change like guppy outputting just the Unit value as a const (rather than a Tag node) would break it. But I think it's ok for now.

PR #406, opened 2024-06-12
feat: tuple unpack rewrite

### LINKED ISSUE #380: Add a pass to remove tuples from circuits
Try to flatten every tuple into individual bits/qubits, so that it can be converted to a pytket circuit.

This may be implemented on the hugr lib instead of here.

aborgna-q:
Adds a pass for removing tuple packs immediately followed by tuple unpacks.

Currently this only removes one-to-one connections.
The code should work for one-to-many, but there's a bug in hugr's `SimpleRewrite` that prevents us from having replacements where multiple outputs are connected to a single input.
See https://github.com/CQCL/hugr/issues/1190.

Closes #380

### DIFF
--- a/tket2/src/passes/tuple_unpack.rs
+++ b/tket2/src/passes/tuple_unpack.rs
@@ -71,8 +71,10 @@ fn make_rewrite<T: HugrView>(circ: &Circuit<T>, cmd: Command<T>) -> Option<Circu
         return None;
     }
 
-    // TODO: SimpleReplacement currently fails when replacements have multiports.
+    // TODO: SimpleReplacement fails when replacements have multiports.
     // We only support matching a single unpack for now.
+    //
+    // This can be removed once https://github.com/CQCL/hugr/pull/1191 gets released.
     if unpack_nodes.len() > 1 {
         return None;
     }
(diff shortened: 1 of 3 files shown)


tket2/src/passes/tuple_unpack.rs
```diff
@@ -0,0 +1,202 @@
...
+
+    if unpack_nodes.is_empty() {
+        return None;
+    }
+
+    // TODO: SimpleReplacement currently fails when replacements have multiports.
+    // We only support matching a single unpack for now.
+    if unpack_nodes.len() > 1 {
+        return None;
+    }
```
aborgna-q, line 78: We can remove this once https://github.com/CQCL/hugr/pull/1191 gets released.

tket2/src/passes/tuple_unpack.rs
```diff
@@ -0,0 +1,202 @@
...
+    // Remove all unpack operations, but only remove the pack operation if all neighbours are unpacks.
+    match links.len() == unpack_nodes.len() {
+        true => Some(remove_pack_unpack(
+            circ,
+            &tuple_types,
+            tuple_node,
+            unpack_nodes,
+        )),
+        false => {
+            // TODO: Add a rewrite to remove some of the unpack operations.
```
cqc-alec, line 89: Should we capture this in an issue?

cqc-alec (approved):
Looks good!

tket2/src/passes/tuple_unpack.rs
aborgna-q: #416

PR #407, opened 2024-06-12
feat: guppy → pytket conversion

aborgna-q:
Adds support for converting **flat** **pure** functions defined in guppy into pytket circuits. See the eexample in `test_guppy.py`.

This required a hacky addition to `Circuit::extract_dfg` to remove unary unit sums in the output of the dataflow blocks generated by guppy.
We should be able to replace this with cfg inlining once that's implemented.


tket2-py/test/test_guppy.py
```diff
@@ -9,8 +10,46 @@
...
+    module = GuppyModule("test")
+    module.load(quantum)
+
+    @guppy(module)
+    @no_type_check
+    def my_func(
+        q0: qubit,
+        q1: qubit,
+    ) -> tuple[qubit, qubit]:  # pragma: no cover
+        q0 = phased_x(q0, py(math.pi / 2), py(-math.pi / 2))
```
ss2165, line 26: tket1 has angles in fractions of pi, are these angles converted correctly?

tket2-py/test/test_guppy.py
```diff
@@ -9,8 +10,46 @@
...
+        q0 = rz(q0, py(math.pi))
+        q1 = rz(q1, py(math.pi))
+        return (q0, q1)
+
+    # Compile the module
+    hugr = module.compile()
+    json = hugr.to_raw().to_json()
+
+    # Load it from the JSON string
+    circ = Tk2Circuit.from_guppy_json(json, "my_func")
```
ss2165, line 40: these three lines probably belong in a function, in _circuit.py orsomething

ss2165 (commented):
Looks good, just a couple of questions. Mainly worried the angles will lead to confusing examples for other people.

tket2-py/test/test_guppy.py
aborgna-q: There's a couple reasons for keeping this explicit for the moment:

Without https://github.com/CQCL/guppylang/issues/246, we need to keep an explicit reference to the hugr if we want to create multiple circuits from the same module.
Once that's published we'll be able to create functions from the function definition itself;
```python
from_guppy(my_func: guppy.Definition) -> Tk2Circuit
```

But more importantly, `tket2` only has a _dev-dependency_ on guppy.
To avoid cyclic dependencies, we cannot expose `from_guppy` in the library API.
In the future I imagine guppy may import tket2 and expose `to_circuit` methods on modules and functions.

tket2-py/test/test_guppy.py
aborgna-q: Good catch. That was missing in the encoder/decoder.

Now we translate between radians and half-turns when converting hugrs to pytket circuits.

tket2/src/serialize/pytket/tests.rs
```diff
@@ -108,8 +108,8 @@ fn circ_add_angles_constants() -> Circuit {
 
     let qb = h.input_wires().next().unwrap();
 
-    let point2 = h.add_load_value(ConstF64::new(0.2));
-    let point3 = h.add_load_value(ConstF64::new(0.3));
+    let point2 = h.add_load_value(ConstF64::new(0.2 * std::f64::consts::PI));
+    let point3 = h.add_load_value(ConstF64::new(0.3 * std::f64::consts::PI));
```
aborgna-q, line 112: I'm surprised this doesn't cause precision errors.
The parameter still gets encoded as "0.2 + 0.3" in the test below.

tket2-py/test/test_guppy.py
aborgna-q: I added a utility functions for testing, but we won't be able to expose it in the tket2 api.

ss2165: makes sense, thanks for the explanation

PR #417, opened 2024-06-18
feat: Initial support for `TailLoop` as circuit parent

doug-q:


aborgna-q:
The problem with tail loops at the moment is that "iterating commands/operations" is not clearly defined.
Currently the command iterator only runs on flat regions with no control flow, and just emits `TailLoop`/`Conditional`/`CFG` commands.

This is rather in the edge between the circuit abstraction and full-fledged HUGRs. We should probably delineate what should a `Circuit` _be_ before merging this.

tket2/src/circuit.rs
```diff
@@ -352,10 +352,14 @@ fn check_hugr(hugr: &impl HugrView, parent: Node) -> Result<(), CircuitError> {
         },
         OpType::DataflowBlock(_) => Ok(()),
         OpType::Case(_) => Ok(()),
-        _ => Err(CircuitError::InvalidParentOp {
-            parent,
-            optype: optype.clone(),
-        }),
+        OpType::TailLoop(_) => Ok(()),
+        _ => {
+            assert_eq!(None, optype.tag().partial_cmp(&OpTag::DataflowParent));
```
aborgna-q, line 357: Avoid panicking in release builds.
```suggestion
            debug_assert_eq!(
                None,
                optype.tag().partial_cmp(&OpTag::DataflowParent),
            );
```

aborgna-q (approved):
> I would argue that a Circuit with a TailLoop as a parent IS a flat region with no control flow (assuming no control flow in it's children).

Fair enough 👍

Issue #44, opened 2023-07-25
Decode TKET1 boxes


aborgna-q:

aborgna-q:
Blocked by https://github.com/CQCL/tket2/issues/308

PR #430, opened 2024-06-25
feat: Add almost all missing pure quantum ops from pytket to `Tk2Op`

### LINKED ISSUE #428: Define missing pure operations from pytket as `Tk2Op`s

aborgna-q:
Adds most of the pure gates from https://docs.rs/tket-json-rs/latest/tket_json_rs/optype/enum.OpType.html.

This is still missing some ops:
- Gates with n-qubit controls. That requires computing the signature dynamically, I'll add an issue for it.
- `Barrier`. Related to the previous point.
- Global `Phase`. I'm not sure if we want an explicit gate for it? We should discuss it first.
- `PhaseGadget`. It's defined in tket-json-rs, but it doesn't appear in the pytket documentation?
- Boxes. That needs explicit support (#44)

I added docstrings, but they are missing the unitary definition. We need to add latex rendering to our rustdocs for that (I'll add an issue).

I filled some commutation properties in `Tk2Op::qubit_commutation`, but I'm sure its missing some more.

Closes #428


ss2165:
I'm not sure we want to do this....I wanted to keep the core opset small and rebase in/out as necessary. 

Most of these ops are only relevant to certain passes.

tket2-py/tket2/ops.py
```diff
@@ -6,27 +6,69 @@
...
     Sdg = auto()
-    ZZMax = auto()
-    Measure = auto()
-    RzF64 = auto()
+    SX = auto()
+    SXdg = auto()
+    V = auto()
+    Vdg = auto()
     RxF64 = auto()
+    RyF64 = auto()
```
cqc-alec, line 22: This is not related to this PR, but I wonder why we added the "F64" suffix to Rx, Ry and Rz, but not to other operations that have angle parameters (e.g. U1).

(I actually prefer without, for conciseness.)

tket2/src/ops.rs
```diff
@@ -39,31 +39,197 @@ use crate::extension::REGISTRY;
...
+    Vdg,
+    /// X rotation with an angle parameter in half-turns
     RxF64,
+    /// Y rotation with an angle parameter in half-turns
+    RyF64,
+    /// Z rotation with an angle parameter in half-turns
+    RzF64,
+    /// TK1 gate
+    ///
+    /// TK1(α, β, γ) = Rz(α) Rx(β) Rz(γ)
```
cqc-alec, line 83:
```suggestion
    /// TK1(α, β, γ) = RzF64(α) RxF64(β) RzF64(γ)
```

tket2/src/ops.rs
```diff
@@ -39,31 +39,197 @@ use crate::extension::REGISTRY;
...
+    ///
+    /// See [Tk2Op::U3]
+    CU3,
+    /// GPi gate
+    GPI,
+    /// GPi2 gate
+    GPI2,
+
+    // Other multi-qubit gates
+    /// ZZPhase gate, with a parameter in half-turns
```
cqc-alec, line 158:
```suggestion
    /// XXPhase gate, with a parameter in half-turns
```

tket2/src/ops.rs
```diff
@@ -39,31 +39,197 @@ use crate::extension::REGISTRY;
...
+    XXPhase,
+    /// ZZPhase gate, with a parameter in half-turns
```
cqc-alec, line 160:
```suggestion
    /// YYPhase gate, with a parameter in half-turns
```

tket2/src/ops.rs
```diff
@@ -39,31 +39,197 @@ use crate::extension::REGISTRY;
...
+
+    Barrier,
+    /// N-qubit controlled X gate
+    CnX,
+    /// N-qubit controlled Y gate
+    CnY,
+    /// N-qubit controlled Z gate
+    CnZ,
+    /// N-qubit controlled Ry gate
+    CnRy,
```
cqc-alec, line 229: CnRx and CnRz exist now too.

cqc-alec (approved):

tket2-py/tket2/ops.py
aborgna-q: Yes, I was going to propose dropping the suffix in another PR

ss2165:
don't care too much about symmetry, zzphase is a native gate on H-series so convenient to add

PR #437, opened 2024-06-27
docs: Better error reporting in portmatching

aborgna-q:
If the pattern had empty wires, `try_from_circuit` failed with a `NotConnected` error. This adds an `EmptyWire` variant instead.

### DIFF
--- a/tket2/src/portmatching/pattern.rs
+++ b/tket2/src/portmatching/pattern.rs
@@ -71,10 +71,19 @@ impl CircuitPattern {
                     .expect("invalid circuit")
             })
             .collect_vec();
-        if inputs.iter().flatten().any(|&(n, _)| n == out) {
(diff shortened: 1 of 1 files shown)


tket2/src/portmatching/pattern.rs
```diff
@@ -120,13 +129,23 @@ impl Debug for CircuitPattern {
 
 /// Conversion error from circuit to pattern.
 #[derive(Debug, Error, PartialEq, Eq)]
+#[non_exhaustive]
 pub enum InvalidPattern {
     /// An empty circuit cannot be a pattern.
-    #[error("empty circuit is invalid pattern")]
+    #[error("Empty circuit are not allowed as patterns")]
```
ss2165, line 135:
```suggestion
    #[error("Empty circuits are not allowed as patterns")]
```

ss2165 (approved):

PR #438, opened 2024-06-27
fix(tket2-py): Replace `with_hugr` helpers with `with_circ`, keeping the circuit parent.

aborgna-q:
Avoids unwrapping the hugr inside `Tk2Circuit`.
This caused problems with circuits with non-root parents. Circuits defined in guppy ended up pointing to the root module instead of the chosen function definition.

Adds the first rust-side test for `tket2-py`.
I'm pretty sure I'll have to install the python interpreter in the CI rust test job, but let's see the error first.

### DIFF
--- a/tket2-py/src/utils.rs
+++ b/tket2-py/src/utils.rs
@@ -1,5 +1,16 @@
 //! Utility functions for the python interface.
 
+use hugr::builder::{
+    BuildError, CircuitBuilder, Container, Dataflow, DataflowSubContainer, FunctionBuilder,
+    HugrBuilder, ModuleBuilder,
+};
+use hugr::extension::prelude::QB_T;
+use hugr::extension::PRELUDE_REGISTRY;
+use hugr::ops::handle::NodeHandle;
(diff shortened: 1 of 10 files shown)


tket2-py/src/utils.rs
```diff
@@ -45,9 +56,51 @@ macro_rules! create_py_exception {
...
 pub(crate) use create_py_exception;
-use itertools::Itertools;
 
 /// Convert an iterator of one type into vector of another type.
 pub fn into_vec<T, S: From<T>>(v: impl IntoIterator<Item = T>) -> Vec<S> {
     v.into_iter().map_into().collect()
 }
+
+/// Utility for building a module with a single circuit definition.
+#[allow(unused)]
```
cqc-alec, line 66: What's unused?

aborgna-q: It's only used when testing, so It may get reported as unused on non-test builds.
I copied this from `tket2`, but maybe it should go inside the `test` module.

PR #439, opened 2024-06-27
feat(tket2-py): Bind the `lower_to_pytket` pass in python

aborgna-q:
This is used to normalise guppy functions into flat pytket-like circuits.
This is useful for using as pattern match rules.

We may remove the method from `tket2.passes` later, but it's useful to have as an internal binding in `tket2._tket2.passes`.


lmondada (approved):

lmondada:
Nice!

PR #442, opened 2024-06-27
fix: `extract_dfg` inserting the output node with an invalid child order

aborgna-q:
When replacing a CFG's output with the correct DFG output operation, we inserted the new node at the end of the container's children, instead of just after the input.

Includes a new unit test which simulates a circuit definition from guppy, where we can run the `lower_to_pytket` pass.

### DIFF
--- a/tket2/src/circuit/extract_dfg.rs
+++ b/tket2/src/circuit/extract_dfg.rs
@@ -47,7 +47,7 @@ fn remove_cfg_empty_output_tuple(
     signature: FunctionType,
 ) -> Result<FunctionType, CircuitMutError> {
     let sig = signature;
-    let parent = circ.parent();
+    let input_node = circ.input_node();
 
     let output_node = circ.output_node();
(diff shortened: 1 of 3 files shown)


tket2/src/circuit/extract_dfg.rs
```diff
@@ -89,8 +89,8 @@ fn remove_cfg_empty_output_tuple(
     let new_op = Output {
         types: new_types.clone().into(),
     };
-    let new_node = hugr.add_node_with_parent(
-        parent,
+    let new_node = hugr.add_node_after(
+        input_node,
```
ss2165, line 93: this should be covered right? the test _looks_ ok to me

aborgna-q: Turns out the "guppy-like" circuit was not guppy-like. I was loading a constant tag instead of using the `Tag` operation...

It should be fixed now.

ss2165 (approved):

PR #443, opened 2024-06-27
docs: Add some example notebooks for the python package.

### LINKED ISSUE #434: Add some initial usage examples to the python lib

aborgna-q:
The code in the notebooks requires #442, #441, and #444

Closes #434


ss2165 (commented):
- guppy is about to stop returning results and use a `result` function. How breaking will this be? https://github.com/CQCL/guppylang/pull/271
- the no_type_checks and pragma: no_covers are adding annoying noise. https://github.com/CQCL/guppylang/issues/277
- define `guppy_to_circuit` in a separate file? jupyter notebooks can do local imports
- for zzphase(0.5) -> cnot the text says "replaced with just cx and rz" but the actual rhs has way more Rzs going on?
- is the pauli propagation copying the one in the tests?

aborgna-q:
> * guppy is about to stop returning results and use a `result` function. How breaking will this be? [feat: Add result function guppylang#271](https://github.com/CQCL/guppylang/pull/271)

We'll just ignore Result operations. I'll have to think about how to encode them into pytket circuits, but for tket2 circuits we only care about the function definitions with explicit outputs. `Result` should be akin to a print statement we'll optimise around.

> * the no_type_checks and pragma: no_covers are adding annoying noise. [Automatically annotate guppy definitions with `@no_type_check` guppylang#277](https://github.com/CQCL/guppylang/issues/277)

As Mark mentioned there, they cannot be added programmatically u.u
But we do not check notebook code, so maybe I should remove them here.

> * define `guppy_to_circuit` in a separate file? jupyter notebooks can do local imports

Will do 👍 

> * for zzphase(0.5) -> cnot the text says "replaced with just cx and rz" but the actual rhs has way more Rzs going on?

I'll fix the text

> * is the pauli propagation copying the one in the tests?

Yes, it's an extract of your code. That works well as an example of how to use the pattern matcher for more complex operations, so it seemed like a good thing to include here.
It some point we may want to expand it on its own notebook, but here we just have a quick overview of all the features.

ss2165 (approved):
:shipit:

PR #465, opened 2024-07-09
feat!: get pauli propagation + examples working with new hugr builder 

ss2165:
Currently requires some horrible hacks like converting nodes via ints and loading operations from bytes. This will be cleared up by fully using the `hugr` package Node, Wire, ops, types in the `TK2Circuit` interface instead of ad-hoc bindings in a follow up PR.

BREAKING CHANGE: `Dfg` binding removed, replaced with builder from `hugr` package.


pyproject.toml
```diff
@@ -36,7 +38,8 @@ mypy = "^1.10.1"
 hypothesis = "^6.105.1"
 graphviz = "^0.20"
 pre-commit = "^3.7.1"
-guppylang = "^0.6.0"
+# guppylang = "^0.6.0"
+guppylang = { git = "https://github.com/CQCL/guppylang", rev = "3da3936" }
```
ss2165, line 42: needed for hugr-py compatibility

tket2-py/src/circuit/tk2circuit.rs
```diff
@@ -178,7 +178,7 @@ impl Tk2Circuit {
         Ok(self.clone())
     }
 
-    fn node_op(&self, node: PyNode) -> PyResult<PyCustomOp> {
+    fn node_op(&self, node: PyNode) -> PyResult<Cow<[u8]>> {
```
ss2165, line 181: temporary hack

pyproject.toml
aborgna-q: This will prevent us from releasing new tket2-py versions. Can we do a guppy release soon?

tket2-py/src/circuit/tk2circuit.rs
aborgna-q: 😬

tket2-py/tket2/circuit/build.py
```diff
@@ -1,137 +1,189 @@
...
+
+Measure = MeasureDef()
+
+
+FLOAT_T = tys.Opaque(
+    extension="arithmetic.float.types",
+    id="float64",
+    args=[],
+    bound=tys.TypeBound.Copyable,
+)
```
aborgna-q, line 120: This should be defined in `hugr-py`. Open an issue there?

aborgna-q (approved):
Nice

I'm not a fan of the `[:]` prefix after `CircBuild.add`, but things look much cleaner now.

tket2-py/tket2/circuit/build.py
ss2165: will do a PR adding some prelude types

ss2165: https://github.com/CQCL/hugr/pull/1288

ss2165:
Will wait for hugr-py 0.4 release before mergin

PR #466, opened 2024-07-10
docs: simplify the getting started notebook

ss2165:
No need to rebuild the same guppy function


aborgna-q (approved):
This will produce different outputs if we run cells out-of-order, but it should be fine.

PR #475, opened 2024-07-11
feat: Partial tuple unpack

### LINKED ISSUE #416: Partial tuple pack-unpack rewrite.
#406 added a simple pass that removes tuple packs immediately followed by one or more tuple unpacks.

This does not support the more complex case where only _some_ of the `MakeTuple` neighbours are `UnpackTuple`s. In that case, we can remove the unpacks but the `MakeTuple` cannot be removed.

We should implement that case too.

https://github.com/CQCL/tket2/blob/425d6b4fcd1ac283fc74d1d23b6820fd5e6f3bbe/tket2/src/passes/tuple_unpack.rs#L91

aborgna-q:
Closes #416.

Requires ~https://github.com/CQCL/hugr/issues/1295~ https://github.com/CQCL/hugr/pull/1324 to be released.

### DIFF
--- a/tket2/src/passes/tuple_unpack.rs
+++ b/tket2/src/passes/tuple_unpack.rs
@@ -3,7 +3,7 @@
 use core::panic;
 
 use hugr::builder::{DFGBuilder, Dataflow, DataflowHugr};
-use hugr::ops::{OpTrait, OpType};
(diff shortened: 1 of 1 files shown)


aborgna-q:
Now blocked by https://github.com/CQCL/hugr/issues/1323

cqc-alec (approved):

PR #477, opened 2024-07-12
chore(py): Specify py dependencies for packaged wheels

### LINKED ISSUE #476: Dependencies not included in package
Originally reported by @sjdilkes 
The wheels being built are not including dependencies. This is probably because the dependencies are only under the poetry tool section of `pyproject.toml`. They may need to be duplicated. Is there a way that avoids duplication?

aborgna-q:
Fixes #476 

This is an annoying duplication due to poetry not following PEP 631 and reading dependencies from `[project.dependencies]`.


ss2165 (approved):
thanks

PR #478, opened 2024-07-12
ci: Group dependabot updates

### LINKED ISSUE #462: ci: Group updates in dependabot
Add some rules to dependabot so we don't get flooded with dependency updates each Monday.
https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/

We'll probably want to keep hugr-* updates separate, and mayor updates too?

aborgna-q:
Closes #462

Tries to avoid the flood of package updates by grouping them into categories.
- Hugr packages get their own group
- Dev dependencies get grouped together as long as they are not breaking changes.
- Patch releases and minor releases get their own groups
- Any major release gets its own PR

image

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,18 @@ updates:
     commit-message:
       # Use a conventional commit tag
       prefix: "chore(deps-rs)"
+    groups:
+      hugr:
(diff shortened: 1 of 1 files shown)


ss2165 (approved):
👍

PR #482, opened 2024-07-15
feat: Add lazify-measure pass

doug-q:


tket2-hseries/src/lazify_measure.rs
```diff
@@ -0,0 +1,284 @@
...
+use crate::extension::{
+    futures::FutureOpBuilder,
+    quantum_lazy::{self, LazyQuantumOpBuilder},
+};
+
+/// A `Hugr -> Hugr` pass that replaces [tket2::Tk2Op::Measure] nodes with
+/// [quantum_lazy::LazyQuantumOp::Measure] nodes. To construct a `LazifyMeasurePass` use
+/// [Default::default].
+#[derive(Default)]
+pub struct LazifyMeaurePass(ValidationLevel);
```
mark-koch, line 32:
```suggestion
pub struct LazifyMeasurePass(ValidationLevel);
```

tket2-hseries/src/lazify_measure.rs
```diff
@@ -0,0 +1,284 @@
...
+
+type Error = Box<dyn std::error::Error>;
+
+impl LazifyMeaurePass {
+    /// Run `LazifyMeasurePass` on the given [HugrMut]. `registry` is used for
+    /// validation, if enabled.
+    pub fn run(&self, hugr: &mut impl HugrMut, registry: &ExtensionRegistry) -> Result<(), Error> {
+        self.0
+            .run_validated_pass(hugr, registry, |hugr, _validation_level| {
+                // TODO: if _validation_level is not None, verify no non-local edges
```
cqc-alec, line 42: Can we do this now?

tket2-hseries/src/lazify_measure.rs
```diff
@@ -0,0 +1,284 @@
...
+        SimpleReplacement::new(g, replacement_hugr, nu_inp, nu_out),
+    )
+}
+
+impl WorkItem {
+    fn work(self, hugr: &mut impl HugrMut) -> Result<impl IntoIterator<Item = Self>, Error> {
+        match self {
+            Self::ReplaceMeasure(node) => {
+                // for now we read immeidately, but when we don't the first
+                // result are the linked inputs we must return
```
cqc-alec, line 179:
```suggestion
                // for now we read immediately, but when we don't the first
                // results are the linked inputs we must return
```

tket2-hseries/src/extension/quantum_lazy.rs
```diff
@@ -0,0 +1,159 @@
...
+)]
+#[allow(missing_docs)]
+#[non_exhaustive]
+pub enum LazyQuantumOp {
+    Measure,
+}
+
+impl MakeOpDef for LazyQuantumOp {
+    fn signature(&self) -> SignatureFunc {
+        FunctionType::new(QB_T, vec![QB_T, future_type(BOOL_T)]).into()
```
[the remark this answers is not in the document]
cqc-alec: I guess the advantage of using a `match` statement now (even if it only has one case) is that we'll get a compiler error if we don't update it when we add new ones.

cqc-alec (approved):
LGTM

PR #484, opened 2024-07-15
fix: Recompile ecc sets after #441

### LINKED ISSUE #483: Default rewrite set for badger does not match any pattern
#441 changed the metadata used for in the patterns, but didn't update the included `.rwr` rewriter files.
This causes `badger` to not match anything when used with the default optimisation rules.

Recompiling the ecc sets should fix this.

aborgna-q:
The sets had to be updated after the changes in #441. Fixes #483


lmondada (approved):
Should we maybe add a test that checks that `badger_pass` does "something", e.g. put a simple CX + CX circuit and see if badger cancels them out.

aborgna-q:
`nam_6_3` takes `>100ms` to load on CI, so it's a bit annoying to use in a test.
But we may be able to test `small_eccs`

PR #496, opened 2024-07-18
feat: Add timeout options and stats to Badger

lmondada:
I'm adding a couple of statistics and a new timeout option to badger. I'd like to use these in a paper I am writing -- having them upstreamed would make reproducing these results easier.

That being said, if you are against tracking these in every future run of badger I am happy to close this without merging.


tket2-py/src/optimiser.rs
```diff
@@ -61,6 +61,12 @@ impl PyBadgerOptimiser {
     ///     If `None` the optimiser will run indefinitely, or until `timeout` is
     ///     reached.
     ///
+    /// * `max_circuit_cnt`: The maximum number of circuits to process before
+    ///     stopping the optimisation.
```
aborgna-q, line 65: Mention that it doesn't apply to multithreaded badger. Maybe log a warning in `badger_multithreaded` if `max_circuit_cnt.is_some()` ?

aborgna-q (approved):

aborgna-q:
It's a cheap operation, so LGTM

lmondada:
I've actually realised the timeout was also implementable in the multi-threading code, so I've added that and removed the warning.

PR #498, opened 2024-07-19
feat!: Compress binary ECCs using zlib

### LINKED ISSUE #488: Compress the ECC sets by default
We should use something like [`zstd`](https://crates.io/crates/zstd) to compress/decompress the rewrite sets, instead of storing the full files.

The included `nam_6_3.rwr` is 70MB, but compresses with `zstd -9` to 4.4MB.

Note that the file already gets compressed when packaging the python wheels, so #486 is still necessary.

aborgna-q:
Breaking change for the compiled `.rwr` ECC sets. The format is now compressed using `zstd`.

Comparative figures for the `nam_6_3.rwr` included in the python package:

- File size reduced from `69MB` to `4.4MB`.
- Load time on my machine reduced from `~560ms` to `~420ms`.
- (offline) ECC compilation time increased from `2.4s` to `3s`.
  This is using the highest compression rate. Faster encodings produce files of around `5MB`.

Closes #488

BREAKING CHANGE: `.rwr` ECC files generated with older versions are no longer supported. Please recompile them, or compress the file with `zstd`.


aborgna-q:
I tried both `zstd` and `flate2` (gzip encoding) here. The differences were minimal;
`flate2` was slower when using high compression levels and produced slightly bigger files.

It seems the main bottleneck when loading ECCs is the rmp deserialisation, so the choice of compression library does not change much.

aborgna-q:
I added a test for the methods that write and read from a buffer.
The file-reading methods are just wrappers over that, and I'd prefer to avoid fiddling with temp files if possible.

PR #495, opened 2024-07-18
docs: Improved multi-threading docs for Badger

lmondada:

PR #505, opened 2024-07-24
test: Add coverage for Badger split circuit multi-threading

lmondada:
Coverage checks for #495 showed a hole in coverage for the Badger optimiser, when using multi-threading with circuit splitting. This PR fixes this.


ss2165 (approved):
nice, thanks

PR #506, opened 2024-07-24
feat: Expose advanced Badger timeout options to tket2-py

lmondada:
#496 introduced a new timeout criterion for Badger based on the number of circuits seen during optimisation. This was not consistently exposed to the Python API.

Fly-by: The `progress_timeout` timeout criterion that was added in #259 was also not exposed properly. This PR also fixes this.

The timeout option `max_circuit_cnt` is now called `max_circuit_count`.


lmondada:
I see I'm missing the arguments in one more function signature. Will add it now.

PR #508, opened 2024-07-25
chore!: bump hugr version to 0.10.0

doug-q:
We remove the cli of tket2-hseries, as the upstream API has changed and
we no longer expect to use this.

BREAKING CHANGE: increased minimum required version of hugr to 0.9.0


.gitignore
```diff
@@ -21,3 +21,6 @@ lcov.info
 
 # python hypothesis testing
 .hypothesis
+
+# local cargo config
+/.cargo/config.toml
```
aborgna-q, line 26: We should probably ignore the full directory
```suggestion
/.cargo
```

tket2-hseries/src/cli.rs
aborgna-q, line 1: We are dropping the whole binary?
Is there still a plan to have a `hugr` subcommand later?

tket2/src/serialize/pytket/op/serialised.rs
```diff
@@ -91,13 +88,13 @@ impl OpaqueTk1Op {
...
         }
-        let Some(TypeArg::Opaque { arg }) = custom_op.args().first() else {
+        let Some(TypeArg::String { arg }) = custom_op.args().first() else {
             return Err(serde_yaml::Error::custom(
                 "Opaque TKET1 operation did not have a yaml-encoded type argument.",
             )
             .into());
         };
-        let op = serde_yaml::from_value(arg.value.clone())?;
+        let op = serde_yaml::from_str(arg)?;
```
aborgna-q, line 97:
```suggestion
        let op = serde_json::from_str(arg)?;
```

You will also have to replace `_yaml` by `_json` in the custom error a couple lines above, and `serde_yaml::Error` with `serde_json::Error` in the `OpConvertError` definition on `serialize/pytket.rs:213`

aborgna-q: After that we should be able to drop the `serde_yaml` dependency.

aborgna-q:
We also need to recompile the included ECC sets because of the breaking serialisation version bump.

Cargo.toml
aborgna-q, line 1: Drop `serde_yaml` from here too

aborgna-q (approved):
LGTM 👍

PR #509, opened 2024-07-25
chore: Replace build.rs with lint config in Cargo.toml

aborgna-q:
Rust 1.80 [just dropped](https://blog.rust-lang.org/2024/07/25/Rust-1.80.0.html), adding a lint option that allows us to avoid the `build.rs` definition added in https://github.com/CQCL/tket2/pull/386.
(The nightly lint got updated before hitting beta, so there shouldn't be compatibility problems there).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ repository = "https://github.com/CQCL/tket2"
 license = "Apache-2.0"
 
 [workspace.lints.rust]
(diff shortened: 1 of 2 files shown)


aborgna-q:
The lint never dropped in stable, so things will still pass there.
The CI error only appeared in the nightly checks, before the lint config was added.

PR #514, opened 2024-07-29
chore: Add script to recompile the ECC sets

aborgna-q:
- Moves the loose ECC files into `test_files/eccs`
- Adds a `just recompile-eccs` command / script to regenerate the `.rwr` files.

Note that the compilation is not deterministic, so we should avoid running it unless needed to avoid bloating the git history.


aborgna-q:
I went with `binary` instead of `-diff`, as it also unsets `merge` and `text`

PR #523, opened 2024-08-02
ci: Avoid building wheels on unrelated releases, split pypi tokens

aborgna-q:
- Stops the wheel building jobs from running on releases of other packages.
  Adding a release for `tket2-eccs` caused `tket2` wheels to be built. Since that's an expensive workflow, this filters it out.
  We continue to build for all platforms on each commit to `main`. We may want to skip some platforms in the future.
  
- Splits the pypi token used for publishing `tket2-py` and `tket2-eccs`.
  Scoped tokens only give permissions to a single project. Splitting the tokens here will let us replace the currently unrestricted token.


lmondada (approved):
Thanks Agustin!

PR #528, opened 2024-08-05
chore: Move parallel evaluation code to `CircuitChunks`

### LINKED ISSUE #266: Move parallel evaluation code to `CircuitChunks`
Badger uses the chunks to split a circuit, optimise each part on a different thread, and join it again afterwards.

This could be a generic operation that can be used in other places too, so it should be added as a method for the chunks (à la `rayon::iter`).

potatoboiler:
resolves #266


tket2/src/optimiser/badger.rs
```diff
@@ -440,7 +441,7 @@ where
         logger.log_best(circ_cost.clone(), num_rewrites);
 
         let (joins, rx_work): (Vec<_>, Vec<_>) = chunks
-            .iter_mut()
```
aborgna-q, line 443: I wasn't sure if this would improve the runtime, since the `map` block below is mostly just about creating new threads. But a quick test showed that this whole `let (joins, rx_works) = ...` block is now a lot faster!

On my laptop, trying multiple shots of
```
cargo run -p badger-optimiser --release -- -e test_files/eccs/nam_6_3.rwr -i 
test_files/barenco_tof_10.json --split-circ -j N
```
while measuring the block time I get

| Threads (`N`) | Block time with `iter_mut` | with `par_iter_mut` | Improvement |
| ------------ | ------- | --------- | -------- |
| 2 | 100ms | 57ms | 0.57 |
| 10 | 719ms | 471ms | 0.65 |

tket2/src/passes/chunks.rs
```diff
@@ -442,6 +444,22 @@ impl CircuitChunks {
...
+        &mut self,
+    ) -> rayon::iter::Map<
+        rayon::slice::IterMut<'_, Chunk>,
+        for<'a> fn(&'a mut Chunk) -> &'a mut Circuit,
+    > {
+        self.chunks
+            .as_parallel_slice_mut()
+            .into_par_iter()
+            .map(Self::par_iter_mut_helper)
+    }
```
aborgna-q, line 462: The helper method is more readable if inlined
```suggestion
    /// Supports implementation of rayon::iter::IntoParallelRefMutIterator
    fn par_iter_mut(
        &mut self,
    ) -> rayon::iter::Map<
        rayon::slice::IterMut<'_, Chunk>,
        for<'a> fn(&'a mut Chunk) -> &'a mut Circuit,
    > {
        self.chunks
            .as_parallel_slice_mut()
            .into_par_iter()
            .map(|chunk| &mut chunk.circ)
    }
```

aborgna-q (approved):
Hi, thanks for the PR! It looks like it improves the _badger optimiser_ startup time quite a lot :)

I just have a small style suggestion.

PR #535, opened 2024-08-13
chore: Update hugr dependency

croyzor:
Update the version of hugr used by the rust and python code.
The main API difference is that extensions must be declared with explicit versions.


pyproject.toml
```diff
@@ -77,7 +77,7 @@ license = { file = "LICENCE" }
 # Note: Be sure to update the dependency versions in [tool.poetry.dependencies] as well
 dependencies = [
     'pytket >= 1.29.2, < 2',
-    'hugr >= 0.5.0, < 0.6',
+    'hugr >= 0.5.0, < 0.7',
```
aborgna-q, line 80: - See the `Note` just above
- Should this be `>= 0.6.0` ?

croyzor: Is the policy to only support one version of hugr? I figured I'd keep 0.5.0 as the minimum bound since it still works, but I'm happy either way

aborgna-q: The main problem rn is serialisation, since we dropped the compatibility assurance.

I'd also check if we don't need to update the rust version too, for the same reason :/
(have we broken serialisation on the `hugr-rs 0.11` release?)

aborgna-q (approved):
👍

PR #545, opened 2024-08-16
ci: Faster coverage tests

aborgna-q:
Copies the idea from `hugr`. 

For the rust checks, the CI ran nightly tests followed by the same tests with coverage instrumentation. This added unnecessary delay to the coverage report.
This PR merges the two steps, so the rust codecov analysis should appear ~2:30 minutes faster.

For python, running with and without coverage takes almost the same time. The PR merges the test runs, but still leaves the codecov upload in a separate job (outside the critical path of the required checks). This saves ~1:30.

Before (Coverage delay ~5mins):
image
After (Coverage delay ~2mins):
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -163,9 +163,9 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        # Pinned nightly version until this gets resolved:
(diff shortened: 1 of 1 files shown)


ss2165 (approved):
thanks for the useful description

PR #554, opened 2024-08-19
feat: Add StaticSizeCircuit

lmondada:
I am starting work on an MVP for Badger v2. This will not be merged into main for a while, as 
i) the performance of `portmatching` is currently significantly worse than the last version
ii) the set of operations and types of circuits that is supported is very limited.

### Description of `StaticSizeCircuit`
This simple `struct` captures circuits with no classical data (and hence also no parametric angle). This will be the starting point to build patterns and compute updateable hashes.

This _could_ in principle be extended in the future to handle more cases, but neither Agustin nor I are convinced that this should ever find its way to `main`... Handling more general cases will require work into fast and updateable hashing.

### DIFF
--- a/tket2/src/static_circ.rs
+++ b/tket2/src/static_circ.rs
@@ -1,0 +1,123 @@
+//! A 2d array-like representation of simple quantum circuits.
+
+mod match_op;
+
+use hugr::{Direction, HugrView};
+pub(crate) use match_op::MatchOp;
+
+use derive_more::{From, Into};
+
+use crate::{circuit::units::filter, Circuit};
+
+/// A circuit with a fixed number of qubits numbered from 0 to `num_qubits - 1`.
+pub(crate) struct StaticSizeCircuit {
+    /// All quantum operations on qubits.
+    qubit_ops: Vec<Vec<StaticOp>>,
+}
+
+impl StaticSizeCircuit {
+    /// Returns the number of qubits in the circuit.
+    #[allow(unused)]
+    pub fn qubit_count(&self) -> usize {
+        self.qubit_ops.len()
+    }
+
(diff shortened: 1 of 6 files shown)


lmondada:
I'm reworking the definition of StaticSizeCirc, I'll let you know when this is ready to review!

tket2/src/static_circ.rs
```diff
@@ -0,0 +1,265 @@
...
+
+use derive_more::{From, Into};
+use thiserror::Error;
+
+use crate::{circuit::units::filter, Circuit};
+
+/// A circuit with a fixed number of qubits numbered from 0 to `num_qubits - 1`.
+#[derive(Clone, Default)]
+pub struct StaticSizeCircuit {
+    /// All quantum operations on qubits.
```
aborgna-q, line 18: Clarify a bit what's each vector about.
```suggestion
    /// All quantum operations on each qubit.
```

tket2/src/static_circ.rs
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    fn get(&self, loc: OpLocation) -> Option<&Rc<MatchOp>> {
+        self.qubit_ops.get(loc.qubit.0)?.get(loc.op_idx)
+    }
+
+    fn exists(&self, loc: OpLocation) -> bool {
+        self.qubit_ops
+            .get(loc.qubit.0)
+            .map_or(false, |ops| ops.get(loc.op_idx).is_some())
```
aborgna-q, line 64:
```suggestion
            .map_or(false, |ops| loc.op_idx < ops.len())
```

aborgna-q (approved):
Looks good 👍

PR #577, opened 2024-09-03
feat!: angle type no longer parametric.

### LINKED ISSUE #576: Angle type should not be parametric
Denominator can just be part of value

aconvert can be removed.

ss2165:
Closes #576

BREAKING CHANGE: Angle type not parametric on log denominator. Aconvert op removed, along with type paramateric helper functions

### DIFF
--- a/tket2/src/extension/angle.rs
+++ b/tket2/src/extension/angle.rs
@@ -1,59 +1,32 @@
-use std::{cmp::max, num::NonZeroU64};
-
+use hugr::extension::prelude::USIZE_T;
+use hugr::extension::simple_op::{MakeOpDef, MakeRegisteredOp};
 use hugr::extension::ExtensionSet;
 use hugr::ops::constant::{downcast_equal_consts, CustomConst};
-use hugr::types::PolyFuncTypeRV;
+use hugr::type_row;
 use hugr::{
-    extension::{prelude::ERROR_TYPE, SignatureError, SignatureFromArgs, TypeDef},
-    types::{
-        type_param::{TypeArgError, TypeParam},
-        ConstTypeError, CustomType, PolyFuncType, Signature, Type, TypeArg, TypeBound,
-    },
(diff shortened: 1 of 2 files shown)


tket2/src/extension/angle.rs
```diff
@@ -129,136 +102,90 @@ impl CustomConst for ConstAngle {
...
         }
+        .into()
+    }
 
-        fn static_params(&self) -> &[TypeParam] {
-            PARAMS
-        }
+    fn description(&self) -> String {
+        match self {
+            AngleOp::atrunc => "truncate an angle to one with a lower log-denominator with the same value, rounding down in [0, 2π) if necessary",
```
cqc-alec, line 145:
```suggestion
            AngleOp::atrunc => "truncate an angle to one with a lower log-denominator with the nearest value, rounding down in [0, 2π) if necessary",
```

tket2/src/extension/angle.rs
```diff
@@ -1,32 +1,26 @@
...
+    CustomType::new_simple(ANGLE_TYPE_ID, TKET2_EXTENSION_ID, TypeBound::Copyable);
 
-pub(super) fn angle_custom_type(extension: &Extension, log_denom_arg: TypeArg) -> CustomType {
-    angle_def(extension).instantiate([log_denom_arg]).unwrap()
-}
-
-fn angle_type(log_denom: u8) -> Type {
-    Type::new_extension(super::angle_custom_type(log_denom))
-}
+/// Dyadic rational angle type  (as [Type])
```
cqc-alec, line 22:
```suggestion
/// Type representing an angle that is a dyadic rational multiple of π (as [Type])
```

ss2165: should it be 2Pi?

cqc-alec: Comes to the same thing.

cqc-alec (approved):

PR #578, opened 2024-09-03
feat!: replace f64 with angle type for tk2 ops

ss2165:
BREAKING CHANGE: Parametric Tk2Ops take angle type rather than float.


aborgna-q (approved):
👍

PR #581, opened 2024-09-04
feat: more angle ops (construct, deconstruct, radians, equality)

ss2165:


tket2/src/extension/angle.rs
```diff
@@ -177,6 +205,13 @@ impl MakeOpDef for AngleOp {
             AngleOp::aadd => "addition of angles",
             AngleOp::asub => "subtraction of the second angle from the first",
             AngleOp::aneg => "negation of an angle",
+            AngleOp::anew => "construct angle from numerator and log-denominator, returning an error if invalid",
+            AngleOp::aparts => "decompose angle into numerator and log-denominator",
+            AngleOp::afromrad => "construct angle from radians, rounding given a log-denominator",
+            AngleOp::atorad => "convert angle to radians",
+            AngleOp::aeq => "check angle equality",
+            AngleOp::amul => "multiply angle by a scalar",
+            AngleOp::adiv => "Divide angle by an integer. If the integer is not a power of 2, or if the resulting denominator would exceed 2^64, the result is rounded to the nearest 2 pi / 2^ 64",
```
mark-koch, line 214:
```suggestion
            AngleOp::adiv => "Divide angle by an integer. If the integer is not a power of 2, or if the resulting denominator would exceed 2^64, the result is rounded to the nearest multiple of 2 pi / 2^ 64",
```
?

mark-koch (approved):

PR #546, opened 2024-08-16
chore: Move `pyproject.toml` to `tket2-py`

aborgna-q:
This should help with `release-please` not detecting changes to `pyproject` as relevant to `tket2-py`.
If things are correct, development commands (maturin / poetry / ...) should run as before.

Closes #541

PR #587, opened 2024-09-05
ci: Fix incorrect wheels name on publish

### LINKED ISSUE #586: ci: Wheels publish job failed due to project name
It seems the python wheels got built with the "tket2-py" name instead of "tket2", so the _Python wheels 🐍_ job failed to push to pypi (the api key is scoped to a single project).
https://github.com/CQCL/tket2/actions/runs/10706181049/job/29685056950

~~This probably got broken with the move to `uv`~~
It broke before that, in #546

aborgna-q:
Fixes #586, which was broken when we moved `pyproject.toml` in #546.

Running `maturin build` from the root dir tries to build the workspace project, creating an invalid wheel name. This PR updates the ci job and the `just build` command to run maturin from inside `/hugr-py/`.

Maturin is still configured as a build tool in the workspace so that `maturin develop` works everywhere. In that case we don't care about the project name.

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -20,6 +20,9 @@ on:
 permissions:
   contents: read
 
+env:
+  MODULE_DIR: tket2-py
+
 jobs:
   # Check if the tag matches the package name,
   # or if the workflow is running on a non-release event.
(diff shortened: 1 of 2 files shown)


ss2165 (approved):
thanks

aborgna-q:
Manually run workflow:
```
📦 Built wheel for CPython 3.10 to dist/tket2-0.3.0-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```
https://github.com/CQCL/tket2/actions/runs/10719118770/job/29722515638#step:4:570
(that is, the wheel name got fixed)

I will manually re-run the job for release 0.3.0 once this is merged

PR #590, opened 2024-09-05
ci: Fix wheels upload on different dir

### LINKED ISSUE #586: ci: Wheels publish job failed due to project name

aborgna-q:
Fixes the fix from #587.

I edited the workflow to run in the PR, so we can test it works (without an api key) 

Closes #586

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -16,6 +16,10 @@ on:
     types:
       - published
   workflow_dispatch:
+  # TODO: Remove this testing trigger
+  pull_request:
+    branches:
+      - main
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


aborgna-q:
This should work now. The workflow succeeded up to actually trying to upload to pypi (and failed due to the missing api token)
https://github.com/CQCL/tket2/actions/runs/10720469490/job/29727314966?pr=590#step:3:46

PR #592, opened 2024-09-05
fix!: remove TryFrom for extension ops use `cast`

ss2165:
BREAKING CHANGE: TryFrom implementations for extension op structs removed, use `cast`


aborgna-q (approved):
👍

PR #611, opened 2024-09-13
feat!: simplify angle extension in to a half turns rotation type

### LINKED ISSUE #605: `AngleOp::adiv` is ambiguous
Angles are equivalence classes mod 2π; but division by an integer (in contrast to multiplication by an integer) is not invariant under that equivalence: e.g. 0 ≡ 2π but 0/2 ≢ (2π)/2.

I would rather like to get rid of this operation for this reason, but if we do keep it we should specify the semantics, probably by coercing the initial argument to be in [0, 2π) or [-π, π). (Note that if we chose [0, 2π) then dividing e.g. the angle -π/2 by 2 would give us 3π/4, which may be unexpected.)

### LINKED ISSUE #608: Angle should always assume maximum precision
Then a single u64 to specify the numerator is isomorphic. 
No need to track lower log_denom.
Smallest denominator rational calculation can be performed by library functions.

### LINKED ISSUE #610: Simplify angle extension to rotations
Rotation defined by number of half turns
(from/to) half turn ops

temporary addition op to support badger and pytket conversions

ss2165:
Closes #610
Closes #608
Closes #605

BREAKING CHANGE: "tket2.angle" extension replaced with "tket2.rotation" extension with rotation type and simplified set of operations.

### DIFF
--- a/tket2-py/tket2/extensions/_json_defs/tket2/rotation.json
+++ b/tket2-py/tket2/extensions/_json_defs/tket2/rotation.json
@@ -16,10 +16,10 @@
   },
   "values": {},
   "operations": {
-    "fromturns": {
+    "from_halfturns": {
       "extension": "tket2.rotation",
-      "name": "fromturns",
-      "description": "Construct rotation from number of turns (would be multiples of π in radians).",
+      "name": "from_halfturns",
+      "description": "Construct rotation from number of half-turns (would be multiples of π in radians).",
       "signature": {
         "params": [],
         "body": {
@@ -92,10 +92,10 @@
       },
       "binary": false
     },
-    "toturns": {
+    "to_halfturns": {
       "extension": "tket2.rotation",
-      "name": "toturns",
-      "description": "Convert rotation to number of turns (would be multiples of π in radians).",
+      "name": "to_halfturns",
+      "description": "Convert rotation to number of half-turns (would be multiples of π in radians).",
       "signature": {
         "params": [],
         "body": {


tket2/src/extension/rotation.rs
```diff
@@ -0,0 +1,285 @@
...
+    ) -> Result<Self, hugr::extension::simple_op::OpLoadError>
+    where
+        Self: Sized,
+    {
+        hugr::extension::simple_op::try_from_name(op_def.name(), op_def.extension())
+    }
+
+    fn signature(&self) -> hugr::extension::SignatureFunc {
+        match self {
+            RotationOp::fromturns => {
```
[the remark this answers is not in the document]
ss2165: Yes sorry I forgot we discussed this, feel free to make the change

tket2/src/extension/rotation.rs
```diff
@@ -0,0 +1,289 @@
...
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct ConstRotation {
+    half_turns: f64,
+}
+
+impl ConstRotation {
+    /// The constant π
+    pub const PI: Self = Self::new_unchecked(1.0);
+    /// The constant 2π
+    pub const TAU: Self = Self::new_unchecked(2.0);
```
cqc-alec, line 51: Not sure how useful this one is. More useful might be:
```rust
    pub const ZERO: Self = Self::new_unchecked(0.0);
```

tket2/src/extension/rotation.rs
```diff
@@ -0,0 +1,289 @@
...
+                type_row![ROTATION_TYPE],
+            ),
+        }
+        .into()
+    }
+
+    fn description(&self) -> String {
+        match self {
+            RotationOp::fromturns => {
+                "Construct rotation from number of turns (would be multiples of π in radians)."
```
cqc-alec, line 152:
```suggestion
                "Construct rotation from number of half turns (would be multiples of π in radians)."
```

tket2/src/extension/rotation.rs
```diff
@@ -0,0 +1,289 @@
...
+            }
+            RotationOp::toturns => {
+                "Convert rotation to number of turns (would be multiples of π in radians)."
```
cqc-alec, line 155:
```suggestion
                "Convert rotation to number of half turns (would be multiples of π in radians)."
```

tket2/src/extension/rotation.rs
```diff
@@ -0,0 +1,289 @@
...
+        }
+        .to_owned()
+    }
+
+    fn extension(&self) -> hugr::extension::ExtensionId {
+        ROTATION_EXTENSION_ID
+    }
+
+    // TODO constant folding
+    // https://github.com/CQCL/tket2/issues/405
```
cqc-alec, line 167: Maybe need to update this issue.

cqc-alec (commented):
Just minor comments.

On a bike-shedding note, not completely happy with the name "rotation" as it suggests an operation rather than a quantity, but I can't think of a better alternative to "angle".

cqc-alec (approved):

PR #615, opened 2024-09-23
chore: Don't `cargo test` tket2-py

aborgna-q:
`tket2-py` does not have rust tests.
Running `cargo test --workspace` on a clean macos installation fails with linking errors due to pyo3 trying to link python libs.

```
note: ld: warning: search path '/install/lib' not found
          ld: library 'python3.12' not found
          clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

This seems to be a known [issue with pyo3](https://pyo3.rs/v0.13.2/faq#i-cant-run-cargo-test-im-having-linker-issues-like-symbol-not-found-or-undefined-reference-to-_pyexc_systemerror).

This PR removes the `--workspace` flag on the local test commands. We'll still test the workspace's default members.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -72,7 +72,7 @@ repos:
       - id: cargo-test
         name: cargo test
         description: Run tests with `cargo test`.
-        entry: uv run -- cargo test --all-features --workspace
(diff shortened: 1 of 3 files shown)


aborgna-q:
Merged as part of #616

PR #618, opened 2024-09-23
feat: Support encoding float and sympy ops

aborgna-q:
So we can encode circuits that either encode sympy expressions, or use the float extension to manipulate the rotations.

Note that this does not yet support `RotationOps::from_halfturns`, since that returns an `Option<Rotation>` since unwrapping that requires control flow operations, so the parameter encoder would need to do constant propagation with parameters across control flow...

drive-by: Mention that `RotationOps::from_halfturns` is fallible in its description. This required a non-breaking update to the extension's `json` definition.

### DIFF
--- a/tket2/src/serialize/pytket/encoder.rs
+++ b/tket2/src/serialize/pytket/encoder.rs
@@ -5,6 +5,9 @@ use std::collections::{HashMap, HashSet, VecDeque};
 
 use hugr::extension::prelude::{BOOL_T, QB_T};
 use hugr::ops::{OpTrait, OpType};
+use hugr::std_extensions::arithmetic::float_ops::FloatOps;
+use hugr::std_extensions::arithmetic::float_types::FLOAT64_TYPE;
+use hugr::types::Type;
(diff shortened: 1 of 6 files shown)


tket2/src/serialize/pytket/encoder.rs
```diff
@@ -619,8 +632,28 @@ impl ParameterTracker {
...
+                {
+                    s
+                } else if let Some(s) = optype
+                    .cast::<SympyOp>()
+                    .and_then(|op| self.encode_sympy_op(&op, inputs.as_slice()))
+                {
+                    s
+                } else {
+                    return Ok(false);
+                }
```
lmondada, line 656: Any way this could be some sort of `match` statement on the optype type instead of trying each out?

lmondada (approved):
LGTM!

tket2/src/serialize/pytket/encoder.rs
aborgna-q: Each cast has a different type, so all options I tried for writing this look messy.

PR #624, opened 2024-09-30
chore!: Replace thiserror with derive_more 1.0

aborgna-q:
`derive_more` 1.0 can now be used as a more granular replacement for `thiserror`.
The main changes required were:

- Explicit derives for `Display` and `From`, in addition to `Error`.
- Use `#[display(...)]` directives instead of `#[error(...)]`. The syntax is sightly different for referencing tuple elements (`_0` vs `0`). `error(transparent)` is also the default behaviour.
- The new `From` derive is less intrusive (derives it for every variant non-`from(ignore)`d, unless explicit variants are picked with `#[from]`).

Those changes are non-breaking (see first commit). Since the next release will be a breaking one, I included some cleanups too.

- Added all missing `#[non_exhaustive]`s to the error enums (this missing is what caused the last breaking change).
- Converted unary tuple variants into named structs.
  `derive_more` assumes unary tuples to be an error source, so we had to add `#[error(ignore)]` in some cases.
- Add more contexts to some errors (e.g. adding the node id to errors that only printed the port offset).

BREAKING CHANGE: Made all errors `non_exhaustive`, and renamed some fields for clarity.


cqc-alec (approved):
LGTM

PR #625, opened 2024-10-01
feat(badger): `cx` and `rz` const functions and strategies for `LexicographicCostFunction`

lmondada:
A couple of points to note:
1. I've made some minor breaking changes to the Rust API of `LexicographicCostFunction`. I think it is cleaner now.
2. I had the choice between keeping `fn` pointers as the cost function type within `LexicographicCostFunction` or moving to `Box<Fn>`. I've stuck to the former for the moment, but I didn't figure out a simple way to reuse the same code for `Tk2Op::CX` and `Tk2Op::RzF64` without using closures. The current code has some duplication as a result, but I think it's bearable.
3. I've tried running badger with `cost_fn='rz'`, but the Rz gate count does not decrease at all. I've looked for an obvious bug but I don't think it is within these changes...

Let me know if you disagree with 1. or 2 and what you think we should do about 3.

---

### Changelog metadata

BEGIN_COMMIT_OVERRIDE
feat: `BadgerOptimiser.load_precompiled`, `BadgerOptimiser.compile_eccs` and `passes.badger_pass` now take an optional `cost_fn` parameter to specify the cost function to minimise. Supported values are `'cx'` (default behaviour) and `'rz'`.

END_COMMIT_OVERRIDE


lmondada:
Update: Running the above script on a different input _does_ work, for example using [qf21_n15](https://github.com/CQCL/pytket-benchmarking-store/blob/benchpress-circuits/benchmarking_circuits/benchpress/qasmbench-medium/qf21_n15.json) and `timeout=10` results in a Rz count decrease from 166 to 149.

(Make sure to rebase to CX + H + Rz first)

aborgna-q:
I edited the changes to make them non-breaking (deprecated the renamed method instead)

aborgna-q (approved):

PR #640, opened 2024-10-08
feat: Add `tket2.rotation.from_halfturns_unchecked` op

### LINKED ISSUE #630: Replace `to_halfturns` with unsafe (infallible) op

aborgna-q:
This is an operation that we are able to process while encoding pytket circuits.
We expect guppy to emit this instead of adding explicit panics to unwrap the fallible result.

The pytket encoder should now support float wires.

Closes #630
I'll open a followup issue on guppy

### DIFF
--- a/tket2-py/tket2/extensions/_json_defs/tket2/rotation.json
+++ b/tket2-py/tket2/extensions/_json_defs/tket2/rotation.json
@@ -55,6 +55,36 @@
       },
       "binary": false
     },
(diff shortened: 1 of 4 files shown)


aborgna-q:
Note: The int ops use a `_checked` for the fallible ops, and no-suffix for the panicking ones.
We could do the same here, but it would be a breaking change.

ss2165 (approved):
👍

PR #644, opened 2024-10-09
feat: Decode pytket op parameters

### LINKED ISSUE #637: Translate tk1 logical expressions
A prototype that does this by calling into python and using sympy: #635

aborgna-q:
This is an alternative to #635, using pure-rust instead of calling to python so we can keep the pytket decoder on the `tket2` crate.

Improves parsing of pytket operation parameters by defining a grammar and parser for sympy expressions using `pest` (based on the [calculator example](https://pest.rs/book/examples/calculator.html) for infix operation precedence).

- Unrecognized operations are still put inside an opaque `SympyOp`, but that should be easy to change in the future.
- I tested multiple sympy expressions to ensure we are able to parse them, but unrecognized ones will also fallback to `SympyOp`.

This is still missing routing parameters to hugr inputs (#628), as it is blocked by https://github.com/CQCL/hugr/issues/1562.

drive-by: Move the pytket parameter encoding/decoding routines to `::serialize::pytket::param::{de,en}code`.

Closes #637.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,6 +51,8 @@ num-complex = "0.4"
 num-rational = "0.4"
 num_cpus = "1.16.0"
 peak_alloc = "0.2.0"
+pest = "2.7.13"
+pest_derive = "2.7.13"
 petgraph = { version = "0.6.3", default-features = false }
 priority-queue = "2.1.1"
 rayon = "1.5"
--- a/tket2/src/circuit.rs
+++ b/tket2/src/circuit.rs
@@ -717,8 +717,6 @@ mod tests {
         assert_eq!(circ.circuit_signature().input_count(), qubits + bits);
         assert_eq!(circ.circuit_signature().output_count(), qubits + bits);
         assert_eq!(circ.qubit_count(), qubits);
-        assert_eq!(circ.num_operations(), 3);
-        assert_eq!(circ.operations().count(), 3);
 
         assert_eq!(circ.units().count(), qubits + bits);
         assert_eq!(circ.nonlinear_units().count(), bits);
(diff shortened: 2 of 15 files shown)


tket2/src/serialize/pytket/encoder.rs
aborgna-q, line 1: The missing code here got moved to `pytket::param::encode`

Cargo.toml
```diff
@@ -51,6 +51,8 @@ num-complex = "0.4"
 num-rational = "0.4"
 num_cpus = "1.16.0"
 peak_alloc = "0.2.0"
+pest = "2.7.13"
+pest_derive = "2.7.13"
```
aborgna-q, line 55: Using `pest` for parsing, as it is the one currently used in `hugr-model`.

tket2/src/circuit.rs
```diff
@@ -717,8 +717,6 @@ mod tests {
         assert_eq!(circ.circuit_signature().input_count(), qubits + bits);
         assert_eq!(circ.circuit_signature().output_count(), qubits + bits);
         assert_eq!(circ.qubit_count(), qubits);
-        assert_eq!(circ.num_operations(), 3);
-        assert_eq!(circ.operations().count(), 3);
```
aborgna-q, line 721: Circuit operation counts are a bit over the place now, as they haven't been updated in a while to reflect the operations we care about.

We'll probably rewrite this in the -future merge, so I'm ignoring the changes for now.

tket2/src/serialize/pytket/param/encode.rs
```diff
@@ -0,0 +1,130 @@
...
+            return Some("pi/2".to_string());
+        } else if const_angle == &ConstRotation::PI_4 {
+            return Some("pi/4".to_string());
+        } else if const_angle == &ConstRotation::TAU {
+            return Some("2pi".to_string());
+        }
+
+        let half_turns = const_angle.half_turns();
+        Some(half_turns.to_string())
+    } else if let Some(const_float) = val.get_custom_value::<ConstF64>() {
```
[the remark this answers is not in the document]
aborgna-q: pytket param expressions are also in half_turns, so no need to convert here.

The problem actually was in the decoding of `pi` constants. That should be decoded to the `3.1415`... float constant rather than a rotation with value `1`.
This lets us write expressions like `2alpha / pi` correctly.

PR #662, opened 2024-10-15
refactor!: Remove deprecated exports

aborgna-q:
Drops some deprecated exports.

drive-by: Remove a couple TODOs about impl iterators.

BREAKING CHANGE: Renamed `LexicographicCostFunction::default_cx` to `LexicographicCostFunction::default_cx_strategy`. Removed sympy constant re-exports in `sympy::extension`.

### DIFF
--- a/tket2/src/rewrite/strategy.rs
+++ b/tket2/src/rewrite/strategy.rs
@@ -349,20 +349,6 @@ impl LexicographicCostFunction<fn(&OpType) -> usize, 2> {
         Self::cx_count().into_greedy_strategy()
     }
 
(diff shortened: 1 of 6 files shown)


tket2/src/rewrite/strategy.rs
```diff
@@ -528,11 +514,11 @@ mod tests {
         let mut circ = n_cx(10);
         let cx_gates = circ.commands().map(|cmd| cmd.node()).collect_vec();
 
-        assert_eq!(circ.rewrite_trace(), None);
```
aborgna-q, line 531: These are just changes to the test cases that expected a `vec`, but now get an `impl Iterator`.

PR #665, opened 2024-10-18
New CircuitPattern, Constraint and Predicate

lmondada:


tket2/src/portmatching/constraint.rs
```diff
@@ -0,0 +1,237 @@
...
+fn unwrap_to_hugr_values<'b, V, B>(args: impl IntoIterator<Item = &'b B>) -> Vec<V>
+where
+    B: Borrow<HugrVariableValue> + 'b,
+    V: TryFrom<HugrVariableValue>,
+    V::Error: Debug,
+{
+    args.into_iter()
+        .map(|arg| {
+            let var = arg.borrow().clone();
+            var.try_into().expect("invalid variable binding type")
```
aborgna-q, line 161: Unsure about the `expect` here, since `Predicate` is a public enum we should fail more gracefully on badly constructed elems.
Would it make sense to just return `false` from `pm::Predicate::check` if the `try_into` fails?

tket2/src/portmatching/indexing.rs
```diff
@@ -68,104 +49,63 @@ pub enum HugrVariableValue {
...
-    },
-    /// A port variable that binds to an incoming port.
-    Incoming {
-        /// The path from the root node to the incoming port.
-        path_from_root: HugrPath,
-    },
+pub struct HugrPortID {
+    /// The node that the port belongs to.
+    pub node: HugrNodeID,
+    /// The port offset.
```
aborgna-q, line 62:
```suggestion
    /// The port index.
```
(it's an offset + a direction flag)

tket2/src/portmatching/indexing.rs
```diff
@@ -68,104 +49,63 @@ pub enum HugrVariableValue {
...
+    pub port: hugr::Port,
 }
 
 impl HugrPortID {
+    pub(super) fn new(node: HugrNodeID, port: hugr::Port) -> Self {
```
aborgna-q, line 67: Why is this not public?
The internals are `pub`, so there's no use in restricting this.
```suggestion
    pub fn new(node: HugrNodeID, port: hugr::Port) -> Self {
```

tket2/src/portmatching/indexing/path.rs
```diff
@@ -301,9 +283,9 @@ fn read_unary(val: usize, start_i: usize, flipped: bool) -> usize {
 impl Debug for HugrPath {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         let mut tuples = vec![];
-        let mut curr = Some(*self);
-        while let Some(curr_some) = curr {
-            let (parent, port, index) = curr_some.uncons();
+        let mut curr = *self;
+        while !curr.is_empty() {
+            let (parent, port, index) = curr.uncons().unwrap();
```
aborgna-q, line 288:
```suggestion
        while let Some((parent, port, index)) = curr.uncons() {
```

tket2/src/portmatching/pattern.rs
```diff
@@ -146,35 +62,387 @@ pub enum InvalidPattern {
...
 }
 
-impl From<NoRootFound> for InvalidPattern {
-    fn from(_: NoRootFound) -> Self {
-        InvalidPattern::NotConnected
+/// Compute the minimum paths from the root to all nodes.
+///
+/// The returned map is a valid [`NodeToPathMap`], i.e. it has a key for every
+/// node in `hugr`.
+fn dijkstra_all_shortest_paths(
```
aborgna-q, line 305: It feels like we should be able to use either portgraph or petgraph for this, but it may not be worth the hassle.

aborgna-q (approved):

tket2/src/portmatching/pattern.rs
lmondada: Yes you are right -- although what makes it difficult is that we keep track of the builder object along the shortest path, so there isn't a clean way to associate each edge with a weight. Rather, each edge corresponds to an "action" on the builder obtained so far.

So as you say, I don't think it's worth the hassle. I've added a comment saying this.

tket2/src/portmatching/constraint.rs
lmondada: I see. I still think a mistyped predicate should be a hard failure (I don't want them to sneak around silently). But you are probably right that I should amend `Predicate::check` to return a Result, and then unwrap this further up the stack, where it would be a proper logic error.

PR #678, opened 2024-10-24
chore(py): Support python 3.13

aborgna-q:
Let's try #653 again.
The [latest `pytket`](https://github.com/CQCL/tket/releases/tag/1.34.0) supports python `3.13`, so we can update here.

We now follow the [general](https://iscinumpy.dev/post/bound-version-constraints/#pinning-the-python-version-is-special) [recommendation](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires) against setting a hard python version bound in the `pyproject.toml`'s `requires-python` field.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,7 @@ env:
   UV_VERSION: "0.4.20"
   UV_FROZEN: 1
   # The highest and lowest supported Python versions, used for testing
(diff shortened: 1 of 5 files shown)


lmondada (approved):
I found it counter intuitive at first that we should put an upper bound on the supported Python version, but the document you linked made good points 👍

PR #682, opened 2024-11-05
feat: move extensions to auxillary package

### LINKED ISSUE #642: Move extension def exports into separate python library
So that both guppy and tket2 can depend on it, without making any cyclic deps.
Since the deps are not expected to suffer breaking changes too often, it should let us do mostly minor releases and keep things in sync.

ss2165:
Closes #642

I will publish, create token etc. once merged.


tket2-exts/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+## 0.1.0 (2024-10-06)
+
+
+### Features
+
+* Add tket2 extension definitions.
+
+## Changelog
```
cqc-alec, line 10: Should this line be here?

tket2-exts/CHANGELOG.md
```diff
@@ -0,0 +1,10 @@
+# Changelog
+
+## 0.1.0 (2024-10-06)
```
cqc-alec, line 3: Date?

aborgna-q (approved):

PR #689, opened 2024-11-06
ci: Fix py coverage not getting uploaded

aborgna-q:
At some point it seems the python coverage step started failing ~~due to some invalid auto-discovered coverage files (I think `download-artifact` creates a directory with the file's name, and we ended up with a `IsADirectoryError`)~~.
https://github.com/CQCL/tket2/actions/runs/11706568650/job/32604131200?pr=680#step:4:57

This change tries to fix that by disabling the automatic discovery, and makes the step fail if the upload fails in the future.

drive-by: Copy `codecov.yml` config from hugr

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,7 +232,7 @@ jobs:
     needs: changes
     if: ${{ needs.changes.outputs.python == 'true' }}
     runs-on: ubuntu-latest
-    name: tests (Python)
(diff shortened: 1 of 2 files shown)


aborgna-q:
Avoiding the `upload-artifact` -> `download-artifact` step fixed it.

Now this is in the critical path for the required checks, but the step only takes ~5s so it should be fine.

cqc-alec (approved):

PR #703, opened 2024-11-26
refactor!: rename hseries to qsystem

### LINKED ISSUE #701: Rename `hseries` to `qsystem`

ss2165:
Closes #701 

Kept the `tket2-hseries` package name the same because it is internal only and already published.

BREAKING CHANGE: all references to "hseries" including extension, modules and structs renamed to "qsystem"


tket2-hseries/src/bin/tket2-hseries.rs
cqc-alec, line 1: Should we rename this file?

ss2165: I thought best not to to align with the crate name, since it defines the cli

cqc-alec (approved):

PR #720, opened 2024-12-12
chore!: Update to next version of hugr

doug-q:
Updates the dependencies to `hugr-rs 0.14` and `hugr-py 0.10`.

Now that hugrs carry their required extensions, we dropped the (internal) `extensions` field from the `Circuit` struct.

BREAKING CHANGE: Updated `hugr` rust dependency to `0.14.0`
BREAKING CHANGE: Extension are now defined in `hugr 0.14` style. Replaced the `ROTATION_TYPE` with a method call, and dropped the per-extension registry definitions.


tket2-py/src/ops.rs
```diff
@@ -216,6 +214,10 @@ impl PyPauliIter {
 }
 
 /// A wrapped custom operation.
+//
+// TODO: These can no longer be constructed from Python. Since `hugr-rs 0.12`
```
ss2165, line 218:
```suggestion
// TODO: These can no longer be constructed from Python. Since `hugr-rs 0.14`
```

tket2/src/circuit.rs
```diff
@@ -44,24 +41,13 @@ pub struct Circuit<T = Hugr> {
     ///
     /// This is checked at runtime to ensure that the node is a DFG node.
     parent: Node,
-    /// An optional set of extensions required to validate the circuit,
-    /// not including the prelude.
-    ///
-    /// Wrapped in an Arc to allow sharing between circuits, specially for borrowed circuits.
-    ///
-    /// Defaults to an standard set of quantum extensions and Hugr's std set.
-    required_extensions: Option<Vec<Arc<Extension>>>,
```
ss2165, line 53: 💥

tket2/src/rewrite/ecc_rewriter.rs
```diff
@@ -175,12 +175,12 @@ impl ECCRewriter {
         Self::load_binary_io(&mut file)
     }
 
-    /// When the ECC gets loaded, all custom operations are an instance of `OpaqueOp`.
-    /// We need to resolve them into `ExtensionOp`s by validating the definitions.
-    fn resolve_extension_ops(&mut self) -> Result<(), OpaqueOpError> {
+    //// When the ECC gets loaded, all custom operations are an instance of `OpaqueOp`.
+    //// We need to resolve them into `ExtensionOp`s by validating the definitions.
```
ss2165, line 179: what's with the extra `/`

ss2165 (approved):

PR #734, opened 2024-12-18
chore: Drop viz_hugr test utils

aborgna-q:
See https://github.com/CQCL/hugr/pull/1789

`webbrowser` now requires `rust 1.81` on linux targets, so we better just drop it.

Also removes a couple unused dependencies detected with `cargo-machete`


ss2165 (approved):
thanks

PR #737, opened 2024-12-27
feat(tket2-hseries): Add `tket2.wasm` extension

doug-q:


tket2-hseries/src/extension/wasm.rs
```diff
@@ -0,0 +1,885 @@
...
+    /// The name of the `tket2.wasm.func` type.
+    pub static ref FUNC_TYPE_NAME: SmolStr = SmolStr::new_inline("func");
+
+    /// The [TypeParam] of `tket2.wasm.lookup` specifying the name of the function.
+    pub static ref NAME_PARAM: TypeParam = TypeParam::String;
+    /// The [TypeParam] of various types and ops specifying the input signature of a function.
+    pub static ref INPUTS_PARAM: TypeParam = TypeParam::List { param: Box::new(TypeBound::Any.into()) };
+    /// The [TypeParam] of various types and ops specifying the output signature of a function.
+    pub static ref OUTPUTS_PARAM: TypeParam = TypeParam::List { param: Box::new(TypeBound::Any.into()) };
+}
```
lmondada, line 122: It might be that cargo fmt-ing this without the macro will break up some of the lines that are looking quite long at the moment.

tket2-hseries/src/extension/wasm.rs
```diff
@@ -0,0 +1,885 @@
...
+            _ => Err(()),
+        }
+    }
+}
+
+/// Fallibly convert a [TypeArg] to a [Type].
+///
+/// This will fail if `arg` is of non-type kind (e.g. String),
+/// or if `arg` contains row variables.
+/// TODO move this to `impl TypeArg` in `hugr-core`
```
lmondada, line 306: Are you leaving this here for the moment because you do not want to make changes to `hugr-core` right now?

tket2-hseries/src/extension/wasm.rs
```diff
@@ -0,0 +1,885 @@
...
+                let outputs = TypeRV::new_row_var_use(2, TypeBound::Copyable);
+
+                let func_type = WasmType::func_custom_type(inputs, outputs, extension_ref).into();
+                PolyFuncTypeRV::new(
+                    [
+                        NAME_PARAM.to_owned(),
+                        INPUTS_PARAM.to_owned(),
+                        OUTPUTS_PARAM.to_owned(),
+                    ],
+                    Signature::new(vec![module_type], vec![func_type]),
```
lmondada, line 397: I'm a noob when it comes to define (parametrised) type signatures, so take this with a grain of salt. I notice that `Self::lookup` and `Self::call` use different ways to define the signatures returned, even though superficially they seem quite similar to me (`Signature::new` vs `FuncValueType::new`, `Type` vs `WasmType` etc)...

It would make it more readable if both code blocks were as similar as possible. If you can't use the same function calls & types for whatever reason, then maybe just put the respective function signatures in a comment -- it took me awhile to parse what they meant and had to refer several times back to the module docs (which were supremely useful btw).

lmondada (approved):
Thanks @doug-q! It's always very fun to read your code, I learn a lot.

I don't have much useful feedback, other than that your design perfectly seems sound!

mark-koch:
The extension design looks good to me 👍

Just to clarify, the `WasmType::Context` is to track global state between wasm calls, right? So it's fine to call `GetContext` multiple times if you don't care about preserving that state?

tket2-hseries/src/extension/wasm.rs
```diff
@@ -0,0 +1,885 @@
...
+//!
+//!  `tket2.wasm.get_context` takes a `prelude.usize`, allowing multiple independent
+//!  contexts to exist simultaneously. `get_context` is fallible, returning
+//!  `None` if the specified context has already been obtained via an earlier
+//!  `get_context` op and not `dispose_context`ed.
+//!
+//!  `tket2.wasm.func` is a type representing a handle to a function in a
+//!  `tket2.wasm.module`. It carries type args defining it's signature, but not
+//!  it's name.
+//!
```
ss2165, line 29:
```suggestion
//!  `tket2.wasm.module`. It carries type args defining its signature, but not
//!  its name.
//!
```

tket2-hseries/src/extension/wasm.rs
```diff
@@ -0,0 +1,885 @@
...
+//!  A `tket2.wasm.func` is obtained from a `tket2.wasm.lookup` op, which takes
+//!  a compile-time name and signature, and a runtime module. TODO Likely the
+//!  module should be compile time here, but I think we need
+//!  extension-op-static-edges to do this properly.
+//!
+//!  `tket.wasm.func`s are called via the `tket2.wasm.call` op. This op takes:
+//!   - a `tket2.wasm.context` identifying where the call will exectue;
```
ss2165, line 36:
```suggestion
//!   - a `tket2.wasm.context` identifying where the call will execute;
```

ss2165 (commented):
Looks good, add to https://github.com/CQCL/tket2/blob/99c7678ba26019d3bbe7c18a98572c5afba4720a/tket2-hseries/src/bin/tket2-hseries.rs#L9-L10 and regenerate extensions?

ss2165:
I would strongly prefer to separate out the breaking change in to a separate commit to main

ss2165 (changes requested):
I would strongly prefer to separate out the breaking change in to a separate commit to main

tket2-hseries/src/extension/wasm.rs
```diff
@@ -766,6 +770,17 @@ mod test {
         assert_eq!(m1.extension_reqs(), ExtensionSet::singleton(EXTENSION_ID));
     }
 
+    #[rstest]
+    #[case(WasmType::Module)]
+    #[case(WasmType::Context)]
+    #[case(WasmType::new_func(type_row![], type_row![]))]
+    #[case(WasmType::new_func(vec![TypeRV::new_row_var_use(0, TypeBound::Any)], vec![bool_t()]))]
+      fn wasm_type(#[case] wasm_t: WasmType) {
```
ss2165, line 778: looks like a formatting issue

ss2165 (approved):
thanks for the changes

Issue #740, opened 2025-01-08
`LazifyMeasurePass` doesn't work on all ops


doug-q:
We only rewrite `Tket2Op::Measure`.

It happens that we rewrite `Tket2Op::Measure` => `TQsystemOp::Measure` before we lazify, so as is it's a no-op pass.

I suggest we leave the pass ordering as is, and add cases to move to rewriting:
* `Tket2Op::Measure`
* `Tket2Op::MeasureFree`
* `QSystemOp::Measure`
* `QSystemOp::MeasureReset`

Let's simultaneously change `LowerTket2ToQSystemPass` to lower to lazy measures directly (sharing code obviously).

PR #741, opened 2025-01-08
feat(tket2-hseries)!: Redefine `QSystemOp::LazyMeasure` and introduce `QSystemOp::LazyMeasureReset`

doug-q:
Done in the process of addressing #740 

BREAKING CHANGE: The signature of `QSystemOp::LazyMeasure` is changed to consume its qubit.

### DIFF
--- a/tket2-exts/src/tket2_exts/data/tket2/qsystem.json
+++ b/tket2-exts/src/tket2_exts/data/tket2/qsystem.json
@@ -23,6 +23,41 @@
           ],
           "output": [
             {
+              "t": "Opaque",
+              "extension": "tket2.futures",
+              "id": "Future",
+              "args": [
+                {
+                  "tya": "Type",
+                  "ty": {
+                    "t": "Sum",
+                    "s": "Unit",
+                    "size": 2
+                  }
+                }
+              ],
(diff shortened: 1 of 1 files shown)


tket2-hseries/src/lazify_measure.rs
```diff
@@ -101,7 +101,7 @@ lazy_static! {
     static ref MEASURE_READ_HUGR: Hugr = {
         let mut builder = DFGBuilder::new(Signature::new(qb_t(), vec![qb_t(), bool_t()])).unwrap();
         let [qb] = builder.input_wires_arr();
-        let [qb, lazy_r] = builder.add_lazy_measure(qb).unwrap();
+        let [qb, lazy_r] = builder.add_lazy_measure_reset(qb).unwrap();
```
ss2165, line 104: does this mean all measurements are re-written to ones that also reset? A normal `Measure` should become a `LazyMeasure` no?

ss2165: Ah this appears to be addressed by the follow-up PR

ss2165 (approved):
looks good assuming follow up merges imminently

PR #742, opened 2025-01-08
feat(tket2-hseries): Lazify more flavours of measure ops

### LINKED ISSUE #740: `LazifyMeasurePass` doesn't work on all ops

doug-q:
Closes #740

### DIFF
--- a/tket2-hseries/src/lazify_measure.rs
+++ b/tket2-hseries/src/lazify_measure.rs
@@ -250,9 +250,9 @@ impl LazifyMeasureRewrite {
             }
             QSystemOp::LazyMeasureReset => {
                 let expected_signature = Signature::new(qb_t(), vec![qb_t(), bool_t()]);
-                if actual_signature
+                if !actual_signature
                     .as_ref()
-                    .map_or(true, |x| x.io() != expected_signature.io())
+                    .is_some_and(|x| x.io() == expected_signature.io())
                 {
                     Err(LazifyMeasurePassError::InvalidOp {
                         node,


ss2165:
> I think I should try harder to use SimpleReplacement. This would allow using Builders for all this rewiring. What do you think?

Yes agreed

tket2-hseries/src/lazify_measure.rs
```diff
@@ -74,158 +82,244 @@ impl LazifyMeasurePass {
...
+    pub fn try_new_measure(
+        node: Node,
+        hugr: impl HugrView,
+    ) -> Result<Self, LazifyMeasurePassError> {
+        Self::check_signature(node, QSystemOp::LazyMeasure, hugr.get_optype(node))?;
+
+        let subgraph = SiblingSubgraph::from_node(node, &hugr);
+        let uses = {
+            let mut v = hugr.linked_inputs(node, 0).collect_vec();
+            v.sort();
```
ss2165, line 138: similar sort question, AFAIK `linked_inputs` should be deterministic

tket2-hseries/src/lazify_measure.rs
```diff
@@ -74,158 +82,244 @@ impl LazifyMeasurePass {
...
+    fn build_futures_gadget(builder: &mut impl Dataflow, wire: Wire, num_uses: usize) -> Vec<Wire> {
+        let future_wires = if num_uses == 0 {
+            builder.add_free(wire, bool_t()).unwrap();
+            vec![]
+        } else {
+            let mut future_wires = vec![wire];
+            for _ in 1..num_uses {
+                let prev_wire = future_wires.last_mut().unwrap();
+                let [wire1, wire2] = builder.add_dup(*prev_wire, bool_t()).unwrap();
+                *prev_wire = wire1;
```
ss2165, line 229: does this line actually do anything? `prev_wire` doesn't seem to be used after this line

tket2-hseries/src/lazify_measure.rs
```diff
@@ -74,158 +85,246 @@ impl LazifyMeasurePass {
...
+            .unwrap()
+            .signature()
+            .into_owned();
+        let actual_signature = op_to_replace
+            .dataflow_signature()
+            .map(std::borrow::Cow::into_owned);
+        if let Some(sig) = &actual_signature {
+            if expected_signature.input() == sig.input()
+                && expected_signature.output().len() == sig.output().len()
+                && itertools::zip_eq(expected_signature.output().iter(), sig.output().iter())
```
ss2165, line 258: could you add a comment explaining these conditions? types matching up to bool == future(bool) ?

tket2-hseries/src/lazify_measure.rs
```diff
@@ -262,8 +360,6 @@ mod test {
             .unwrap()
             .outputs_arr();
         let mut hugr = builder.finish_hugr_with_outputs([qb]).unwrap();
-
-        assert!(hugr.validate_no_extensions().is_ok());
```
ss2165, line 266: we don't want to validate any more?

tket2-hseries/src/lazify_measure.rs
ss2165: muich

ss2165 (approved):
nice!

tket2-hseries/src/lazify_measure.rs
```diff
@@ -74,158 +79,250 @@ impl LazifyMeasurePass {
...
+        qsystem_op: QSystemOp,
+        op_to_replace: &hugr::ops::OpType,
+    ) -> Result<(), LazifyMeasurePassError> {
+        let actual_signature = op_to_replace.dataflow_signature().map(|x| x.into_owned());
+        match qsystem_op {
+            QSystemOp::LazyMeasure => {
+                let expected_signature = Signature::new(qb_t(), bool_t());
+                if actual_signature
+                    .as_ref()
+                    .map_or(true, |x| x.io() != expected_signature.io())
```
ss2165, line 242:
```suggestion
                if !actual_signature
                    .as_ref()
                    .is_some_and(|x| x.io() == expected_signature.io())
```

?

PR #755, opened 2025-01-21
feat: Add portmatching Predicates and Constraints

lmondada:
wip...


tket2/src/portmatching/predicate.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+/// The valid "signatures" of predicates, useful for argument parsing.
+enum PredicateArguments {
+    /// Argument for a [`Predicate::IsOpEqual`] predicate.
+    Node(hugr::Node),
+    /// Arguments for a [`Predicate::IsWireSource`] or [`Predicate::IsWireSink`] predicate.
+    NodeAndWire(hugr::Node, hugr::Wire),
+    /// Arguments for a [`Predicate::IsDistinctFrom`] predicate.
+    ManyNodes(Vec<hugr::Node>),
```
cqc-melf, line 81: Why do you not have an option for many wires?

tket2/src/portmatching/predicate.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    fn wire_pred(port: hugr::Port) -> Predicate {
+        match port.as_directed() {
+            Either::Left(in_port) => Predicate::IsWireSink(in_port),
+            Either::Right(out_port) => Predicate::IsWireSource(out_port),
+        }
+    }
+
+    #[rstest]
+    fn test_check_shared_edge(circ_with_copy: Circuit) {
```
cqc-melf, line 244: I don't think this covers all the predicates Are there more tests somewhere else and is there a CI check for coverage?

cqc-melf (commented):
I am not sure I understood everything, so if some of the comments are indicating missing details, please let me know.

tket2/src/optimiser/badger.rs
```diff
@@ -636,7 +636,9 @@ mod tests {
     }
 
     #[rstest]
+    #[ignore = "wip"]
```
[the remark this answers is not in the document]
lmondada: yes 👍  I broke up the update into small PRs. These tests will be reactivated by some of the next PRs, before being merged into main

tket2/src/portmatching/predicate.rs
lmondada: I only added the variants that I needed, given the predicates and their signatures. I will rewrite this a bit to make it clearer

tket2/src/portmatching/predicate.rs
lmondada: Yes! There are more tests (that depend on other parts of the code, so I'll add them in an upcoming PR). The reason coverage tests don't run here is because this does not merge into `main`. I'll run coverage tests properly before the final merge.

lmondada: That said, you are right: I will add a test for other predicates

cqc-melf (approved):
Looks good

PR #759, opened 2025-01-27
fix: Use `RemoveDeadFuncsPass` instead of deprecated `remove_polyfuncs()`

### LINKED ISSUE #758: Error from clippy check on CI due to deprecated `remove_polyfuncs`
The clippy check on the CI is failing: https://github.com/CQCL/tket2/actions/runs/12980837714/job/36198388971 .

The suggested fix is to use `hugr::algorithms::dead_funcs::RemoveDeadFuncsPass` instead, but this suggestion is incorrect (probably out of date), as `RemoveDeadFuncsPass` is exported by `hugr_passes`.

So `hugr_passes` needs to be added as a dependency. Even then, it seems that `RemoveDeadFuncsPass` is not a drop-in replacement for `remove_polyfuncs` (different signature).

cqc-alec:
Fixes #758 .


uv.lock
aborgna-q, line 1: Outdated `uv` ?

tket2-hseries/Cargo.toml
```diff
@@ -23,6 +23,7 @@ required-features = ["cli"]
 
 [dependencies]
 hugr.workspace = true
+hugr-passes = "0.14.2"
```
aborgna-q, line 26: - Dependencies should go on the workspace's `Cargo.toml`, and `.workspace = true`'d here.

- We are currently at `hugr 0.14.1` on the workspace file, we should use the same versions all around.

tket2-hseries/src/lib.rs
```diff
@@ -80,7 +81,7 @@ impl QSystemPass {
             self.validation_level.run_validated_pass(hugr, |hugr, _| {
                 let mut owned_hugr = Hugr::default();
                 mem::swap(&mut owned_hugr, hugr);
-                owned_hugr = remove_polyfuncs(owned_hugr);
+                RemoveDeadFuncsPass::default().run(&mut owned_hugr).unwrap();
```
aborgna-q, line 84: The `unwrap` here is annoying, but returning the error requires a release with https://github.com/CQCL/hugr/pull/1883.
We should have a `TODO` here linking to that PR.

https://github.com/CQCL/tket2/pull/753 should fix it later.

aborgna-q (approved):

PR #763, opened 2025-02-05
fix: restore dead funcs pass

doug-q:


aborgna-q (approved):
👍 

We could have the "search main node" as a separate function, but that's just style nitpicking.

At some point we will add "entrypoint" definitions to hugrs. We have this currently in `hugr-py` with the ["pointed packages" protocols](https://github.com/CQCL/hugr/pull/1573), but not in rust nor in the model.
See some discussion over here: https://github.com/CQCL/hugr/discussions/1642
If that comes to be we should be able to modify this callsite non-breakingly.

PR #773, opened 2025-02-14
feat: Add CircuitDiff

lmondada:
I'm introducing `CircuitDiff`s, for persistent graph rewriting! Once done, this will speed up badger-like workflows significantly (see below).

`CircuitDiff`s, which could be renamed to `CircuitPatch`es if you prefer, store a subgraph for each of their parent diffs, along with the new replacement graph to rewrite the subgraphs to. Currently, diffs with no parents are created from a circuit, whilst other diffs are created by applying a `SimpleReplacement` onto a parent diff. This means that right now, only diffs with a single parent can be constructed (see below)

## How speedup?
Given rewrites A, B and C with disjoint support, instead of having to consider individual graphs with respectively rewrites A, B, C, AB, BC, AC and ABC applied, the three rewrites can be applied independently and there will only be `CircuitDiff` for A, B and C. At pattern matching time, the matcher will have to consider the various combinations that are possible, but only within the local region that matches overlap with.

## What's next
- Implement a `CircuitHistory`: a set of diffs that are compatible with each other (i.e. there are no conflicting rewrites). Then rewrites can be applied to histories, which will create diffs with multiple parents. I will try to implement `HugrView` on these
- Dominator trees and hashing: once we have a dag of diffs, I will keep pointers to the dominator of each, diff, this will make checking for compatible rewrites faster. We can then also keep the hashes of all dominated diffs in the dominator, to exclude create the same diffs multiple times, without relying on global hashes


lmondada:
> This PR contains breaking changes to the public Rust API. Please deprecate the old API instead (if possible), or mark the PR with a `!` to indicate a breaking change.
> 
> cargo-semver-checks summary

I've double checked, there are no breaking changes once commit [f109186](https://github.com/CQCL/tket2/pull/773/commits/f109186d28beb4eaf9ff8d7ac96cfca2e57234ef) is reverted. But it will require [hugr#1920](https://github.com/CQCL/hugr/pull/1920) to be merged, along with a hugr release.

aborgna-q:
Fyi the `semver-checks` failure is due to it ignoring any cargo patches, so it tries to compile with `hugr = "0.14.3"` instead.

PR #778, opened 2025-02-18
feat: Add CircuitHistory

lmondada:
I'm putting this out here for discussion. My main question: **how should I expose the hugr-like interface of `CircuitHistory`?**. I've proposed an implementation of `HugrView` on a wrapper type `ExperimentalHugrWrapper` of the circuit history. See `src/diff/experimental.rs`. 

A `CircuitHistory` is always equivalent to a valid Hugr, but implementing `HugrView` is difficult because
- there is no "base hugr", so implementing `HugrInternals` is impossible. This could be implemented if the base hugr was node specific, i.e. `fn base_hugr(&self, node: Node) -> &Hugr`
- the `hugr::Node` type is fixed, but in a CircuitHistory the "natural" node index is a pair `(Hugr, Node)`, i.e. an index/ref to the correct hugr in the history, along with a node within it. Right now, I have "solved" this with the world's greatest hack, using the integer within `Node` to encode both a hugr index and the node index. This will overflow at `2^16` which we are bound to reach pretty quickly
- Enumerating all nodes and counting them requires traversing the hugr, so this is a slow operation. Same with edge_count

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -946,7 +946,7 @@ dependencies = [
 [[package]]
 name = "hugr"
 version = "0.14.3"
-source = "git+https://github.com/CQCL/hugr?rev=0efc806f#0efc806f99888e64a0da6f639e4dbb83afebd4bf"
+source = "git+https://github.com/CQCL/hugr?rev=739cf944#739cf94404d45622ee196b1c68de6bec6a2b28d8"
 dependencies = [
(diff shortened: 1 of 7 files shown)


lmondada:
This is being moved to hugr, see https://github.com/CQCL/hugr/pull/2080

PR #787, opened 2025-02-28
chore: Update to pytket 2

cqc-alec:


aborgna-q:
`pytket 2.0.0` changelog:

```
API changes:

* Remove `GlobalisePhasedX` transform and pass.
* Remove `SynthesiseUMD` pass.
* Remove `NoiseAwarePlacement.modify_config` and `GraphPlacement.modify_config`
  methods.
* Remove `ClassicalExpBox` and related methods.

Fixes:

* Fix segfault listing registers of circuit containing unindexed units.
```

I gather that the serialization format hasn't changed, so we shouldn't need to update `tket-json-rs` ?

aborgna-q:
As long as no things are removed we should be fine on the `tket-json-rs` side. We could maybe deprecate them to avoid people writing new code that generates those ops, but it's better to keep the backwards compatibility.

On the `tket2` side we do not generate those ops, so no new changes should be needed.

aborgna-q (approved):

Issue #789, opened 2025-03-03
tket2-hseries: add extension op for "order" in the utils extenstion


narmlu:
Analogous to #767 but for the "order" or "order16". Takes 16 qubits as parameters and returns nothing.

PR #792, opened 2025-03-05
feat(tket2-hseries)!: Add order_in_zones extension op

narmlu:
Closes: #789

BREAKING CHANGE:
This adds an entry to the `UtilsOp` enum. Now it looks like this:
```
tket2_hseries::extension::utils
pub enum UtilsOp {
    GetCurrentShot,
    OrderInZones,
}
```


narmlu:
Updated the description and yes, I like the idea about hoisting 16 into a constant. I just put it into `ORDER_LENGTH` at the top of the file.

PR #794, opened 2025-03-06
chore(tket2-py): bump tket2_exts constraint

qartik:


aborgna-q (commented):
We re-export all the definitions from `tket2_exts`, so this is a breaking change too...
(We can reuse the same breaking change notice).

https://github.com/CQCL/tket2/blob/main/tket2-py/tket2/extensions/__init__.py

aborgna-q (approved):

PR #795, opened 2025-03-06
chore(tket2-exts): Force non-breaking python release

aborgna-q:
#789 added a breaking change to the rust crate, but the breaking flag also affects the python side since `tket2-exts` got modified.

This empty change tells `release-please` to ignore that and do a non-breaking version bump on the next release PR.

The magic is in having a commit on the tree containing this comment:

Release-As: 0.5.1


qartik (approved):

aborgna-q:
I added a COMMIT_OVERRIDE instead on #792, so this is no longer needed

PR #797, opened 2025-03-06
fix(tket2-hseries)!: rm unneeded ext, make `UtilsOp` enum `non_exhaustive`

qartik:
Revert #792 

BREAKING CHANGE: Make `UtilsOp` enum `non_exhaustive` so that future additions are not considered API-breaking.

### DIFF
--- a/tket2-hseries/src/extension/utils.rs
+++ b/tket2-hseries/src/extension/utils.rs
@@ -7,11 +7,11 @@ use derive_more::derive::Display;
 use hugr::{
     builder::{BuildError, Dataflow},
     extension::{
-        prelude::{qb_t, UnwrapBuilder},
+        prelude::UnwrapBuilder,
         simple_op::{try_from_name, MakeOpDef, MakeRegisteredOp},
         ExtensionId, ExtensionRegistry, ExtensionSet, OpDef, SignatureFunc, Version, PRELUDE,
     },
-    std_extensions::{arithmetic::int_types::int_type, collections::array::array_type},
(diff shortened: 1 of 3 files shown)


tket2-hseries/src/extension/utils.rs
```diff
@@ -61,12 +63,15 @@ lazy_static! {
     EnumString,
     Display,
 )]
+#[non_exhaustive]
```
qartik, line 66: @aborgna-q does this attribute only take effect for semver checks in the next PR that adds to `UtilsOp`. Is there another way to mark this PR as not breaking the public Rust API?

tket2-hseries/src/extension/utils.rs
```diff
@@ -77,6 +82,10 @@ impl MakeOpDef for UtilsOp {
                 array_type(ORDER_LENGTH, qb_t()),
                 array_type(ORDER_LENGTH, qb_t()),
             ),
+            UtilsOp::RPC => Signature::new(
+                array_type(RPC_LENGTH, int_type(6)),
+                array_type(RPC_LENGTH, int_type(6)),
+            ),
```
qartik, line 88: It might be nice to not restrict the array length here.

tket2-hseries/src/extension/utils.rs
aborgna-q: Adding the `non_exhaustive` flag is a breaking change on itself, since dependents are now required to have a default branch when matching the enum.

After this, adding new variants becomes non-breaking since they can always match that default branch.

PR #813, opened 2025-03-14
feat: Use hugr envelopes to store/load circuits

aborgna-q:
Deprecates the Circuit's "read/write package" API in favour of using the new HUGR envelopes.

#806 modified these to make them work with envelopes as a hack. This PR rolls those changes back and instead adds new methods for these.

I also renamed some hugr storing methods, so now these all have the same names as their `hugr` counterparts.

##### Rust side
- `store` / `store_str` / `store_hugr` to write circuits
- `load_function` / `load_function_str` / `load_hugr` to read circuits.
  - In the future we may add `load` methods that do not need the extra `function_name` attribute, once we have hugr root pointers.
  
  ##### Python side
  - `to_bytes` / `to_str`
  - `from_bytes` / `from_str` (with optional function name parameters)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,12 @@ repository = "https://github.com/CQCL/tket2"
 license = "Apache-2.0"
 
 [workspace.lints.rust]
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(ci_run)'] }
+unexpected_cfgs = { level = "warn", check-cfg = [
+    # Set by our CI
+    'cfg(ci_run)',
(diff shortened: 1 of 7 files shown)


aborgna-q:
Adding a deprecation marker is a semver failure for `semver-checks`, [but only requires a minor version bump](https://www.github.com/obi1kenobi/cargo-semver-checks/issues/57).
I'll have to update the semver-checks-action at some point to not fail on those, since I don't think they should require a conventional commit breaking flag.

PR #818, opened 2025-03-17
chore: Disable ordered_qalloc test in miri

doug-q:


ss2165 (approved):

aborgna-q:
We are not currently running `miri` on this repo. Do you think we should do so?

We have a single local **unsafe** trait impl for `bytemuck::TransparentWrapper`, though we can always test that the dependencies are behaving.

aborgna-q:
You're right!
I should add some just recipes for local dev and update that workflow 🤔

PR #823, opened 2025-03-20
feat: Add `tket2.bool` extension

### LINKED ISSUE #817: Add opaque bool extension to tket2
Need a type and a `CustomConst`.

We need `to` and `from` ops.

We may have:
* eq
* and/or/not/xor

tatiana-s:
Closes #817, enables https://github.com/quantinuum-dev/hugrverse/issues/130


tket2-exts/src/tket2_exts/data/tket2/bool.json
aborgna-q, line 1: This has to be added to `tket2-exts/src/tket2_exts/__init__.py` too.

It looks like `wasm` is missing from there already, if you want to do a drive-by :)

tket2-hseries/src/extension/bool.rs
aborgna-q, line 1: We should explain the reason why this exists in addition to sums.

Is this only meant as a guppy artifact? The documentation seems to imply so.

PR #824, opened 2025-03-20
refactor!: Remove node parameter from Circuit

aborgna-q:
The update to `hugr 0.15` added a `Node` associated type to `HugrView`.
Over here, that got implemented as a new type parameter in `Circuit`,
```rust
struct Circuit<T = Hugr, N = Node>
```
All `impl` blocks however restrain it to the specific hugr support,
```rust
impl<T: HugrView> Circuit<T, T::Node> { ... }
```
Given how both types are connected, we should be using `T::Node` directly since an unrelated node type is not usable here.

This PR changes the definition to
```rust
pub struct Circuit<T: HugrView = Hugr>
```

As this is a breaking change, it can wait until the next breaking release.

BREAKING CHANGE: Removed node type parameter from `Circuit`

### DIFF
--- a/tket2/src/circuit.rs
+++ b/tket2/src/circuit.rs
@@ -34,16 +34,16 @@ use self::units::{filter, LinearUnit, Units};
 
 /// A quantum circuit, represented as a function in a HUGR.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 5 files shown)


aborgna-q:
Re-requesting the review since we're merging breaking changes already

PR #849, opened 2025-04-14
refactor!: More flexible pytket encoding

aborgna-q:
This PR rewrites the pytket encoding framework for hugr programs, and serves as a base for PRs to come.

Main changes from the existing implementation:

- Extensibility:
  Extensions can define their own type, operation, and constant translations and register them at runtime.
  This is done by registering `impl PytketEmitter` objects, that lets us emit arbitrary pytket commands for a hugr operation.
  See [tket2/src/serialize/pytket/extension/](https://github.com/CQCL/tket2/pull/849/files#diff-bbf828871c3cd1a6b763a955c1cd1cc3403e87ae29eead6eea0ed940551283f6) for how it's done.
- Support for arbitrary HugrViews with generic Node types (not just `Hugr`)
	- This includes sub-circuit support, although it's not exposed yet in the API.
- HUGR wires can now be associated with a list of values
  - Support custom (extensible) types: Arrays, nested tuples, ...
  - Extension declare how many qubits, bits, and params are contained in their custom types.
- Translate unsupported operations into black box barriers, grouping them as much as possible into convex subgraphs.

Circuits that could already be encoded produce the same pytket circuit (up to register reorderings) after this refactor.
Previously unsupported circuits will now be encoded using black box barriers, but we do not include decoding support here.

Not in this PR:

- Support nested dfgs and (indirect) function calls. This will probably go in a followup PR.
- Order edges between ops should generate barriers
- Extracting the blackbox hugrs encoded in a pytket circuit's and merging back the chunks.
- Tests of black-box cases. This is waiting until we can test the roundtrip.
- Documentation an examples on how to define new custom emitters.
- Define an encoder for `-hseries` ops.
- Encoder for `bool` and `int` operations. This requires some logic dealing with pytket's ClExprs that's create too much noise here. I'll make a different PR.

BREAKING CHANGE: Removed some unused variants from `TK1ConvertError` and `OpConvertError`.

### DIFF
--- a/tket2/src/serialize/pytket/extension/float.rs
+++ b/tket2/src/serialize/pytket/extension/float.rs
@@ -72,19 +72,32 @@ impl<H: HugrView> PytketEmitter<H> for FloatEmitter {
         value: &OpaqueValue,
         encoder: &mut Tk1EncoderContext<H>,
     ) -> Result<Option<TrackedValues>, Tk1ConvertError<H::Node>> {
+        use std::f64::consts::PI;
+
         let Some(const_f) = value.value().downcast_ref::<float_types::ConstF64>() else {
             return Ok(None);
         };
 
         let float = const_f.value();
-        // Special case for pi rotations
-        let val = if float == std::f64::consts::PI {
-            "pi".to_string()
-        } else {
-            float.to_string()
-        };
+        // Special cases for pi rotations
+        let approx_eq = |a: f64, b: f64| (a - b).abs() < 1e-10;
+        const VALS: [(f64, &str); 7] = [
+            (PI, "pi"),
+            (PI / 2., "pi/2"),
+            (-PI / 2., "-pi/2"),
+            (PI / 4., "pi/4"),
+            (3. * PI / 4., "3pi/4"),
+            (-PI / 4., "-pi/4"),
+            (-3. * PI / 4., "-3pi/4"),
+        ];
+        for (val, name) in VALS.iter() {
+            if approx_eq(float, *val) {
+                let param = encoder.values.new_param(name.to_string());
+                return Ok(Some(TrackedValues::new_params([param])));
+            }
+        }
 
-        let param = encoder.values.new_param(val);
+        let param = encoder.values.new_param(float.to_string());
         Ok(Some(TrackedValues::new_params([param])))
     }
 }
--- a/tket2/src/serialize/pytket/tests.rs
+++ b/tket2/src/serialize/pytket/tests.rs
@@ -66,8 +66,8 @@ const PARAMETERIZED: &str = r#"{
         "commands": [
             {"args":[["q",[0]]],"op":{"type":"H"}},
             {"args":[["q",[1]],["q",[0]]],"op":{"type":"CX"}},
-            {"args":[["q",[0]]],"op":{"params":["(1.5707963267948966) / (pi)"],"type":"Rz"}},
-            {"args": [["q", [0]]], "op": {"params": ["(3.141596) / (pi)", "alpha", "(0.7853981633974483) / (pi)"], "type": "TK1"}}
+            {"args":[["q",[0]]],"op":{"params":["((pi) / (2)) / (pi)"],"type":"Rz"}},
+            {"args": [["q", [0]]], "op": {"params": ["(3.141596) / (pi)", "alpha", "((pi) / (4)) / (pi)"], "type": "TK1"}}
         ],
         "created_qubits": [],
         "discarded_qubits": [],


aborgna-q:
The blackbox encoding path is missing some test coverage.
I plan to include roundtrip tests in the decoding PR.

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-        let output_regs: HashSet<RegisterHash> = outputs.iter().map(RegisterHash::from).collect();
-
-        for inp in &self.inputs {
-            if !output_regs.contains(&inp.into()) {
-                outputs.push(inp.clone());
+    /// Auxiliary function used to collect the input values of a node.
+    /// See [`Tk1EncoderContext::get_input_values`].
+    ///
+    /// Given a node in the HUGR, returns all the [`TrackedValue`]s associated
+    /// with its inputs. Calls
```
cqc-alec, line 205: Sentence incomplete?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
+    /// Helper to emit a new tket1 command corresponding to a single HUGR node,
+    /// with parameter outputs. Use [`Tk1EncoderContext::emit_node`] for nodes
+    /// that don't require computing parameter outputs.
+    ///
+    /// See [`Tk1EncoderContext::emit_command`] for more general cases.
+    ///
+    /// ## Arguments
+    ///
+    /// - `tk1_operation`: The tket1 operation type to emit.
+    /// - `node`: The HUGR for which to emit the command. Qubits and bits are
```
cqc-alec, line 282:
```suggestion
    /// - `node`: The HUGR node for which to emit the command. Qubits and bits are
```
?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-    /// If the circuit contains more bit inputs than the provided list,
-    /// new registers are created for the remaining bits.
+    /// See [`Tk1EncoderContext::emit_command`] for a general case emitter.
     ///
-    /// TODO: Compute output bit permutations when finishing the circuit.
-    pub fn new(circ: &Circuit<impl HugrView<Node = Node>>) -> Self {
-        let mut tracker = BitTracker::default();
+    /// ## Arguments
+    ///
+    /// - `node`: The HUGR for which to emit the command. Qubits and bits are
```
cqc-alec, line 315:
```suggestion
    /// - `node`: The HUGR node for which to emit the command. Qubits and bits are
```
?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-}
+    /// Helper to emit a new tket1 command corresponding to a single HUGR node,
+    /// with parameter outputs. Use [`Tk1EncoderContext::emit_node`] for nodes
+    /// that don't require computing parameter outputs.
+    ///
+    /// See [`Tk1EncoderContext::emit_command`] for more general cases.
+    ///
+    /// ## Arguments
+    ///
+    /// - `tk1_operation`: The tket1 operation type to emit.
```
cqc-alec, line 281:
```suggestion
    /// - `tk1_optype`: The tket1 operation type to emit.
```

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
+    /// - `node`: The HUGR for which to emit the command. Qubits and bits are
+    ///   automatically retrieved from the node's inputs/outputs.
+    /// - `circ`: The circuit containing the node.
```
cqc-alec, line 284: Reorder `node` and `circ`.

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
+    ///
+    /// The node's inputs should have exactly the same number of qubits and
+    /// bits. This method will return an error if that is not the case.
+    ///
+    /// You must also ensure that all input and output types are supported by
+    /// the encoder. Otherwise, the function will return an error.
+    ///
+    /// ## Arguments
+    ///
+    /// - `node`: The HUGR for which to emit the command. Qubits and bits are
```
cqc-alec, line 384:
```suggestion
    /// - `node`: The HUGR node for which to emit the command. Qubits and bits are
```
?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
+            return Err(Tk1ConvertError::custom(format!(
+                "Mismatched number of input and output bits while trying to emit a transparent operation for {}. We have {} inputs but {} outputs.",
+                circ.hugr().get_optype(node),
+                input_values.bits.len(),
+                total_out_count.bits,
+            )));
+        }
+        if out_params.len() != total_out_count.params {
+            return Err(Tk1ConvertError::custom(format!(
+                "Not enough parameters in the input values for a {}. Expected {} but got {}.",
```
cqc-alec, line 430: Could it be too many?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-    fn record_parameters<T: HugrView<Node = Node>>(
+    /// Helper to register values for a node's output wires.
+    ///
+    /// Returns any new value associated with the output wires.
+    ///
+    /// ## Arguments
+    ///
+    /// - `node`: The node to register the outputs for.
+    /// - `circ`: The circuit containing the node.
+    /// - `qubit_values`: An iterator of existing qubit ids to re-use for the output.
```
cqc-alec, line 645:
```suggestion
    /// - `qubits`: An iterator of existing qubit ids to re-use for the output.
```

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-                    typ,
-                    optype: optype.clone(),
-                    node: command.node(),
-                });
-            };
-            inputs.push(param.as_str());
+        // Check that we got the expected number of outputs.
+        if out_params.len() != total_out_count.params {
+            return Err(Tk1ConvertError::custom(format!(
+                "Not enough parameters in the input values for a {}. Expected {} but got {}.",
```
cqc-alec, line 674: Could it be too many?

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
+                out_params.len()
+            )));
         }
 
-        let Some(param) = fold_param_op(optype, &inputs) else {
-            return Ok(false);
-        };
+        // Update the values in the node's outputs.
+        //
+        // We preserve the order of linear values in thpe input
```
cqc-alec, line 683:
```suggestion
        // We preserve the order of linear values in the input
```

cqc-alec (changes requested):

tket2/src/serialize/pytket/tests.rs
```diff
@@ -116,21 +118,36 @@ fn compare_serial_circs(a: &SerialCircuit, b: &SerialCircuit) {
...
-    for (a, b) in a.commands.iter().zip(b.commands.iter()) {
-        assert_eq!(a.op.op_type, b.op.op_type);
-        assert_eq!(a.op.params, b.op.params);
-        assert_eq!(a.args.len(), b.args.len());
+    let b_commands: HashSet<_> = b
+        .commands
+        .iter()
+        .map(|c| serde_json::to_string(c).unwrap())
+        .collect();
+    for b in &b.commands {
```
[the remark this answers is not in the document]
aborgna-q: Good catch -.-'

We actually cannot compare commands directly, since the arguments names may change depending on the encoding toposort.
I added a partial comparison instead.

tket2/src/serialize/pytket/extension/prelude.rs
```diff
@@ -0,0 +1,92 @@
...
+            return self.tuple_op_to_pytket(node, op, &tuple_op, circ, encoder);
+        };
+        Ok(false)
+    }
+
+    fn type_to_pytket(
+        &self,
+        typ: &hugr::types::CustomType,
+    ) -> Result<Option<RegisterCount>, Tk1ConvertError<<H>::Node>> {
+        match typ.name().as_str() {
```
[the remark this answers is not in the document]
aborgna-q: I split this into a separate PR.
It is not required here (bools weren't supported before), and it's a bit noisy due to `ClExpr` handling.

tket2/src/serialize/pytket/extension/float.rs
```diff
@@ -0,0 +1,107 @@
...
+        values.params.push(param);
+        Ok(Some(values))
+    }
+}
+
+impl FloatEmitter {
+    /// Encode a rotation operation into a pytket param expression.
+    fn encode_rotation_op(op: &FloatOps, inputs: &[String]) -> Option<String> {
+        let s = match op {
+            FloatOps::fadd => format!("({} + {})", inputs[0], inputs[1]),
```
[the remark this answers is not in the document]
aborgna-q: We have a choice here between either always producing fully wrapped terms, or always doing the extra wrapping.

Old style vs alternative:

- `(2 + 4)` vs `(2) + (5)`
- `(floor(pi) ** 8.5)` vs `(floor(pi)) ** (8.5)`
- `min(10, (pi ** 2))` vs `min(10, (pi) ** (2))`

Both are equivalent and the decoder [doesn't care](https://github.com/CQCL/tket2/blob/6607ea1f6f96d92b3fa8f78d0bccfb0eb484e376/tket2/src/serialize/pytket/decoder/param.pest#L48) about which one we use.

I think the former gives sightly nicer terms in some cases, and it's what we already used, but I can change it.

tket2/src/serialize/pytket/encoder/config.rs
```diff
@@ -0,0 +1,252 @@
...
+        value: &Value,
+        encoder: &mut Tk1EncoderContext<H>,
+    ) -> Result<Option<TrackedValues>, Tk1ConvertError<H::Node>> {
+        let mut values = TrackedValues::default();
+        let mut queue = VecDeque::from([value]);
+        while let Some(value) = queue.pop_front() {
+            match value {
+                Value::Sum(sum) => {
+                    if sum.sum_type == SumType::new_unary(2) {
+                        //TODO: Add a bit and sets its value based on sum.tag
```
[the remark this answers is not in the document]
aborgna-q: Will add it in the `tket2.bool` emission PR.

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,796 @@
...
-            } else {
-                return Err(OpConvertError::UnsupportedOutputType {
-                    typ: ty.clone(),
-                    optype: optype.clone(),
-                    node: command.node(),
-                });
+        circ: &Circuit<H>,
+    ) -> Result<(), Tk1ConvertError<H::Node>> {
+        // Normally we'd use `SiblingGraph` here, but it doesn't support generic node types.
+        // See https://github.com/CQCL/hugr/issues/2010
```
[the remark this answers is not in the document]
aborgna-q: No longer relevant now that we use `region_portgraph`

tket2/src/serialize/pytket/extension/float.rs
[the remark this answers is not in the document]
cqc-alec: As a counterargument, this only works if we _know_ that the subexpressions also follow the same rule, whereas the other way works whether or not they follow it. So it makes the logic a bit less robust i.m.o.

tket2/src/serialize/pytket/extension/float.rs
```diff
@@ -0,0 +1,112 @@
...
+        value: &OpaqueValue,
+        encoder: &mut Tk1EncoderContext<H>,
+    ) -> Result<Option<TrackedValues>, Tk1ConvertError<H::Node>> {
+        let Some(const_f) = value.value().downcast_ref::<float_types::ConstF64>() else {
+            return Ok(None);
+        };
+
+        let float = const_f.value();
+        // Special case for pi rotations
+        let val = if float == std::f64::consts::PI {
```
cqc-alec, line 81: Floating point equality check alert. Perhaps we should allow a margin of error here.

We could also special-case other common multiples of pi such as pi/2 ... ?

tket2/src/serialize/pytket/extension/float.rs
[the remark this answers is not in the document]
aborgna-q: Fair enough, let's go with the that then

tket2/src/serialize/pytket/encoder.rs
```diff
@@ -1,694 +1,811 @@
...
-        }
-        RegisterUnitGenerator {
-            register,
-            next_unit: last_unit.map_or(0, |i| i + 1),
-        }
-    }
+/// Input arguments to the output parameter computation method in
+/// [`Tk1EncoderContext::emit_node_command`].
+///
+/// This can be passed
```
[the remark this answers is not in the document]
aborgna-q: Thanks for checking all of these -.-'

cqc-alec (approved):
Great, thanks!

PR #870, opened 2025-05-09
feat(tket2-py): Expose `ConstWasmModule`

croyzor:
Expose `ConstWasmModule` in python so that WASM modules can be constructed from the python tket2 interface.

This is the canonical way to make a module argument, so I think it makes sense to expose it from the api rather than require some custom python. AFAICT hugr-py doesn't allow us to define a new `CustomConst` type, so I'm not sure a pure python approach would work.


croyzor:
Closing since Seyon has showed me how to do this in pure python

PR #862, opened 2025-05-01
chore!: Bump to hugr-0.20

doug-q:
BREAKING CHANGE: bump to hugr 0.20

PR #873, opened 2025-05-15
fix!: Do not use SimpleReplacement in lazify

lmondada:
BREAKING CHANGE: Removed `LazifyMeasureRewrite` constructors. Use enum variants instead.


tket2-hseries/src/lazify_measure.rs
```diff
@@ -276,29 +182,140 @@ impl<N: HugrNode> LazifyMeasureRewrite<N> {
...
+
+        // 3. Remove the `meas_node` from the HUGR (and store its parent)
+        let meas_parent = hugr.get_parent(meas_node).expect("measure has parent");
+        hugr.remove_node(meas_node);
+
+        // 4. Create the replacement graph
+        let (repl_meas, replacement) = {
+            let n_bools = sig.output_count() - matches!(self, Self::MeasureReset(_)) as usize;
+            dbg!(meas_node, n_bools);
+            dbg!(&outputs);
```
ss2165, line 271:
```suggestion
```

check for others

ss2165 (changes requested):

aborgna-q:
Let's send it

aborgna-q:
I say let's merge #862 first and merge this to `main` afterwards

ss2165 (approved):
I'm not thrilled about two opaque hugrs being added for a test - would be nice to boil these down/use the s-expression syntax so we can see what they're doing. Maybe leave a TODO at least.

lmondada:
I can reduce it to a MWE. I've never used/seen Hugr s-exprs, can you point me to a good example?

ss2165:
> I've never used/seen Hugr s-exprs, can you point me to a good example?

No, they aren't really being used yet, this was only a suggestion for the future. I think if you are wanting to do a MWE use hugr builder as normal

PR #875, opened 2025-05-16
chore(py): update exts and eccs dependencies

ss2165:
messed up tket2-py release without doing this first, may need to be yanked 😞


aborgna-q (approved):
I guess we need a CI check for this :)

PR #888, opened 2025-05-27
feat!: Change array result ops signature to return array result

tatiana-s:
Needed for https://github.com/CQCL/guppylang/issues/981

BREAKING CHANGE: `ResultOpDef::ArrBool`,  `ResultOpDef::ArrInt`, `ResultOpDef::ArrUInt` and `ResultOpDef::ArrF64` signatures now return array results

### DIFF
--- a/tket2-hseries/src/extension/result.rs
+++ b/tket2-hseries/src/extension/result.rs
@@ -161,11 +161,27 @@ impl ResultOpDef {
     }
 
     fn result_signature(&self) -> SignatureFunc {
-        PolyFuncType::new(
-            [vec![TypeParam::String], self.type_params()].concat(),
-            Signature::new(self.arg_type(), type_row![]),
+        if self.is_array_result_op() {
+            // Arrays need to be returned to preserve no-implicit copy guarantees.
+            PolyFuncType::new(
+                [vec![TypeParam::String], self.type_params()].concat(),
+                Signature::new(self.arg_type(), vec![self.arg_type()]),
+            )
+            .into()
+        } else {
+            PolyFuncType::new(
+                [vec![TypeParam::String], self.type_params()].concat(),
+                Signature::new(self.arg_type(), type_row![]),
+            )
+            .into()
+        }
+    }
+
+    fn is_array_result_op(&self) -> bool {
+        matches!(
+            self,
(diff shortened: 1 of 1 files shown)


tket2-hseries/src/extension/result.rs
```diff
@@ -161,11 +161,27 @@ impl ResultOpDef {
...
     fn result_signature(&self) -> SignatureFunc {
-        PolyFuncType::new(
-            [vec![TypeParam::String], self.type_params()].concat(),
-            Signature::new(self.arg_type(), type_row![]),
+        if self.is_array_result_op() {
+            // Arrays need to be returned to preserve no-implicit copy guarantees.
+            PolyFuncType::new(
+                [vec![TypeParam::String], self.type_params()].concat(),
+                Signature::new(self.arg_type(), vec![self.arg_type()]),
+            )
```
lmondada, line 169: Remove the vec![] for consistency (handled by Into).

Consider also just defining `let output = if ... else ...;` and then moving the shared code for the returned value after the conditional. Keeps the branched code shorter and removes duplication.

```suggestion
                Signature::new(self.arg_type(), self.arg_type()),
```

tket2-hseries/src/extension/result.rs
```diff
@@ -161,11 +161,27 @@ impl ResultOpDef {
     }
 
     fn result_signature(&self) -> SignatureFunc {
-        PolyFuncType::new(
-            [vec![TypeParam::String], self.type_params()].concat(),
-            Signature::new(self.arg_type(), type_row![]),
+        if self.is_array_result_op() {
+            // Arrays need to be returned to preserve no-implicit copy guarantees.
```
lmondada, line 165: There is not really a "no-implicit copy guarantee" in tket2 -- that's a guppy thing, right? I'm not sure how to phrase this comment in a way that makes sense for tket2. Maybe something along the lines of "Do not consume input arrays to allow linearity"?

lmondada (approved):
I don't know the downstream implications of making breaking changes to tket2, but this seems necessary and I'm assuming other people (e.g. Seyon) know this is required. There's already a breaking change in main anyways https://github.com/CQCL/tket2/pull/854

So LGTM!

PR #900, opened 2025-06-05
feat: Add llvm lowering for debug extension

tatiana-s:
See https://github.com/quantinuum-dev/hugrverse/issues/65 for context

(Needs to be tested with selene and guppy again before merging)


ss2165 (approved):
👍

PR #904, opened 2025-06-06
test: Add some example guppy-generated hugrs

aborgna-q:
Adds guppy program examples described by @cqc-alec in https://github.com/CQCL/hugr/issues/2246.

These are _example programs_ intended to test compiler passes internally. We are not interested in testing _guppy_ here but rather programs that resemble guppy's compilation output in their structure, hence why we use a pinned guppy version and don't enforce the artifacts to be up-to-date.

We also generate `.mmd` mermaid diagrams if the example defines a `.generate_mmd` marker file.


aborgna-q:
Currently waiting for the model-loading error to be fixed, so we can generate non-empty mermaids.

test_files/guppy_optimization/justfile
```diff
@@ -0,0 +1,28 @@
...
+        uv run --no-project "$example"
+    done
+
+# Re-generate the mermaid diagrams from hugr files, for examples directories containing a `.generate_mmd` empty file.
+regenerate-mermaid:
+    #!/usr/bin/env sh
+    echo "Re-generating hugr diagrams"
+    cargo install hugr-cli
+    # Find directories containing a `.generate_mmd` file
+    for dir in $(find . -type f -name ".generate_mmd" -exec dirname {} \;); do
```
acl-cqc, line 23: snazzy!! 👍

PR #912, opened 2025-06-12
chore(pytket-encoding): Replace bool return type with EncodeStatus enum

aborgna-q:
Defines an enum to flag recoverable encoding statuses instead of returning a `bool`.
This is distinct from `Result::Error`s, as those signal an unexpected failure that should halt the conversion.

>nit: the return type seems a bit convoluted - is "subcircuit was not succesfully emitted" a possible error variant instead?

_Originally posted by @ss2165 in https://github.com/CQCL/tket2/pull/908#discussion_r2142127840_

### DIFF
--- a/tket2/src/serialize/pytket/extension.rs
+++ b/tket2/src/serialize/pytket/extension.rs
@@ -60,8 +60,8 @@ pub trait PytketEmitter<H: HugrView> {
     ///
     /// Returns an [`EncodeStatus`] indicating if the operation was successfully
     /// converted. If the operation is not supported by the encoder, it's
-    /// important to **not** the `encoder` context as that may invalidate the
-    /// context for other encoders that may be called afterwards.
+    /// important to **not** modify the `encoder` context as that may invalidate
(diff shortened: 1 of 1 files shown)


tket2/src/serialize/pytket/extension.rs
```diff
@@ -57,21 +58,19 @@ pub trait PytketEmitter<H: HugrView> {
...
     ///
-    /// Returns `true` if the operation was successfully converted. If that is
-    /// the case, no further encoders will be called.
-    ///
-    /// If the operation is not supported by the encoder, return `false`. It's
-    /// important not to modify the `encoder` in this case, as that may
-    /// invalidate the context for other encoders that may be called afterwards.
+    /// Returns an [`EncodeStatus`] indicating if the operation was successfully
+    /// converted. If the operation is not supported by the encoder, it's
+    /// important to **not** the `encoder` context as that may invalidate the
```
ss2165, line 63: not what?

ss2165 (approved):
thanks

PR #921, opened 2025-06-18
fix: Revert "fix: Change array result ops signature to return array resul…

ss2165:
…t (#888)"

This reverts commit 4bde6d9d5906f355dd809b6a7fc804680f99ddeb.


tket2-hseries/src/llvm/result.rs
```diff
@@ -154,6 +154,7 @@ impl<'c, H: HugrView<Node = Node>> ResultEmitter<'c, '_, '_, H> {
         tag_ptr: BasicValueEnum,
         tag_len: IntValue,
     ) -> Result<()> {
+        // TODO update to return array after https://github.com/CQCL/tket2/pull/888
```
aborgna-q, line 157: Should we update the TODO? Are we still planning to add that?

aborgna-q (approved):

PR #928, opened 2025-06-24
feat: Add badgerv2_unstable feature, RewriteSpace and CommitFactory

lmondada:
This requires https://github.com/CQCL/hugr/pull/2349 and https://github.com/CQCL/hugr/pull/2402 to be merged in, as well as a breaking hugr release before it can be merged.


tket2/src/rewrite_space.rs
```diff
@@ -0,0 +1,190 @@
...
+                    .parents(commit_id)
+                    .map(|cm| (fmt_commit_id(commit_id), fmt_commit_id(cm))),
+            );
+
+            // Add mutual exclusion between siblings if their invalidation sets overlap
+            let children = self.state_space.children(commit_id);
+            for children_pair in children.combinations(2) {
+                let &[c1, c2] = children_pair.as_slice() else {
+                    panic!("Expected 2 children, got {}", children_pair.len());
+                };
```
aborgna-q, line 134:
```suggestion
            for &[c1, c2] in children.array_combinations() {
```

aborgna-q, line 134: How much do we expect the set of children to grow? Is it worth trying to reduce the quadratic cost?
(there's some tradeoffs we could take vs. the size of the invalidation sets)

As a starter, we should probably take the `invalidation_set(c1, commit_id).collect()`s out of the double-loop, and using `BTreeSet::is_disjoint` inside.

aborgna-q (commented):
Adding some comments, I still need to read through `rewrite_space/`

kmk142789 (approved):

aborgna-q:
Closing as out of date 
We'll see if we can refactor this later on.

🥲

PR #939, opened 2025-06-26
feat: Add tket1-passes subcrate

### LINKED ISSUE #940: Cache tket1-passes library in-between workflow runs
We'd need a mechanism to detect when to rebuild the library... (presumably checking the latest version of tket on the server)

lmondada:
This introduces a way to call into TKET1's C++ library from here (tket2). In writing this, it was my priority to 

1. not complicate the build process for anyone that is not interested in this functionality, and
2. not slow down CI too much.

By not including `tket1-passes` in Cargo's `default-members`, this will not be compiled unless required explicitly. One way to simplify the compilation process further would be to provide pre-compiled libraries for the main platforms. This requires a certain maintenance burden on us, however, which is why I haven't gone down this route (yet).

On CI speed: this uses a pre-compiled version of tket1, so that the compilation times for the library are low (~30s). The library is uploaded as artifact and re-used for every job. I've also amended the CI job `rust-stable-no-features` to not compile `tket1-passes` (as well as the other CLI tools that have no tests) for further speed up. A way to speed up the CI further would be to cache the compiled library across workflow runs. I'll create an issue for this.

Finally, a note on platform support: the current build scripts will probably not work on Windows. I am not familliar with Windows tooling and had no way of testing it. Is this something that must be tackled before merging @aborgna-q ?

Closes #940


lmondada:
#940

tket1-passes/rust/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "tket1-passes"
```
aborgna-q, line 2: Naming thread: Can we replace the `1` here? Maybe `legacy`?

tket1-passes/rust/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+description = "Rust interface for tket one-pass optimization library"
```
aborgna-q, line 8: Add a README for the crate too

tket1-passes/rust/src/lib.rs
```diff
@@ -0,0 +1,238 @@
...
+mod ffi {
+    #![allow(non_camel_case_types)]
+    #![allow(non_upper_case_globals)]
+    #![allow(dead_code)]
+    #![allow(non_snake_case)]
+    include!(concat!(env!("OUT_DIR"), "/bindings.rs"));
+}
+
+/// Error types that can occur when using the tket library
+#[derive(Error, Debug)]
```
aborgna-q, line 21:
```suggestion
#[derive(Error, Debug, Clone)]
#[non_exhaustive]
```

tket1-passes/rust/Cargo.toml
```diff
@@ -0,0 +1,29 @@
+[package]
+name = "tket1-passes"
+version = "0.1.0"
```
aborgna-q, line 3: Pre-release version (so the first one gets the 0.1.0 value)
```suggestion
version = "0.0.0"
```
Are we planning a separate versioning from the main `tket2` crates?

lmondada:
ah my bad, let's see if this fixes it!

.github/workflows/ci.yml
```diff
@@ -63,9 +63,31 @@ jobs:
       with:
         filters: .github/change-filters.yml
 
+
+  build-tket1-passes:
+    name: Build tket1-passes C++ library
+    runs-on: ubuntu-latest
```
aborgna-q, line 69: This should be at least filtered by the rust changes flag
```suggestion
  build-tket1-passes:
    name: Build tket1-passes C++ library
    if: ${{ needs.changes.outputs.rust == 'true' }}
    runs-on: ubuntu-latest
```
Be sure to add the new dirs to `.github/change-filters.yml`.

My main issue is that it's adding 30% runtime to the critical path of _every_ PR required checks run.

Do you think caching will work well here?
Would it make sense to **only** build/test this on changes to `tket1-passes` (at least for the required checks)?

.github/workflows/semver-checks.yml
```diff
@@ -39,11 +39,38 @@ jobs:
...
-    with:
-      apt-dependencies: llvm-14
-    secrets:
-      GITHUB_PAT: ${{ secrets.HUGRBOT_PAT }}
+    steps:
+      - name: Download tket1-passes C++ library
+        uses: actions/download-artifact@v4
+        with:
+          name: tket1-passes-lib
+          path: tket1-passes/lib
```
aborgna-q, line 71: I don't think this will work :/
`semver-checks` runs in its own container, hence why we had to add the `apt-dependencies` parameter.

I'm not sure what's a nice solution here

lmondada: failing indeed :/

.github/workflows/ci.yml
lmondada: if we specify the version of tket to be building against, then we can just watch local files, cache the compiled library and only rebuild when one of our files locally is changed. I just don't know how/where to cache the library, but I'll look it up now.

lmondada: ok, seems like caching is quite easy. Will implement that.

tket1-passes/rust/Cargo.toml
lmondada: Yes, I don't think there will be very frequent updates to this crate (and it might be more in sync with tket1 than the rest of tket2...). I also don't think we should adopt the tket1 versioning, to give us the flexibility of releasing when useful to us.

.github/workflows/semver-checks.yml
lmondada: I don't know what to do about this 😢  I'll come back to it another time.

cqc-alec (commented):
One more suggestion: please add the build artifacts to `.gitignore`:
```
tket1-passes/cpp/.cache
tket1-passes/cpp/CMakeUserPresets.json
tket1-passes/cpp/build
tket1-passes/lib
```

tket1-passes/cpp/build.sh
```diff
@@ -0,0 +1,37 @@
...
+# Build the shared library
+make -j$(nproc 2>/dev/null || echo 4)
+
+# Copy the library to lib/ for Rust usage
+make install
+
+echo "Build complete!"
+
+cd ../..
+echo "Dynamic library available at: $(pwd)/lib/libtket1-passes.dylib"
```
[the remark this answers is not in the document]
lmondada: Yep apologies, I adjusted the build scripts to be platform agnostic but didnt fix the comments/messages. Fixed now

.github/change-filters.yml
```diff
@@ -11,6 +11,7 @@ rust:
   - "tket2-hseries/**"
   - "badger-optimiser/**"
   - "compile-rewriter/**"
+  - "tket1-passes/rust/**"
```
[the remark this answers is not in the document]
lmondada: Yes, you're right actually that we also need to trigger a re-build if any of the C++ changes. I will amend this to `tket1-passes/cpp`.

tket1-passes/cpp/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+add_library(tket1-passes SHARED
+    src/tket1-passes.cpp
+)
+
+# Link against tket and all its dependencies (since tket links them as PRIVATE)
+target_link_libraries(tket1-passes PRIVATE tket::tket)
+
+# Set the library properties for shared library creation
+set_target_properties(tket1-passes PROPERTIES
+    VERSION 0.0.1
```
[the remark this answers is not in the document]
lmondada: I've removed it.

tket1-passes/cpp/src/tket1-passes.cpp
```diff
@@ -0,0 +1,104 @@
...
+    return TKET_ERROR_UNKNOWN;
+  }
+}
+
+TketError tket_clifford_resynthesis(TketCircuit *tc, bool allow_swaps) {
+  if (!tc)
+    return TKET_ERROR_NULL_POINTER;
+  try {
+    std::lock_guard<std::mutex> lock(global_mutex);
+    Transforms::clifford_resynthesis(std::nullopt, allow_swaps)
```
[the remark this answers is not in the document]
lmondada: yes, i do. I will fix this

tket1-passes/cpp/src/tket1-passes.h
```diff
@@ -0,0 +1,72 @@
...
+ * decomposition. Can decompose to TK2 or CX gates.
+ *
+ * @param circuit Circuit to transform (modified in-place)
+ * @param target_gate Target two-qubit gate type (CX or TK2)
+ * @param cx_fidelity Estimated CX gate fidelity (used when target_gate=CX)
+ * @param allow_swaps Whether to allow implicit wire swaps
+ * @return TKET_SUCCESS if successful, error code otherwise
+ */
+TketError tket_two_qubit_squash(
+    TketCircuit* circuit, TketTargetGate target_gate, double cx_fidelity,
```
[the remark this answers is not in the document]
lmondada: This is within a "extern C" code block and is only called from Rust, over the FFI boundary, so I think we need to stick to C functions.

tket1-passes/cpp/src/tket1-passes.h
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * Resynthesise all Clifford subcircuits and simplify using Clifford rules.
+ * This can significantly reduce the two-qubit gate count for Clifford-heavy
+ * circuits.
+ *
+ * @param circuit Circuit to transform (modified in-place)
+ * @param allow_swaps Whether the rewriting may introduce wire swaps
+ * @return TKET_SUCCESS if successful, error code otherwise
+ */
+TketError tket_clifford_resynthesis(TketCircuit* circuit, bool allow_swaps);
```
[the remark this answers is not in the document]
lmondada: see above :)

tket1-passes/cpp/src/tket1-passes.cpp
```diff
@@ -0,0 +1,104 @@
...
+        .apply(tc->circuit);
+    return TKET_SUCCESS;
+  } catch (...) {
+    return TKET_ERROR_UNKNOWN;
+  }
+}
+
+const char *tket_error_string(TketError error) {
```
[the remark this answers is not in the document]
lmondada: Oups you are right. Fixed it.

tket1-passes/rust/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+    /// Memory allocation failed
+    #[error("Memory allocation failed")]
+    Memory,
+    /// Failed to process JSON
+    #[error("Failed to process JSON: {0}")]
+    JsonError(String),
+    /// Unknown error occurred
+    #[error("Unknown error occurred")]
+    Unknown,
+    /// FFI error
```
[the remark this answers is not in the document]
lmondada: Fixed :)

tket1-passes/rust/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+impl Drop for Tket1Circuit {
+    fn drop(&mut self) {
+        if !self.inner.is_null() {
+            unsafe { ffi::tket_circuit_destroy(self.inner) };
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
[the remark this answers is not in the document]
lmondada: It's not easy to provoke errors within the C++. I've added two tests that cover the errors that I can create easily from Rust (invalid target gates, unexpected null pointers and badly formatted JSON).

lmondada:
Thanks @cqc-alec

cqc-alec (approved):
Thanks!

PR #954, opened 2025-07-04
feat: Add KAK Commit Factory

daniel-mills-cqc:


tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+pub struct KAKFactory;
+
+struct SearchingWalker <'w> {
+    walker: Walker <'w>,                    // Initial state of walker.
+    frontier_wires: Vec<PersistentWire>,    // Wires to be expanded.
+    pattern: KAKResynthesis,                // Current pattern.
+    two_qubit_node: Option<PatchNode>,      // If a 2q gate has been found along one wire,
+                                            // but not yet the other, then it is stored here.
```
lmondada, line 34: Put the documentation for the fields on a separate line above the fields, starting with `///`. That way rust will automatically pick it up as docstrings (and include it in automatically generated documentation) :)

```suggestion
    /// Initial state of walker.
    walker: Walker <'w>,
    /// Wires to be expanded.
    frontier_wires: Vec<PersistentWire>,
    /// Current pattern.
    pattern: KAKResynthesis,
    /// If a 2q gate has been found along one wire,
    /// but not yet the other, then it is stored here.
    two_qubit_node: Option<PatchNode>,
```

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+    println!("{}", hugr.mermaid_string());
+
+    let factory = KAKFactory;
+
+    let pers_hugr = PersistentHugr::with_base(hugr);
+    for node in pers_hugr.nodes() {
+        let walker = Walker::from_pinned_node(node, pers_hugr.as_state_space());
+        // walker.get_wire(node, OutgoingPort::from(0));
+        factory.find_pattern_matches(node, walker);
+    }
```
lmondada, line 249: Add an assertion to check this does what you expect.

The simplest would be to add a variable `let mut n_matches = 0` before the loop, and then count the number of matches you find:
```rust
n_matches += factory.find_pattern_matches(node, walker).count();
```

After the loop, you would write `assert_eq(n_matches, 1)`.

A stronger test would involve looking into the matches obtained and checking it contains the gates you expect.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+#[test]
+fn test_my_test() {
+    println!("hello world")
+}
```
lmondada, line 255: remove :)

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+fn one_cz_2qb_hugr() -> Hugr {
+    let mut builder = DFGBuilder::new(endo_sig(vec![qb_t(), qb_t()])).unwrap();
+    let [q0, q1] = builder.input_wires_arr();
+    let cz1 = builder.add_dataflow_op(Tk2Op::CZ, vec![q0, q1]).unwrap();
+    let [q0, q1] = cz1.outputs_arr();
+    builder.finish_hugr_with_outputs(vec![q0, q1]).unwrap()
+}
+
+#[test]
+fn test_kak_resynthesis_init() {
```
lmondada, line 217: not important but useful to know:

if you replace `#[test]` with `#[rstest]` (you need to import it with `use rstest::rstest`) and you add `#[fixture]` (also imported from `rstest`) in front of your helper function `one_cz_2qb_hugr`, then you can write the test as

```rust
#[rstest]
fn test_kak_resynthesis_init(one_cz_2qb_hugr: Hugr) { ... }
```

It's just syntactic sugar to move the construction of the `Hugr` from a function call in your function body as you have it now, to something that looks like an argument passed to the test.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+
+pub struct KAKResynthesis{
+    wires: Vec<PersistentWire>
+}
+
+impl IterMatchedWires for KAKResynthesis {
+    fn matched_wires(&self) -> impl Iterator<Item = &PersistentWire> + '_ {
+        self.wires[0].iter().chain(self.wires[1].iter())
+    }
+}
```
lmondada, line 25: There's something weird here: the way this is written worked for `wires: [Vec<PersistentWire>; 2]`, but it looks like you've changed it to a simple `Vec<_>`. In this case, `self.wires.iter()` should be all you need here.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+                matches.push((searching_walker.pattern, searching_walker.walker));
+                continue;
+            };
+            
+            // Assuming now that there are wires to be searched, get
+            // the input port of the wire. There should be exactly one.
+            let (frontier_node, _) = searching_walker.walker.wire_pinned_inports(&wire_to_search).next().expect("error"); 
+            
+            // Expand the wire into the future
+            for expanded_walker in searching_walker.walker.expand(&wire_to_search, Direction::Outgoing) {
```
lmondada, line 88: I think you are traversing wires backwards here: you are getting an _incoming_ port, and traversing it to get the next _outgoing_ port. It should be the opposite:

```suggestion
            let (frontier_node, _) = searching_walker.walker.wire_pinned_outports(&wire_to_search).next().expect("error"); 
            
            // Expand the wire into the future
            for expanded_walker in searching_walker.walker.expand(&wire_to_search, Direction::Incoming) {
```

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+                
+                // Get the new expanded version of the wire to search along.
+                let expanded_wire = expanded_walker.get_wire(
+                    frontier_node, 
+                    IncomingPort::from(0),
```
lmondada, line 93: I don't think this will always be port 0? Also, if you change the above direction, you will need to change it here too.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+                );
+
+                // Get the node at the end of this newly expanded wire.
+                let (expanded_node, _) = expanded_walker.wire_pinned_outport(&expanded_wire).expect("error");
+
+                // Get the type of operation this node implements
+                let ext_op = expanded_walker.as_hugr_view().get_optype(expanded_node).as_extension_op().expect("Not extension op");
+                let tket2_op = Tk2Op::from_extension_op(ext_op).expect("Not TKET2 OP");
```
lmondada, line 101: I wouldn't error if it is not a TKET2 (it could be an input/output node). Instead you could write

```suggestion
                let Ok(tket2_op) = Tk2Op::from_extension_op(ext_op) else {
                    continue;
                };
```

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+
+                // If it it not a quantum operation then there is no need to continue
+                // search as this is not a 2q sub circuit.
+                if !tket2_op.is_quantum() {
+                    continue;
+
```
lmondada, line 107: Actually, both here and just above in the suggestion I made about ignoring non-tket2 ops: you need to make sure that in this case you still add the existing match to the list of completed matches to be returned.

Probably the simplest way to do this would be to add `searching_walker` back to the queue.

tket2/src/rewrite_space/factory/kak.rs
lmondada: See also comment below about adding `searching_walker` back to the queue in this case.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+                        // We add the pattern to our matches, and add the
+                        // search back to the queue to continue exploring
+                        // along the outputs of the new 2q gate.
+                        // If it is not the same then we don't need to do
+                        // anything as this is not a 2q sub circuit.
+                        if two_qubit_node == expanded_node {
+
+                            searching_walker.pattern.wires.push(expanded_wire);
+
+                            matches.push((searching_walker.pattern, searching_walker.walker));
```
lmondada, line 137: I suspect the compiler must be complaining here about using `searching_walker.pattern` and `searching_walker.walker` in more than one place (more specifically, it will say "use of moved value" further below). You must clone the data:

```suggestion
                            matches.push((searching_walker.pattern.clone(), searching_walker.walker.clone()));
```

That being said, this will add all possible subcircuits (not just the maximal ones). Is this what you want? Might indeed be the right way to go provided we don't get far too many matches.

tket2/src/rewrite_space/factory/kak.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+fn one_cz_2qb_hugr() -> Hugr {
+    let mut builder = DFGBuilder::new(endo_sig(vec![qb_t(), qb_t()])).unwrap();
+    let [q0, q1] = builder.input_wires_arr();
+    let cz1 = builder.add_dataflow_op(Tk2Op::CZ, vec![q0, q1]).unwrap();
+    let [q0, q1] = cz1.outputs_arr();
+    builder.finish_hugr_with_outputs(vec![q0, q1]).unwrap()
+}
+
```
lmondada, line 215: Add a Hugr with more CZ and more than one match to test on. E.g.:

```rust
fn three_cz_hugr() -> Hugr {
    let mut builder = DFGBuilder::new(endo_sig(vec![qb_t(); 3])).unwrap();
    let [q0, q1, q2] = builder.input_wires_arr();
    let cz1 = builder.add_dataflow_op(Tk2Op::CZ, vec![q0, q1]).unwrap();
    let [q0, q1] = cz1.outputs_arr();
    let cz2 = builder.add_dataflow_op(Tk2Op::CZ, vec![q0, q1]).unwrap();
    let [q0, q1] = cz2.outputs_arr();
    let cz3 = builder.add_dataflow_op(Tk2Op::CZ, vec![q0, q2]).unwrap();
    let [q0, q2] = cz3.outputs_arr();
    builder.finish_hugr_with_outputs(vec![q0, q1, q2]).unwrap()
}

lmondada (commented):
Hey Dan!

The logic is great and I like the names (`SearchingWalker`, `frontier` etc). Most of my comments are about Rust or some inconsistencies.

General Rust code standards:

- make sure you build this and run your tests: `cargo test` (will run all tket2 tests) or `cargo test kak` (will run the tests in this file)
- highly recommend you use `cargo fmt` to format your file in the standard way. Some formatting conventions in rust get a second to get used to (esp. around iterators imo), but the consistency is v nice to have
- less important but good to know: `cargo clippy` is the Rust linter. If you run `cargo clippy --fix` it will adjust your code to follow all best practices. Note that `--fix` will not work if you have any files not commited in your repo. Either clean up your directory or use `cargo clippy --fix --allow-dirty`

Let me know if you have any issues with making your code compile :) You should be pretty close to a working solution and I look forward to testing it!

tket2/src/rewrite_space/factory/kak.rs
daniel-mills-cqc: Thanks!
Adding all matches was indeed what I was going for as a first try.

PR #960, opened 2025-07-08
docs: Add tket-c-api setup recipe and instructions

aborgna-q:


justfile
```diff
@@ -4,10 +4,16 @@ help:
...
 # setting up the pre-commit hooks.
-setup:
+setup: setup-tket-c-api
     uv sync
     [[ -n "${TKET2_JUST_INHIBIT_GIT_HOOKS:-}" ]] || uv run pre-commit install -t pre-commit
 
+# Configure the conan cache for tket-c-api.
+setup-tket-c-api:
+    #!/usr/bin/env bash
+    uv run conan remote add tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs --index 0
```
aborgna-q, line 14: This line fails if the remote is already configured. The bash subshell runs with `+e`, so the error gets ignored and the next line is still executed.

lmondada, line 14:
```suggestion
    uv run conan remote add -f tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs --index 0
```

By adding `-f` this should not error if already present.

lmondada (approved):
Unfortunately, there does not seem to be a way to specify project-specific remotes or passing them on the command line.

So `uv run conan remote add -f tket-libs url` is the right approach.

Thanks!

PR #962, opened 2025-07-09
feat: Add guppy extension with drop operation

### LINKED ISSUE #946: Define drop operation
Add a new extension op `drop<T: Any>: T -> ()`.

The plan is to use this op in Guppy to handle affine types like arrays: Whenever an affine type is dropped in Guppy, it will emit this op. After monomorphisation, we need to replace this op with code to e.g. drop nested array etc.

This essentially an ad hoc type class

tatiana-s:
Closes #946


mark-koch (approved):

aborgna-q (approved):

tket2-exts/src/tket2_exts/data/tket2/guppy.json
```diff
@@ -0,0 +1,31 @@
+{
+  "version": "0.1.0",
+  "name": "tket2.guppy",
+  "types": {},
+  "operations": {
+    "drop": {
+      "extension": "tket2.guppy",
+      "name": "drop",
+      "description": "Drop the input wire.",
```
acl-cqc, line 9:
```suggestion
      "description": "Drop the input wire. Applicable to guppy affine types only.",
```

PR #975, opened 2025-07-17
feat: Add a `borrow_array` type replacement pass

tatiana-s:
Draft because performance benchmarks with Guppy using this are needed first to see if this is suitable. There are probably a lot of ways that some of the more repetitive code in all the `dest` functions could be generalised but for now I prioritised getting something working.


tket-qsystem/src/replace_bools.rs
```diff
@@ -213,6 +219,21 @@ fn measure_reset_dest() -> NodeTemplate {
...
 }
 
+fn array_clone_dest(size: u64, elem_ty: Type) -> NodeTemplate {
+    let array_ty = array_type(size, elem_ty.clone());
+    let mut dfb = DFGBuilder::new(inout_sig(
+        vec![array_ty.clone()],
+        vec![array_ty.clone(), array_ty],
+    ))
+    .unwrap();
+    // TODO: Find less hacky solution.
```
acl-cqc, line 229: Have we lowered drops already? (If not, can we lower to a drop?)

tket-qsystem/src/replace_bools.rs
```diff
@@ -277,6 +298,49 @@ fn lowerer() -> ReplaceTypes {
...
+    lw.replace_parametrized_op_with(
+        array::EXTENSION.get_op(ARRAY_CLONE_OP_ID.as_str()).unwrap(),
+        move |args| {
+            let [size, elem_ty] = args else {
+                unreachable!()
+            };
+            let size = size.as_nat().unwrap();
+            let elem_ty = elem_ty.as_runtime().unwrap();
+            if !elem_ty.copyable() {
+                Some(array_clone_dest(size, elem_ty))
```
acl-cqc, line 312: Consider `(!elem_ty.copyable()).then(|| array_clone_dest(size,elem_ty))` instead of the if...else

tket-qsystem/src/replace_bools.rs
```diff
@@ -277,6 +298,49 @@ fn lowerer() -> ReplaceTypes {
...
+        array::EXTENSION
+            .get_op(ARRAY_DISCARD_OP_ID.as_str())
+            .unwrap(),
+        move |args| {
+            let [size, elem_ty] = args else {
+                unreachable!()
+            };
+            let size = size.as_nat().unwrap();
+            let elem_ty = elem_ty.as_runtime().unwrap();
+            let drop_op_def = GUPPY_EXTENSION.get_op(DROP_OP_NAME.as_str()).unwrap();
```
acl-cqc, line 330: as above, consider `(!elem_ty.copyable()).then(|| ...)` below, or else an early `if elem_ty.copyable() {return None}` here

tket-qsystem/src/lower_drops.rs
```diff
@@ -18,6 +22,28 @@ impl<H: HugrMut<Node = Node>> ComposablePass<H> for LowerDropsPass {
 
     fn run(&self, hugr: &mut H) -> Result<Self::Result, Self::Error> {
         let mut rt = ReplaceTypes::default();
+
+        // future(bool) is not in the default linearizer handler so we add it here.
```
acl-cqc, line 26: I think I favour commoning this up with the equivalent code in `replace_bools.rs` by making a helper-method that makes a ReplaceTypes with future-bool linearizer already set-up.

tket2-hseries/src/replace_borrow_arrays.rs
```diff
@@ -0,0 +1,1095 @@
...
+#[derive(Default, Debug, Clone)]
+pub struct ReplaceBorrowArrayPass;
+
+impl<H: HugrMut<Node = Node>> ComposablePass<H> for ReplaceBorrowArrayPass {
+    type Error = ReplaceBorrowArrayPassError<H::Node>;
+    type Result = ();
+
+    fn run(&self, hugr: &mut H) -> Result<(), Self::Error> {
+        // TODO uncomment once https://github.com/CQCL/hugr/issues/1234 is complete
+        // ensure_no_nonlocal_edges(hugr)?;
```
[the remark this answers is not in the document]
acl-cqc: Not quite sure what's going on here but we "temporarily" added a guppy fix to avoid generating nonlocals, however now Hugr has LocalizeEdges pass, so can we remove the guppy workaround? Not for this PR, but do we have a guppy issue?

Here, I'm not sure it's worth adding an error variant for the existence of nonlocal edges, to me I think this is an assert (a bug in our code) if we have not run LocalizeEdges already (or if there is a bug in LocalizeEdges !).

acl-cqc (commented):
TBH I feel that if this is passing tests then rather than benchmarking performance first, we should just get this in; and then work on adding benchmarks and improving/optimizing later. (e.g. from serialized Hugrs, so we are not waiting for guppy-`main` to generate them; and measuring compilation time and size of compiled output, down to whatever level is easy to test - maybe the Hugr *input* to hugr-llvm, or the LLVM-IR output from hugr-llvm, without running LLVM itself.) @mark-koch what do you think?

PR #977, opened 2025-07-19
refactor: Create backtracking optimiser, refactor badger to use it

lmondada:
The code for badger currently mixes the "rewriting logic" (what rewrites can be applied and should be considered) with the "optimisation logic" (how do I explore the possible rewrites to find a cost-minimising circuit.

This PR refactors the optimisation logic out, as a result:
- the optimiser can be reused in the future in other contexts
- the code for badger is a lot more readable and maintainable (this is the part that I intend to modify further nex)

Note that this does not touch any of the multi-threaded versions of badger. Those are harder to refactor in the same way.


kmk142789 (approved):

mariagg-quantinuum (approved):
I checked out the branch locally, built it, and ran the tests with --features portmatching. Everything passed. The refactor makes sense, and the existing tests cover the functionality. Looks good to me 👍

ss2165 (approved):
👍

PR #979, opened 2025-07-23
feat: Lower "Drop" op in guppy extension, update hugr minor version

acl-cqc:
Add LowerDropsPass, use in QSystemPass.

hugr 0.22.0 -> 0.22.1 gets the "copy_discard_array" being in the default ReplaceTypes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1033,9 +1033,9 @@ dependencies = [
 
 [[package]]
 name = "hugr"
-version = "0.22.0"
+version = "0.22.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cc1bdd463fd409d55a1fb3a2e3aeded7567fe7852129d8b40bd09c5070661694"
+checksum = "ccc992a69300fb5835dd9fd259620b21b726bc22ade4aea9fcd20fbdd580e653"
 dependencies = [
  "hugr-core",
  "hugr-llvm",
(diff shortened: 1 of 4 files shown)


tket2-hseries/src/lower_drops.rs
```diff
@@ -0,0 +1,74 @@
...
+impl<H: HugrMut<Node = Node>> ComposablePass<H> for LowerDropsPass {
+    type Error = ReplaceTypesError;
+
+    /// Returns whether any drops were lowered
+    type Result = bool;
+
+    fn run(&self, hugr: &mut H) -> Result<Self::Result, Self::Error> {
+        let mut rt = ReplaceTypes::default();
+        rt.linearizer()
+            .register_callback(array_type_def(), copy_discard_array);
```
acl-cqc, line 22: This should possibly be done in `DelegatingLinearizer::default()`, which currently adds only `linearize_array` (for ValueArray). Or, perhaps that change happens when we remove ValueArray?

I also worry that there are other types we need to add (i.e. anything guppy-affine but hugr-linear)....however I think if there were any, we would need them in the `linearize_arrays` pass, and the only linearizer callback I see there is...this one!

acl-cqc: Each `ReplaceTypes` has its own (independent) list of callbacks, and the `linearize_array` callback is stateless, so it should work. But it is more of a maintenance issue - i.e. we are now getting up to 3 passes.

If multiple `ReplaceTypes` passes actually get run, we'll probably need to make sure `linearize_array` is added to all of them (not just one) - each pass only runs linearization on *the parts of the Hugr that it changes*.

So, might be better to add this to `ReplaceTypes::default()` now (as a Hugr PR, for the hugr-v0.22 release), and then one of us can remove the existing tket `register_callback` rather than add another....it's a very simple Hugr PR!

PR #980, opened 2025-07-24
feat!: Update `hugr` dependency to `0.22.0`

aborgna-q:
BREAKING CHANGE: Updated public dependency `hugr` to `0.22.0`


lmondada (approved):

lmondada:
I can't solve this right now sorry 😐 I will look early tomorrow?

It's a weird one... The only thing I can think of is trying with portmatching `0.3.1`, see if it's got something to do with that update.

(A rust serialization test will fail, but you could ignore it, that's not important)

PR #982, opened 2025-07-25
test: Add test of importing extensions in python

croyzor:
Historical trivia: the WASM extension was broken when we updated hugr to 0.21, but there was no test to show it! Now there is 🙂

### DIFF
--- a/tket-exts/tests/test_validate_exts.py
+++ b/tket-exts/tests/test_validate_exts.py
@@ -33,5 +33,5 @@ exts = [
 @pytest.mark.parametrize("ext", exts)
 def test_validate_extension(ext):
     e = ext()
-    assert(len(e.types) + len(e.operations) > 0)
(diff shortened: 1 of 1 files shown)


tket-exts/tests/test_validate_exts.py
```diff
@@ -0,0 +1,36 @@
...
+    quantum,
+    result,
+    rotation,
+    wasm,
+]
+
+
+@pytest.mark.parametrize("ext", exts)
+def test_validate_extension(ext):
+    ext()
```
cqc-alec, line 35: It would be nice to add an actual assert to the test, other than just not-raising -- is there a simple thing we can check about the result of `ext()`?

croyzor: Hmm, could do a json serialisation round trip? Or something like `len(ext().types) + len(ext().operations) > 1`

PR #987, opened 2025-07-25
feat!: Rename tket2* libs to tket*

aborgna-q:
This affects both rust and python libraries.
Since we are renaming the main packages, I just renamed all the sub-crates / libs too to keep thing consistent and reduce confusion.

Extension names have **not** been updated, as I worry about further breakage.
Should we update them (in a separate PR)?

### Project breakage

- For dependents, the breakage should be limited to having to update the dependency names and some types like `Tk2Op` to `TketOp`, `TKET2_EXTENSION` to `TKET_EXTENSION`, etc.

- For the local automation, I tried to update the workflows as much as possible, but we can only test them once merged.
  - `release-please`'s manifest has been updated. I'll recreate the last python release tags without the `2`, and the workflow should pick up from there. We'll need to swap some pypi API keys before merging the release PR.
  - `release-plz` will probably get confused with which commits should be picked.
    - ~It also does not have permission to publish new crates, so we'll need to manually push `tket-hseries` and update the API keys.~
    - Both the `tket` and `tket-qsystem` crates have been created and setup for trusted publishing, so the release should work correctly.
  
### Follow up and deprecation.

We'll need to make patch releases for the `tket2*` libraries adding a deprecation warning to the README and directing users to the new libraries.

- Por python, take a look at the `pypi-rename` [README template](https://github.com/simonw/pypi-rename/blob/main/%7B%7Bcookiecutter.old_package_name%7D%7D/README.md).
- For rust, make sure to update both the README and `lib.rs` rusdocs.
  - Rustdoc recently added support for warning boxes: [rust-lang/rust#106561](https://www.github.com/rust-lang/rust/pull/106561)

BREAKING CHANGES: Yes
BREAKING CHANGE: Libraries renamed from `tket2*` to `tket*`


DEVELOPMENT.md
```diff
@@ -226,4 +226,19 @@ ready, create a draft PR so that the release team can review it.
...
 The wheel building process and publication to PyPI is handled by the CI.
 Just create a [github release](https://github.com/CQCL/tket2/releases/new) from the **unmerged** branch.
-The release tag should follow the format used in the previous releases, e.g. `tket2-py-v0.1.1`.
+The release tag should follow the format used in the previous releases, e.g. `tket-py-v0.1.1`.
+
+## 🎫 TKET, pytket, and TKET2
+
+For historical reasons, the repository hosting this project is called `tket2`.
+
+`pytket` / `tket1` was the first version of the tket library, written in C++ and Python.
```
cqc-alec, line 235:
```suggestion
`pytket` / `tket` (a.k.a. `tket1`) was the first version of the tket library, written in C++ and Python.
```

cqc-alec:
Couple of mentions of "Tket2" in `tket-py/examples/2-Rewriting-Circuits.ipynb`.

PR #999, opened 2025-07-29
ci: switch to trusted publishing for tket pypi

ss2165:
See: https://www.maturin.rs/distribution.html#using-pypis-trusted-publishing


lmondada (approved):
👍

PR #1002, opened 2025-07-29
chore: Fetch tket-c-api from conan remote

lmondada:
This greatly simplifies the build! No setup is required anymore. Just `uv run cargo build` and you are good to go -- or `cargo build` with `conan` in your path. In addition, no C++ compilation should happen if you are on one of the supported platforms (Linux X86, Mac ARM, Mac X86 or Windows X86).

As before, If you do not want to use `conan`, you just need to set `TKET_C_API_PATH` to point to the install folder of a compiled dynamic library of `tket-c-api`.


tket1-passes/README.md
```diff
@@ -7,26 +7,19 @@ API.
 
 ## Building instructions
 
-Building `tket1-passes` is done using `conan >= 2.0`. After installing `conan`,
-(`pip install conan` or use `pipx`), you must set up `conan` to use the
-artifactory containing the tket package:
-
-```
-conan remote add tket-libs https://quantinuumsw.jfrog.io/artifactory/api/conan/tket1-libs --index 0
```
cqc-melf, line 15: Are you not using the jgrof remote anymore?

tket1-passes/README.md
```diff
@@ -7,26 +7,19 @@ API.
...
-
-For the time being, one extra step is required to export `tket-c-api` to the conan cache:
-
-```
-cd tket-c-api && conan create . --build=missing"
-```
-
-That's it! You can now build the library using `cargo build`.
+Building `tket1-passes` is done using `conan >= 2.0`. This will be installed
+automatically by `uv` if using `just` or `uv`. If you are running `cargo build`
```
cqc-melf, line 11: It is not fully clear to me what the just is referring to, could you maybe give some more details?

tket1-passes/README.md
lmondada: Yes I am, but I'm (sneakily) adding it in the `tket1-passes/build.rs` build script (see line 28), so that nothing is required from the user.

tket1-passes/README.md
lmondada: `just` is a build tool very similar to `make` that is used in `tket2` (see `Justfile` for all the recipes). I've added a reference to `DEVELOPMENT.md` that describes how to build tket2 in more details. Is that ok with you or is there something specific you wanted me to clarify?

cqc-melf (approved):

PR #1012, opened 2025-07-31
feat!: Redesign Subcircuit

lmondada:
This might be slightly controversial: I posit that `Subcircuit` in its current form is rather useless: it is just a wrapper around hugr's `SiblingSubgraph`. I suggest to reuse that construct for something slightly different, i.e. a subgraph within a circuit, defined not by the set of nodes that induce it but by the intervals on paths that it covers. This is a more memory-efficient representation, convexity checks are much faster (see https://github.com/CQCL/portgraph/pull/240) and most importantly, it is simple to extend Subcircuits to larger ones.

A subcircuit can (failibly) be converted to a SiblingSubgraph when required.

This PR thus proceeds in two steps (three really, the first one is just a temporary pin of the `portgraph` and `hugr` dependencies to git, pending releases of these packages). See commits:

- removal of `Subcircuit` and replace its use in `CircuitRewrite` with `SiblingSubgraph`
- addition of the "new" `Subcircuit` in `src/subcircuit.rs`.

drive-by: Add <Node> parameter to `Rewrite` trait (also breaking).

This PR will require a (non-breaking) hugr release. Currently pinning portgraph and hugr to main (to be reverted).

### DIFF
--- a/tket/src/rewrite/strategy.rs
+++ b/tket/src/rewrite/strategy.rs
@@ -66,9 +66,7 @@ pub trait RewriteStrategy {
     /// Returns the cost of a rewrite's matched subcircuit before replacing it.
     #[inline]
     fn pre_rewrite_cost(&self, rw: &CircuitRewrite, circ: &Circuit) -> Self::Cost {
-        circ.nodes_cost(rw.subcircuit().nodes().iter().copied(), |op| {
-            self.op_cost(op)
-        })
+        circ.nodes_cost(rw.subgraph().nodes().iter().copied(), |op| self.op_cost(op))
     }
 
     /// Returns the expected cost of a rewrite's matched subcircuit after replacing it.
@@ -129,14 +127,14 @@ impl RewriteStrategy for GreedyRewriteStrategy {
         let mut circ = circ.clone();
         for rewrite in rewrites {
             if rewrite
-                .subcircuit()
+                .subgraph()
                 .nodes()
                 .iter()
                 .any(|n| changed_nodes.contains(n))
             {
                 continue;
             }
-            changed_nodes.extend(rewrite.subcircuit().nodes().iter().copied());
+            changed_nodes.extend(rewrite.subgraph().nodes().iter().copied());
             cost_delta += rewrite.node_count_delta();
             rewrite
                 .apply(&mut circ)
--- a/tket/src/rewrite/strategy.rs
+++ b/tket/src/rewrite/strategy.rs
@@ -478,7 +478,8 @@ mod tests {
     use crate::rewrite::trace::REWRITE_TRACING_ENABLED;
     use crate::{
         circuit::Circuit,
-        rewrite::{CircuitRewrite, Subcircuit},
+        rewrite::CircuitRewrite,
+        subcircuit::{LineConvexChecker, Subcircuit},
         utils::build_simple_circuit,
     };
 
@@ -494,18 +495,20 @@ mod tests {
     }
 
     /// Rewrite cx_nodes -> empty
-    fn rw_to_empty(circ: &Circuit, cx_nodes: impl Into<Vec<Node>>) -> CircuitRewrite {
-        let subcirc = Subcircuit::try_from_nodes(cx_nodes, circ).unwrap();
+    fn rw_to_empty(circ: &Circuit, cx_nodes: impl IntoIterator<Item = Node>) -> CircuitRewrite {
+        let conv_checker = LineConvexChecker::new(circ.hugr(), circ.hugr().entrypoint());
+        let subcirc = Subcircuit::try_from_nodes(cx_nodes, &conv_checker).unwrap();
         subcirc
-            .create_rewrite(circ, n_cx(0))
+            .create_rewrite(n_cx(0), &conv_checker)
             .unwrap_or_else(|e| panic!("{}", e))
     }
 
     /// Rewrite cx_nodes -> 10x CX
-    fn rw_to_full(circ: &Circuit, cx_nodes: impl Into<Vec<Node>>) -> CircuitRewrite {
-        let subcirc = Subcircuit::try_from_nodes(cx_nodes, circ).unwrap();
+    fn rw_to_full(circ: &Circuit, cx_nodes: impl IntoIterator<Item = Node>) -> CircuitRewrite {
+        let conv_checker = LineConvexChecker::new(circ.hugr(), circ.hugr().entrypoint());
+        let subcirc = Subcircuit::try_from_nodes(cx_nodes, &conv_checker).unwrap();
         subcirc
-            .create_rewrite(circ, n_cx(10))
+            .create_rewrite(n_cx(10), &conv_checker)
             .unwrap_or_else(|e| panic!("{}", e))
     }
 
--- a/tket/src/subcircuit.rs
+++ b/tket/src/subcircuit.rs
@@ -1,0 +1,167 @@
+//! Subcircuits of circuits.
+//!
+//! This provides a specialised version of hugr's [`SiblingSubgraph`], for the
+//! case where the HUGR region is a [`Circuit`].
+
+use hugr::hugr::views::sibling_subgraph::{InvalidReplacement, InvalidSubgraph};
+use hugr::hugr::views::SiblingSubgraph;
+use hugr::HugrView;
+use itertools::Itertools;
+use portgraph::algorithms::convex::{LineIndex, LineInterval, LineIntervals};
+
+use crate::circuit::Circuit;
+use crate::rewrite::CircuitRewrite;
+
+pub use hugr::hugr::views::sibling_subgraph::LineConvexChecker;
+
+/// A subcircuit of a circuit.
+///
+/// Store convex subgraphs of Circuit-like regions of HUGRs as intervals on the
+/// "lines" of the circuit (see below and [`LineConvexChecker`]).
+///
+/// Lines are the result of a partition of the DFG
+/// into edge-disjoint paths; in the case of purely quantum ops, the lines are
+/// the qubits
+///
+/// ## Differences with [`SiblingSubgraph`]
+///
+/// [`Subcircuit`]s have two main distinguishing features over
+/// [`SiblingSubgraph`], that make it more suitable in some contexts, especially
+/// when working with subgraphs of circuits:
+///  - The representation of the subcircuit uses the circuit structure to be more
+///    memory efficient. The size of a [`Subcircuit`] instance is not linear in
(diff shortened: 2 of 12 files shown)


lmondada:
We probably don't want breaking releases now, so should we reinstate a "breaking release branch"?

lmondada:
Why Rewriter should be generic on `C`:
- in the future we will need to distinguish `Rewriter` that work with Circuits vs PersistentHugr (not all rewriters will work on both)
- A same rewriter should hopefully work on multiple Cs, so this should be a trait generic rather than an associated type
- `Rewriter` could be generic on the node type `N: HugrNode` rather than the circuit, but this would make `Rewriter` not [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility) (the generic impl HugrView<Node = N> argument in `get_rewrites` is not allowed). Having dyn-compatibility will be useful for collections of rewriters of type `Vec<Box<dyn Rewriter<C>>>`

acl-cqc:
Not got all the way through yet but I've looked at the "remove Subcircuit / use SiblingSubgraph" commit and that all looked good**, my only thought being "make the type parameter on `Rewriter` default to `Node`". However, you then reparametrize Rewriter by CircuitLike in the last commit, which....well, may be unavoidable, but I wonder if there are alternatives.
* The goal of parametrization by CircuitLike is to allow operating on things that *are not HugrView*s (for any Node type), right?
* I had been thinking that a PersistentHugr doesn't implement HugrView itself, but returns a thing that does (`fn as_view(&self) -> impl HugrView`)....but actually I see that PersistentHugr *does* implement HugrView! So there's no problem with having a `Circuit<PersistentHugr>`....
* hence rather than `Rewriter<C: CircuitLike>` can we not have `Rewriter<H:HugrView>` that operates on a `&Circuit<H>` ?

**I guess the other question would be, do we want to "break" Subcircuit like this, or deprecate it, and introduce the new subcircuit in parallel? I think deprecation would mean introducing all the SiblingSubgraph methods alongside Subcircuit equivalents, so I agree this seems like a lot of work, and I don't really want to ask you to do that!! But have you mentioned in standup, let's at least do that?

Now onto the new Subcircuit itself....:-)

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+///    recompute the subgraph from scratch and rechecking convexity.
+///
+/// Along with these features come two caveats to be aware of:
+///  - Unlike subgraphs, [`Subcircuit`]s do not maintain an ordered boundary.
+///    This is because the ordering would be hard to maintain when the subcircuit
+///    is expanded. When a subcircuit is converted into a [`SiblingSubgraph`],
+///    a boundary order gets fixed.
+///  - Constructing a [`LineConvexChecker`] (required to construct [`Subcircuit`]s)
+///    is currently ~2.5x more expensive than constructing the default
+///    [`TopoConvexChecker`]. This must be amortised over many subcircuit
```
acl-cqc, line 45:
```suggestion
///    [`TopoConvexChecker`]. This may be amortised over many subcircuit
```
(the caller does that by reusing the `LineConvexChecker` instance)

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+///    memory efficient. The size of a [`Subcircuit`] instance is not linear in
+///    the number of nodes in the subgraph, but linear in the number of lines
+///    (~qubits) in the subcircuit.
+///  - Subcircuits can be updated by extending the intervals without having to
+///    recompute the subgraph from scratch and rechecking convexity.
+///
+/// Along with these features come two caveats to be aware of:
+///  - Unlike subgraphs, [`Subcircuit`]s do not maintain an ordered boundary.
+///    This is because the ordering would be hard to maintain when the subcircuit
+///    is expanded. When a subcircuit is converted into a [`SiblingSubgraph`],
```
acl-cqc, line 41: I take it there is no handy ordering of lines in a LineConvexChecker that could be used here?

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+///
+/// [`Subcircuit`]s have two main distinguishing features over
+/// [`SiblingSubgraph`], that make it more suitable in some contexts, especially
+/// when working with subgraphs of circuits:
+///  - The representation of the subcircuit uses the circuit structure to be more
+///    memory efficient. The size of a [`Subcircuit`] instance is not linear in
+///    the number of nodes in the subgraph, but linear in the number of lines
+///    (~qubits) in the subcircuit.
+///  - Subcircuits can be updated by extending the intervals without having to
+///    recompute the subgraph from scratch and rechecking convexity.
```
acl-cqc, line 36:
```suggestion
///    recompute the subgraph from scratch or rechecking convexity.
```
(we're within scope of "without having to")

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+    ) -> Result<Self, InvalidSubgraph<H::Node>> {
+        let intervals = checker
+            .get_intervals_from_nodes(nodes.into_iter())
+            .ok_or(InvalidSubgraph::NotConvex)?;
+        Ok(Self { intervals })
+    }
+
+    /// Create a new subcircuit from a set of intervals.
+    ///
+    /// Warning: this does not check that the intervals represent a valid
```
acl-cqc, line 81: Hmmm ok so `checker.get_intervals_from_nodes` must return either a checked/valid Intervals, or fails. But here we see that invalid `LineIntervals` also exist - there's nothing in the signature of `checker.get_intervals_from_nodes` to suggest that there's anything special about those LineIntervals that mean this isn't `unchecked`, I suspect there should be....

(i.e. contrast `try_from_nodes(nodes, checker)` with `from_intervals_unchecked(checker.get_intervals_from_nodes(nodes).unwrap())` - what info have we lost that the latter has to be "unchecked"?)

This suggests that we might want a way to verify whether a LineIntervals is valid, and/or fixes on the Hugr side, so this seems reasonable (here in tket2) for now.

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Create a new empty subcircuit.
+    pub fn new_empty() -> Self {
+        Self {
+            intervals: LineIntervals::default(),
+        }
+    }
+
+    /// Nodes in the subcircuit.
+    pub fn nodes<'a, 'g: 'a, H: 'g + HugrView>(
```
acl-cqc, line 95: Yeah, is there a possibility of keeping a handle to the LineConvexChecker? AFAICS the latter is immutable so...

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Convert the subcircuit to a [`SiblingSubgraph`].
+    pub fn try_to_subgraph<H: HugrView>(
+        &self,
+        checker: &LineConvexChecker<H>,
+    ) -> Result<SiblingSubgraph<H::Node>, InvalidSubgraph<H::Node>> {
+        let Self { intervals } = self;
+        SiblingSubgraph::try_from_nodes_with_intervals(
+            checker.nodes_in_intervals(intervals).collect_vec(),
```
acl-cqc, line 129: nit: I'd just use `self.intervals` here and drop the deconstruction

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+    ///
+    /// # Parameters
+    /// * `circuit` - The base circuit that contains the subcircuit.
+    /// * `replacement` - The new circuit to replace the subcircuit with.
+    pub fn create_rewrite(
+        &self,
+        replacement: Circuit<impl HugrView<Node = hugr::Node>>,
+        checker: &LineConvexChecker<impl HugrView<Node = hugr::Node>>,
+    ) -> Result<CircuitRewrite<hugr::Node>, InvalidReplacement> {
+        // The replacement must be a Dfg rooted hugr.
```
acl-cqc, line 145: Is this guaranteed by the fact that `replacement` is a `Circuit`? If so, probably drop the comment.

If it's not guaranteed, then (a) put this in the method doc, (b) how are you making sure that it is - by SiblingSubgraph::try_from? (Seems fair)

acl-cqc (commented):
Nice, thanks Luca! Most of the work/algorithm etc. is done by LineConvexChecker rather than this PR, so not a lot to say here - but mostly (I'm gonna give you a little bit of time to think about this) is there a way to avoid passing LineConvexChecker in every time? Either borrow the thing in the Subcircuit struct, or can you borrow or even Arc some part of the LineConvexChecker (/result of `.init_checker`) ?

tket/src/rewrite/strategy.rs
```diff
@@ -496,18 +495,20 @@ mod tests {
     }
 
     /// Rewrite cx_nodes -> empty
-    fn rw_to_empty(circ: &Circuit, cx_nodes: impl Into<Vec<Node>>) -> CircuitRewrite {
-        let subcirc = Subcircuit::try_from_nodes(cx_nodes, circ).unwrap();
+    fn rw_to_empty(circ: &Circuit, cx_nodes: impl IntoIterator<Item = Node>) -> CircuitRewrite {
+        let conv_checker = LineConvexChecker::new(circ.hugr(), circ.hugr().entrypoint());
+        let subcirc = Subcircuit::try_from_nodes(cx_nodes, &conv_checker).unwrap();
```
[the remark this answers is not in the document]
lmondada: I am changing how the convex checker will be used/defined, so should solve this.

tket/src/subcircuit.rs
```diff
@@ -0,0 +1,167 @@
...
+/// Lines are the result of a partition of the DFG
+/// into edge-disjoint paths; in the case of purely quantum ops, the lines are
+/// the qubits
+///
+/// ## Differences with [`SiblingSubgraph`]
+///
+/// [`Subcircuit`]s have two main distinguishing features over
+/// [`SiblingSubgraph`], that make it more suitable in some contexts, especially
+/// when working with subgraphs of circuits:
+///  - The representation of the subcircuit uses the circuit structure to be more
```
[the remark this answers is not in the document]
lmondada: I'll rephrase this :)

tket/src/subcircuit.rs
lmondada: An interval will be invalid if it was computed for a different HUGR (just like an index into a vector X may be invalid when using it to index Y).

I agree that the API is confusing, will work on this too.

lmondada:
I'll close this PR for now as it needs a significant redesign. Thanks Alan for the review, I have integrated all the comments into my branch for the time being.

PR #1030, opened 2025-08-12
feat!: Reworked pytket decoder framework

aborgna-q:
- Adds a `PytketDecoder` trait that let us extend the decoding framework to external extensions.
- Translations are no longer 1-to-1 between pytket commands and hugr ops, the decoding call has access to  a hugr builder and its context.
- Allows tracking wires with complex/composite types. We don't provide helpers to convert types yet, it's up to the extension emitter to look at the incoming wire types. (Note: I splitted this change to a previous PR, #1036)

TODO:
- [x] Implement the decoder trait for all standard extensions.
- [x] Fix remaining test errors.
- [ ] Add a nicer guide for writing extension traits in the module docs.

BREAKING CHANGE: Split pytket encoder/decoder errors into separate structs.


aborgna-q:
Marking this as ready for release.
There's multiple items left for followup PRs:
- Add an implementors guide to the `tket::serialize::pytket::extension` docs.
- Add a `qsystem` encoder/decoder and a default config containing it
- Add more roundtrip tests
- Decode unsupported hugrs from inside pytket barriers
- Let the extension emiters define automatic type conversions (e.g. array destructuring, unpack tuple, etc.)

tket/src/serialize/pytket/config/type_translators.rs
```diff
@@ -1,6 +1,6 @@
 //! Sets of extensions for translating HUGR types into pytket registers.
 //!
-//! Defines sets of [`TypeTranslator`]s used by the encoders and decoders to
+//! Defines sets of [`PytketTypeTranslator`]s used by the encoders and decoders to
```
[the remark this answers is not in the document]
aborgna-q: The change already uses "PytketTypeTranslator" ?

tket/src/serialize/pytket/extension.rs
```diff
@@ -88,6 +96,43 @@ pub trait PytketEmitter<H: HugrView> {
...
+pub trait PytketDecoder {
+    /// A list of pytket's [`tket_json_rs::OpType`] supported by this decoder.
+    ///
+    /// [`PytketDecoder::op_to_hugr`] will only be called for commands
+    /// containing these.
+    fn op_types(&self) -> Vec<tket_json_rs::OpType>;
+
+    /// Given a pytket [`tket_json_rs::circuit_json::Operation`] node in the
+    /// HUGR circuit and its operation type, try to convert it to a pytket
+    /// operation and add it to the pytket encoder.
```
lmondada, line 113: add it to the pytket _decoder_.

tket/src/serialize/pytket/decoder.rs
```diff
@@ -112,259 +198,443 @@ impl Tk1DecoderContext {
...
             }
-        }
+            .wrap()
+            .hugr_op(&op));
+        };
+
+        // Gather the input wires, with the types needed by the operation.
+        let input_wires =
+            self.find_typed_wires(sig.input_types(), qubits.iter(), bits.iter(), params)?;
+        debug_assert_eq!(op_input_count, input_wires.register_count());
```
lmondada, line 527: Why is this definition of `qubits` different from the one in inputs above?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -112,259 +198,443 @@ impl Tk1DecoderContext {
...
+        // Qubit registers get reused for both input and output, bit registers are not.
+        let output_qubits = qubits.iter().take(op_output_count.qubits).cloned();
+        let output_bits = bits
+            .iter()
+            .skip(op_input_count.bits)
+            .take(op_output_count.bits)
+            .cloned();
 
-        Ok((inputs, outputs))
+        self.register_node_outputs(node.node(), output_qubits, output_bits)?;
```
lmondada, line 545: Am I right to think that if the number of input qubits is larger than the number of output qubits, then the qubits not in the output would not get invalidated in the tracker, even though they have been consumed by an op?

This is obviously a pathological case (I don't think there are (nor can be) pytket ops like this). I just wonder if we could model the properties of pytket ops (and also most tket quantum ops) better. E.g. the guarantee that there are as many input as output qubit types would be useful in my pattern matching APIs too.

For your tracked wires, you'd then "borrow" quantum wires and would return them before the end of the scope, whereas classical wires would just be consumed or produced.

This is definitely out of scope for this PR, just wanted your thoughts on it :)

tket/src/serialize/pytket/extension.rs
```diff
@@ -88,6 +96,43 @@ pub trait PytketEmitter<H: HugrView> {
     }
 }
 
+/// A decoder of pytket operations and types that transforms them into HUGR
+/// primitives.
+///
+/// A [decoder configuration](crate::serialize::pytket::PytketDecoderConfig)
+/// contains a list of such decoders.
+pub trait PytketDecoder {
```
lmondada, line 104: Can you add a mention of `PytketDecoder` to the module docs?

tket/src/serialize/pytket/extension/prelude.rs
```diff
@@ -1,15 +1,20 @@
...
 use crate::Circuit;
-use hugr::extension::prelude::{BarrierDef, TupleOpDef, PRELUDE_ID};
+use hugr::extension::prelude::{bool_t, qb_t, BarrierDef, Noop, TupleOpDef, PRELUDE_ID};
 use hugr::extension::simple_op::MakeExtensionOp;
 use hugr::extension::ExtensionId;
-use hugr::ops::ExtensionOp;
+use hugr::ops::{ExtensionOp, OpType};
 use hugr::types::TypeArg;
 use hugr::HugrView;
 use tket_json_rs::optype::OpType as Tk1OpType;
```
lmondada, line 20: Rename to `PytketOpType`?

tket/src/serialize/pytket/extension/bool.rs
```diff
@@ -1,14 +1,19 @@
...
 use super::PytketEmitter;
 use crate::extension::bool::{BoolOp, ConstBool, BOOL_EXTENSION_ID, BOOL_TYPE_NAME};
 use crate::serialize::pytket::config::TypeTranslatorSet;
+use crate::serialize::pytket::decoder::{
+    DecodeStatus, LoadedParameter, PytketDecoderContext, TrackedBit, TrackedQubit,
+};
 use crate::serialize::pytket::encoder::{
     make_tk1_classical_expression, make_tk1_classical_operation, EncodeStatus,
     PytketEncoderContext, TrackedValues,
 };
```
lmondada, line 14: Whilst we are breaking stuff, rename `Tk1ConvertError` to `PytketConvertError`.

lmondada (changes requested):
Very cool!

lmondada:
Urgh, I used github's PR VSCode extension, and it's very buggy. It lost some of the comments I've made. I've added them back myself, as far as I could remember, hopefully I didn't miss anything important :(

lmondada:
Coverage seems to say that `impl PytketDecoder for BoolEmitter` is not tested? Might be worth adding a test for that...

tket/src/serialize/pytket/decoder.rs
aborgna-q: Yeah, this needs some thinking 🤔 

I think a breaking example would be a destructive measurement followed by a reset operation.

We wouldn't update the input register after the measurement, so if the reset tries to (incorrectly) fetch the wire for the named qubit it would get an already used data. This shouldn't happen in practice, but it's a bug waiting to happen in user's code.

We could add a "mark outdated" call on `WireTracker` and check the flag when getting the latest tracked value for a register.

tket/src/serialize/pytket/extension.rs
aborgna-q: Added a paragraph. We'll expand the module docs in a followup PR

tket/src/serialize/pytket/decoder.rs
lmondada: awesome, thanks!

tket/src/serialize/pytket/extension/prelude.rs
lmondada: ok got it, but this would just be a file-local rename... You can probably do it 🙃

tket/src/serialize/pytket/decoder.rs
```diff
@@ -56,94 +57,68 @@ impl<'h> PytketDecoderContext<'h> {
...
-            FunctionBuilder::with_hugr(hugr, name, sig).unwrap();
+            FunctionBuilder::with_hugr(hugr, name, signature.clone()).unwrap();
 
         Self::init_metadata(&mut dfg, serialcirc);
-        let wire_tracker =
-            Self::init_wire_tracker(serialcirc, &mut dfg, &input_types, input_params, &config)?;
+        let wire_tracker = Self::init_wire_tracker(
+            serialcirc,
+            &mut dfg,
+            &signature.input,
```
lmondada, line 118: nit: you can probably recover the input types from the dfg, and avoid a `signature.clone()` above?

tket/src/serialize/pytket/extension.rs
```diff
@@ -110,7 +114,7 @@ pub trait PytketDecoder {
 
     /// Given a pytket [`tket_json_rs::circuit_json::Operation`] node in the
     /// HUGR circuit and its operation type, try to convert it to a pytket
-    /// operation and add it to the pytket encoder.
+    /// operation and add it to the pytket decoder.
     ///
     /// Returns an [`DecodeStatus`] indicating if the operation was successfully
     /// converted. If the operation is not supported by the encoder, it's
```
lmondada, line 120:
```suggestion
    /// converted. If the operation is not supported by the decoder, it's
```

lmondada: also on the two lines below.

lmondada (approved):
Awesome, thanks a lot!

Just one more suggestion: consider using "Resource" instead of "Wire", when referring to _pytket wire_s. E.g. `ResourceTracker` instead of `WireTracker`.

Would be good if "wire" only referred to HUGR wires :)

tket/src/serialize/pytket/decoder.rs
aborgna-q: We're passing `&mut dfg` around, so we can't have a reference to the signature -.-

aborgna-q:
> Awesome, thanks a lot!
>
> Just one more suggestion: consider using "Resource" instead of "Wire", when referring to pytket wires. E.g. ResourceTracker instead of WireTracker.
>
> Would be good if "wire" only referred to HUGR wires :)

`WireTracker` is a mix of both, it tracks HUGR wires that carry pytket resources.

We could use some documentation standarisation about those terms, but thankfully they are all private types so we can rename them later 🤷

PR #1037, opened 2025-08-14
refactor!: Rename tk2 encoder names to tket

aborgna-q:
No functionality changes, only renames that we didn't do in #987 

I'm not touching the decoder code, since it's already being modified by #1030

BREAKING CHANGE: Renamed various pytket encoder-related structs and methods from `tk1*` to `pytket`


cqc-alec:
Looks like there are some merge conflicts.

cqc-alec (approved):

PR #1041, opened 2025-08-15
feat!: Use qsystem encoder/decoders in tket-py

aborgna-q:
Depends on #1039

- Adds an optional config parameter to all the `load_tk1` / `save_tk1` utilities, so we don't use the default by mistake.
- Uses the new qsystem encoder/decoders in `tket-py`, so now we can load `ZZMax`/`ZZPhase`/`PhasedX`

BREAKING CHANGE: Added encode/decode config arguments to pytket load/store utilities

This is not a breaking change for the python lib, so I'll override the flag for release-please:

BEGIN_COMMIT_OVERRIDE

feat: Support qsystem native operations when loading pytket circuits (#1041)

END_COMMIT_OVERRIDE

### DIFF
--- a/tket-py/test/test_circuit.py
+++ b/tket-py/test/test_circuit.py
@@ -54,3 +54,25 @@ def test_conversion():
 
     assert tk1_back == tk1
     assert type(tk1_back) is Circuit
+
+
+def test_conversion_qsystem():
+    tk1 = Circuit(2).ZZPhase(0.75, 0, 1).PhasedX(0.25, 0.33, 1)
+    tk1_dot = render_circuit_dot(tk1)
+
+    tk2 = Tk2Circuit(tk1)
+    tk2_dot = render_circuit_dot(tk2)
+
+    assert type(tk2) is Tk2Circuit
+    assert tk1_dot == tk2_dot
+
+    # Check that we didn't use the opaque tk1 op fallback.
+    # TODO: There's no way to traverse the circuit nodes??
+    assert "TKET1.tk1op" not in tk1_dot
+    assert "tket.qsystem.PhasedX" in tk1_dot
+    assert "tket.qsystem.ZZPhase" in tk1_dot
+
+    tk1_back = tk2.to_tket1()
+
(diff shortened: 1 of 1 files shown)


tket-py/test/test_circuit.py
```diff
@@ -54,3 +54,23 @@ def test_conversion():
...
+
+    tk2 = Tk2Circuit(tk1)
+    tk2_dot = render_circuit_dot(tk2)
+
+    assert type(tk2) is Tk2Circuit
+    assert tk1_dot == tk2_dot
+
+    # Check that we didn't use the opaque tk1 op fallback.
+    # TODO: There's no way to traverse the circuit nodes??
+    assert "TKET1.tk1op" not in tk1_dot
```
aborgna-q, line 71: x.x

At least we can check that we're doing the right thing in the render

```mermaid
graph LR
    subgraph 0 ["(0) Module"]
        direction LR
        subgraph 1 ["(1) [**FuncDefn: #quot;#quot;**]"]
            direction LR
            2["(2) Input"]
            3["(3) Output"]
            4["(4) const:custom:f64(0.75)"]
            5["(5) LoadConstant"]
            6["(6) tket.qsystem.ZZPhase"]
            7["(7) const:custom:f64(0.25)"]
            8["(8) LoadConstant"]
            9["(9) const:custom:f64(0.33)"]
            10["(10) LoadConstant"]
            11["(11) tket.qsystem.PhasedX"]
            2--"0:0<br>qubit"-->6
            2--"1:1<br>qubit"-->6
            4--"0:0<br>float64"-->5
            5--"0:2<br>float64"-->6
            6--"0:0<br>qubit"-->3
            6--"1:0<br>qubit"-->11
            7--"0:0<br>float64"-->8
            8--"0:1<br>float64"-->11
            9--"0:0<br>float64"-->10
            10--"0:2<br>float64"-->11
            11--"0:1<br>qubit"-->3
        end
    end
```

aborgna-q:
I should probably split out the python changes, since there's no breaking change there

Issue #1034, opened 2025-08-13
Build and publish python docs


aborgna-q:
See hugr docs [workflow](https://github.com/CQCL/hugr/blob/main/.github/workflows/docs.yml).

We would like to have this mainly for the main `tket` package.
We re-export all of `tket-exts` there, so a single project docs should be enough.

PR #1046, opened 2025-08-18
feat: Explicit exports for tket_exts ops and types

### LINKED ISSUE #925: (tket2-exts): Export ops and types for each extension
Currently to use a quantum operation from python, one must resort to stringly typed lookups:

```python
QUANTUM_EXTENSION = tket2_exts.quantum()
QAlloc = QUANTUM_EXTENSION.get_op("QAlloc").instantiate()

...

fn.add_op(QAlloc, ...)
```

This is quite clunky and error-prone.
We should try and export the names here, ensuring we test them agains the exported json to make sure the names are correct.

aborgna-q:
Closes #925

Lists each extension's operations and types, so they appear in the docs and users don't have to guess what ops are available.

- This is a backwards compatible change. `tket_exts.opaque_bool` has been deprecated in favour of `tket_exts.bool`.

TODO:
- [x] Add missing ops and types (I've only done bool atm)
- [ ] ~See how this renders on the docs.~ (waiting on #1034)


tket-exts/src/tket_exts/tket/bool.py
```diff
@@ -0,0 +1,78 @@
...
+        return load_extension("tket.bool")
+
+    def TYPES(self) -> List[TypeDef]:
+        """Return the types defined by this extension"""
+        return [self.bool_t.type_def]
+
+    def OPS(self) -> List[OpDef]:
+        """Return the operations defined by this extension"""
+        return [
+            self.and_op.op_def(),
```
croyzor, line 31: nit: This isn't consistent with the other extensions, which define methods do get the OpDef and methods for the ExtOp

croyzor (approved):

PR #1052, opened 2025-08-19
feat: Add ResourceScope

lmondada:
Alright, couldn't resist it. Here's a PR that moves a lot of the logic that I tried to put into `portgraph` into `tket`. I think it makes much more sense this way.

I'm interested to see what you think! (a lot of the diff is docs and snapshot tests, so don't worry :P)


lmondada:
Three things on my mind
- could we make this share more code/structs with the CircuitBuilder, esp. CircuitUnit?
- idem, with the pytket encoder/decoder
- imo this is a candidate to replace the Circuit struct if we get rid of it 😅

acl-cqc:
I mean, if this went into Portgraph, it'd have to do so without any mention of linearity, so probably leaving out any (useful!) implementation of the ResourceFlow trait entirely; that would then have to be done in tket...

lmondada:
I will address your comments one by one. Before that though, one question:

I could replace `CopyableValueID` with just `Wire` - would that resolve your qualms? That begs the question: should I just use the `CircuitUnit` Hugr type? It's isomorphic to the type I need, but I was wondering:
- is it worth having a tket-owned type for this?
- this will be exposed in the Python API. Is `CircuitUnit` a good name? Is `OpValue` better? Is there something else we should consider
- Using `CircuitUnit::Linear` means resources are represented by a `usize`, rather than a distinct type index. That's not great -- but does it justify another type?

Of course, conversions between `CircuitUnit` and whatever we'd use here could be implemented easily. The advantage the same type in both cases would be that someone familiar with one of them immediately recognises the type in the other context...

tket/src/resource.rs
```diff
@@ -0,0 +1,160 @@
...
+//! resource-preserving ops acting on a same resource form a so-called resource
+//! path.
+//!
+//! # Resources and Copyable Values
+//!
+//! Resource tracking distinguishes between two types of values:
+//!
+//! - **Linear resources**: Non-copyable values that form resource paths through
+//!   the circuit. Each resource has a unique [`ResourceId`] and operations on
+//!   the same resource are ordered by a [`Position`].
```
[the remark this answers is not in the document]
lmondada: Added, thank you!

tket/src/resource/flow.rs
```diff
@@ -0,0 +1,136 @@
...
+    fn map_resources(
+        &self,
+        op: &OpType,
+        inputs: &[Option<ResourceId>],
+    ) -> Result<Vec<Option<ResourceId>>, UnsupportedOp>;
+}
+
+/// Default implementation of ResourceFlow.
+///
+/// This implementation considers that an operation is resource-preserving if
```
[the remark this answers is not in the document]
lmondada: Yes, took the latter and changed it a bit. See if you like the new docs.

tket/src/resource/interval.rs
```diff
@@ -0,0 +1,294 @@
...
+                    Err(InvalidInterval::NotContiguous(node))
+                }
+            }
+        }
+    }
+
+    /// Include the given node in the interval.
+    ///
+    /// Does not check if the node is contiguous with the interval.
+    pub(crate) fn include_node(&mut self, node: N, pos: Position) {
```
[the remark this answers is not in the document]
lmondada: I've gone with `add_node_unchecked`.

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,596 @@
...
+    /// node in the given direction.
+    pub fn resource_path_iter(
+        &self,
+        resource_id: ResourceId,
+        start_node: H::Node,
+        direction: Direction,
+    ) -> impl Iterator<Item = H::Node> + '_ {
+        let mut curr_node = start_node;
+
+        iter::from_fn(move || {
```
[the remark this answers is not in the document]
lmondada: oh, super neat!! Did not know about it!

lmondada:
Having slept on it, I will try to use the CircuitUnit enum. We might regret it if we need to make changes to it later, but that's fien.

lmondada:
Note: I've had to change `trait ResourceFlow` to be parametrised on `H: HugrView`. The reason for this is that I have generalised the `fn map_resources` function from taking an `OpType` to taking a `Node` and `HugrView`:

```rust
    fn map_resources(
        &self,
        node: H::Node,
        hugr: &H,
        inputs: &[Option<ResourceId>],
    ) -> Result<Vec<Option<ResourceId>>, UnsupportedOp>;
```

This was required to implement `ResourceFlow` e.g. on `Call` nodes, where the flow does not just depend on the optype, but on the function definition that is linked in.

I chose `ResourceFlow<H>` rather than `impl HugrView` in the function definition because the former trait is dyn-compatible, the latter is not.

tket/src/resource/interval.rs
```diff
@@ -0,0 +1,302 @@
...
+        scope: &ResourceScope<impl HugrView<Node = N>>,
+    ) -> Self {
+        let start_pos = scope
+            .get_position(start_node)
+            .expect("start node not on resource path");
+        let end_pos = scope
+            .get_position(end_node)
+            .expect("end node not on resource path");
+
+        assert!(start_pos <= end_pos);
```
acl-cqc, line 54: The other thing to check is that `end_node` is after `start_node`, rather than before it.

I might favour a `try_new` (maybe just returning an Option if you feel lazy); `fn new(...) {try_new(...).unwrap()}` is very simple if you want that as well (maybe you do!), and document the conditions that it asserts/panics.

tket/src/resource/interval.rs
```diff
@@ -0,0 +1,302 @@
...
+            self.positions[1] = pos;
+            self.nodes[1] = node;
+        }
+    }
+
+    /// Whether `pos` is smaller, larger or within the interval.
+    ///
+    /// [Ordering] is used to express the relative position of `pos` with
+    /// respect to the interval:
+    /// - [`Ordering::Less`] if `pos` is smaller than the interval,
```
acl-cqc, line 152:
```suggestion
    /// - [`Ordering::Less`] if `pos` is before the interval,
```

tket/src/resource/interval.rs
```diff
@@ -0,0 +1,302 @@
...
+    /// - [`Ordering::Greater`] if `pos` is larger than the interval, and
```
acl-cqc, line 153:
```suggestion
    /// - [`Ordering::Greater`] if `pos` is after the interval, and
```

tket/src/resource/interval.rs
```diff
@@ -0,0 +1,302 @@
...
+        let cx_nodes = subgraph.nodes().to_owned();
+        let scope = super::ResourceScope::new(&circ, subgraph);
+
+        let resource_id = ResourceId::new(0);
+
+        // Create an interval from nodes 2 to 3 (middle of circuit)
+        let mut interval = Interval::new(resource_id, cx_nodes[2], cx_nodes[3], &scope);
+
+        // Apply the test case
+        let result = interval.try_extend(cx_nodes[node_to_extend], &scope);
```
acl-cqc, line 283: Given the brevity of the actual test case (here), I think I might separate the test into two (success cases, failure cases) with a common setup method doing all the prep. before this; it'd simplify the `match expected_result` and might even remove the `unimplemented!`

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,613 @@
...
+        let port_map = self.port_map(node)?;
+        Some(port_map.get_slice(direction))
+    }
+
+    /// Get the port of node on the given resource path.
+    ///
+    /// The returned port will have the direction `dir`.
+    pub fn get_port(&self, node: H::Node, resource_id: ResourceId, dir: Direction) -> Option<Port> {
+        let units = self.get_circuit_units_slice(node, dir)?;
+        let offset = units.iter().position(|unit| match unit {
```
acl-cqc, line 145:
```suggestion
        let offset = units.iter().position(|unit| unit.as_resource() == Some(resource_id))
```

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,613 @@
...
+    }
+
+    /// All resource IDs on the ports of `node` in the given direction.
+    pub fn get_resources(
+        &self,
+        node: H::Node,
+        dir: Direction,
+    ) -> impl Iterator<Item = ResourceId> + '_ {
+        let units = self.get_circuit_units_slice(node, dir);
+        units.into_iter().flatten().filter_map(|unit| match unit {
```
acl-cqc, line 166: ditto using `as_resource`

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,596 @@
...
+            self.subgraph.incoming_ports().to_owned(),
+            &mut allocator,
+            sentinel,
+        );
+
+        // Proceed to propagating the op values through the subgraph, in topological
+        // order.
+        for node in toposort_subgraph(&self.hugr, &self.subgraph, self.find_sources()) {
+            if self.hugr.get_optype(node).dataflow_signature().is_none() {
+                // ignore non-dataflow ops
```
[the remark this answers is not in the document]
acl-cqc: No...certainly they feel like no-ops, and I think they are. I wonder whether the `continue` is necessary, will the skipped code actually do anything?

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,613 @@
...
+        flows: &[Box<dyn '_ + ResourceFlow<H>>],
+        allocator: &mut CircuitUnitAllocator,
+    ) {
+        let port_map =
+            &mut node_circuit_units_mut(&mut self.circuit_units, node, &self.hugr).port_map;
+
+        let inp_resources = port_map
+            .get_slice(Direction::Incoming)
+            .iter()
+            .map(|&op_val| match op_val {
```
acl-cqc, line 366:
```suggestion
            .map(CircuitUnit::as_resource)
```

tket/src/resource/types.rs
```diff
@@ -0,0 +1,246 @@
...
+impl<N: HugrNode> From<CircuitUnit<N>> for hugr::CircuitUnit<N> {
+    fn from(value: CircuitUnit<N>) -> Self {
+        match value {
+            CircuitUnit::Resource(resource_id) => Self::Linear(resource_id.as_usize()),
+            CircuitUnit::Copyable(wire) => Self::Wire(wire),
+        }
+    }
+}
+
+impl<N: HugrNode> From<hugr::CircuitUnit<N>> for CircuitUnit<N> {
```
acl-cqc, line 96: This rather depends on the ResourceIds having been allocated exactly as you expect (i.e. the same ordering as the Hugr convention), right? Would it be better as a method on ResourceScope (that can look up in the map for the Input node)?

tket/src/resource/types.rs
```diff
@@ -0,0 +1,246 @@
...
+            hugr::CircuitUnit::Linear(resource_id) => {
+                CircuitUnit::Resource(ResourceId::new(resource_id))
+            }
+        }
+    }
+}
+
+impl<N: HugrNode> CircuitUnit<N> {
+    /// Returns true if this is a resource value.
+    pub fn is_resource(&self) -> bool {
```
acl-cqc, line 109: not used - and `as_resource().is_some()` works almost as well...

acl-cqc (commented):
Looking good! I suspect I won't make any more comments but I'll send now just to give myself time to think ;-). Or @tatiana-s do you want to take a look?

tket/src/resource/flow.rs
```diff
@@ -0,0 +1,171 @@
...
+/// corresponding output. (All outputs with no corresponding input, must be
+/// copyable.)
+///
+/// If on the other hand an operation is not resource preserving, all input
+/// resources are discarded and all outputs will be given fresh resource IDs.
+#[derive(Debug, Clone, Default)]
+pub struct DefaultResourceFlow;
+
+impl DefaultResourceFlow {
+    /// Create a new DefaultResourceFlow instance.
```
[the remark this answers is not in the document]
lmondada: removed!

tket/src/resource/types.rs
lmondada: I'll leave this, I don't see any harm in it.

tket/src/resource/flow.rs
```diff
@@ -0,0 +1,165 @@
...
+    ///   types)
+    fn map_resources(
+        &self,
+        node: H::Node,
+        hugr: &H,
+        inputs: &[Option<ResourceId>],
+    ) -> Result<Vec<Option<ResourceId>>, UnsupportedOp>;
+
+    /// Convert this ResourceFlow into a boxed trait object.
+    fn into_boxed<'a>(self) -> Box<dyn 'a + ResourceFlow<H>>
```
acl-cqc, line 60:
```suggestion
    fn into_boxed<'a>(self) -> Box<dyn 'a + Self>
```

acl-cqc: Maybe `Box<dyn '_ + Self` works without the `<'a>` ?

tket/src/resource/flow.rs
```diff
@@ -0,0 +1,165 @@
...
+        self.as_ref().map_resources(node, hugr, inputs)
+    }
+}
+
+/// Default implementation of ResourceFlow.
+///
+/// This implementation considers that an operation is resource-preserving if
+/// for all port indices i, either
+///  - the i-th input and i-th output are both linear and are of the same type,
+///  - or the i-th input and i-th output are both copyable, or one is copyable
```
acl-cqc, line 84:
```suggestion
///  - or the i-th input and i-th output are both copyable,
///  - or one is copyable and the other does not exist
```

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,621 @@
...
+    /// resource.
+    pub fn is_resource_start(&self, node: H::Node, resource_id: ResourceId) -> bool {
+        self.get_port(node, resource_id, Direction::Outgoing)
+            .is_some()
+            && self
+                .get_port(node, resource_id, Direction::Incoming)
+                .is_none()
+    }
+
+    /// Iterate over all resources in the scope.
```
acl-cqc, line 193:
```suggestion
    /// Iterate over all distinct resources in the scope.
```

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,621 @@
...
+        Self {
+            port_map: PortMap::with_default(default, signature),
+            position: Position::default(),
+        }
+    }
+}
+
+/// Configuration for a ResourceScope.
+pub struct ResourceScopeConfig<'a, H: HugrView> {
+    flows: Vec<Box<dyn 'a + ResourceFlow<H>>>,
```
acl-cqc, line 61: Might be worth documenting that these will be tried in order (it wasn't obvious to me that the second would only be used if the first failed)

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,621 @@
...
+                    continue;
+                };
+                next_node_units.port_map.set(in_port, unit);
+                next_node_units.position = cmp::max(next_node_units.position, pos.increment());
+            }
+        }
+    }
+}
+
+/// Get the circuit units for the given node, creating them if they don't exist.
```
acl-cqc, line 427: If they don't exist then create them....so when do we return `None`? Ah - for non-dataflow ops

tket/src/resource/scope.rs
```diff
@@ -0,0 +1,621 @@
...
+    match all_circuit_units.entry(node) {
+        Entry::Occupied(occupied_entry) => Some(occupied_entry.into_mut()),
+        Entry::Vacant(vacant_entry) => {
+            let signature = hugr.get_optype(node).dataflow_signature()?;
+            vacant_entry
+                .insert(NodeCircuitUnits::with_default(
+                    CircuitUnit::sentinel(),
+                    &signature,
+                ))
+                .into()
```
acl-cqc, line 442: The `.into` adds the `Some` here? I think it might be clearer to do `Some(...)` explicitly...

tket/src/resource/types.rs
```diff
@@ -0,0 +1,226 @@
...
+    }
+
+    /// Get position as f64, rounded to the given precision.
+    pub fn to_f64(&self, precision: usize) -> f64 {
+        let big = self.0 * Rational64::from_integer(10).pow(precision as i32);
+        big.round().to_integer() as f64 / 10f64.powi(precision as i32)
+    }
+
+    /// Increment the position by 1.
+    pub fn increment(&self) -> Self {
```
acl-cqc, line 65: You might want `add_int(&self, usize) -> Self`

tket/src/resource/types.rs
```diff
@@ -0,0 +1,226 @@
...
+        Self(self.0 + 1)
+    }
+}
+
+/// A value associated with a dataflow port, identified either by a resource ID
+/// (for linear values) or by its wire (for copyable values).
+///
+/// This can currently be converted to and from [`hugr::CircuitUnit`], but
```
acl-cqc, line 73: Remove this comment now (I think conversions in both directions are gone)

tket/src/resource/types.rs
```diff
@@ -0,0 +1,226 @@
...
+impl ResourceId {
+    /// Create a new ResourceId.
+    ///
+    /// This method should only be called by ResourceAllocator and tests.
+    pub(super) fn new(id: usize) -> Self {
+        Self(id)
+    }
+
+    /// Get the underlying usize value of this ResourceId.
+    pub fn as_usize(self) -> usize {
```
acl-cqc, line 30: I note this isn't used!! I guess it's reasonable to have it for indexing into an array etc.

tket/src/resource/types.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    /// Allocate the next available ResourceId.
+    pub fn allocate(&mut self) -> ResourceId {
+        let id = ResourceId::new(self.next_id);
+        self.next_id += 1;
+        id
+    }
+}
+
+impl Default for ResourceAllocator {
```
acl-cqc, line 222: I *think* you'll get this if you just derive Default, one could even argue there's no need for `::new()`

acl-cqc (approved):
Great stuff Luca, thanks for all the changes.

These are all 1-liner "nits" but there are a couple of wrong comments etc.

tket/src/resource.rs
```diff
@@ -0,0 +1,177 @@
...
+//! circuits represented as HUGR subgraphs.
+//!
+//! # Overview
+//!
+//! As far as HUGR is concerned, a linear value (or any value, for that matter)
+//! is created at an op's output and destroyed at the next op's input. TKET
+//! introduces the notion of "Resource" to extend the lifetime of a linear value
+//! over multiple ops.
+//!
+//! If a linear value appears both in an op's input and output, we say that it
```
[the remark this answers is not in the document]
lmondada: It's now longer but more precise:

> Every linear value is associated with a resource. If a linear value passed as input to an op is also returned by the op as an output, then both input and output values are associated with the same resource. We say that the op "preserves" the resource.

tket/src/resource/flow.rs
lmondada: yep!

tket/src/resource/scope.rs
lmondada: Done

tket/src/resource/types.rs
lmondada: I'll leave this to a minimum for the time being, it's always easy to add more if needed.

tket/src/resource/types.rs
lmondada: Yes pretty sure I need on one of my many local branches 😅

tket/src/resource/types.rs
lmondada: Yes that's true. Removed! I'll define `new` in terms of `Default`.

PR #1054, opened 2025-08-20
feat!: Replace Subcircuit // take 2

lmondada:
Hey Alan, I'm interested in your take on this new version of the PR. The first two PRs are the same ones from the previous PR.

The main change is that instead of relying on portgraph's `LineConvexChecker`, this is now on top of a new struct `ResourceScope` that tracks all the linear values throughout a `Circuit`. See #1052 

I think it's a nice API, and I am interested in your thoughts. I am not sure yet whether this will be merged in in its current shape, it depends what Agustin and you think of `ResourceScope`.

Note that for this to be useful, I will need to implement the Subcircuit -> SiblingSubgraph conversion. I need to implement convexity checking for that, will work on that next :)


lmondada:
Copying over a comment from the previous PR that is still relevant:

Why Rewriter should be generic on the circuit type C:

in the (near) future we will need to distinguish Rewriter that work with Circuits vs PersistentHugr (not all rewriters will work on both)
A same rewriter should hopefully work on multiple Cs, so this should be a trait generic rather than an associated type
Rewriter could be generic on the node type N: HugrNode rather than the circuit, but this would make Rewriter not [dyn-compatible](https://doc.rust-lang.org/reference/items/traits.html#dyn-compatibility) (the generic impl HugrView<Node = N> argument in get_rewrites is not allowed). Having dyn-compatibility will be useful for collections of rewriters of type Vec<Box<dyn Rewriter<C>>>

lmondada:
if https://github.com/CQCL/hugr/pull/2541 goes through I'll be able to remove the `CircuitLike` trait.

PR #1061, opened 2025-08-21
fix!: Fix rotation -> float param type conversion

aborgna-q:
Fixes inconsistent encoding/decoding parameters. Now we keep them in _half-turns_ most of the time, unless an specific operation requires them to be in radians.

- The `ParameterType` used while decoding now has variants `Rotation | FloatHalfTurns | FloatRadians`.
  Operation decoders must now specify which one they expect for their inputs.
- Decoding a pytket parameter emits float operations over `FloatHalfTurns`, to avoid unnecessary conversions from/to radians.
- The encoder always stores parameters expressions in half-turns, `qsystem` ops must then add a `/ pi` before emitting their command.

BREAKING CHANGE: (pytket decoder) Renamed `LoadedParameterType` to `ParameterType` and changed its variants.
BREAKING CHANGE: (pytket decoder) Made `LoadedParameter` attributes private, and modified its API to support the new type variants.
BREAKING CHANGE: (pytket encoder) Changed `MakeOperationArgs`'s parameter list to a `Cow`.

This is not a breaking change for python;
 
BEGIN_COMMIT_OVERRIDE

fix: Fix erroneous parameters being decoded from pytket for qsystem gates (#1061)

END_COMMIT_OVERRIDE

### DIFF
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -612,7 +612,6 @@ impl<'h> PytketDecoderContext<'h> {
     pub fn load_parameter_with_type(&mut self, param: &str, typ: ParameterType) -> LoadedParameter {
         self.wire_tracker
             .load_parameter(&mut self.builder, param, Some(typ))
-            .with_type(typ, &mut self.builder)
     }
 }
 
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -501,10 +501,22 @@ impl<'h> PytketDecoderContext<'h> {
             .map_err(|e| e.hugr_op(&op_name))?;
         debug_assert_eq!(op_input_count, input_wires.register_count());
 
+        // Convert the parameter wires to the required types.
+        // TODO: This should be done by `find_typed_wires`, but it requires changing the API to take `&mut self`.
+        let mut param_wires = Vec::with_capacity(params.len());
+        let mut params = input_wires.iter_parameters();
+        for ty in sig.input_types() {
+            let Some(expected_type) = LoadedParameterType::from_type(ty) else {
+                continue;
+            };
+            let param = *params.next().unwrap();
+            param_wires.push(param.with_type(expected_type, &mut self.builder).wire);
+        }
+
         // Add the node to the HUGR.
         let node = self
             .builder
-            .add_dataflow_op(op, input_wires.wires())
+            .add_dataflow_op(op, input_wires.value_wires().chain(param_wires.into_iter()))
             .map_err(|e| PytketDecodeError::custom(e).hugr_op(&op_name))?;
 
         // Register the output wires.
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -4,7 +4,7 @@ mod param;
 mod tracked_elem;
 mod wires;
 
-pub use param::{LoadedParameter, LoadedParameterType};
+pub use param::{LoadedParameter, ParameterType};
 pub use tracked_elem::{TrackedBit, TrackedQubit};
 pub use wires::TrackedWires;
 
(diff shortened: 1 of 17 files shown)


tket/src/serialize/pytket/decoder/param/parser.rs
```diff
@@ -37,8 +38,12 @@ pub enum PytketParam<'a> {
     /// An operation on some nested expressions.
     #[display("{}({})", op.to_string(), args.iter().map(|a| a.to_string()).join(", "))]
     Operation {
+        /// The HUGR operation used to implement this node.
         op: OpType,
+        /// Input arguments to the operation.
         args: Vec<PytketParam<'a>>,
+        /// The parameter types used for the inputs and outputs of this operation.
```
ss2165, line 45: lets hope we don't run in to any that have mixed eh...

tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -623,80 +623,98 @@ impl WireTracker {
...
+                    Some(ParameterType::FloatHalfTurns) => {
+                        let value: Value = ConstF64::new(half_turns).into();
+                        let wire = hugr.add_load_const(value);
+                        LoadedParameter::float_half_turns(wire)
+                    }
+                    _ => {
+                        let value: Value = ConstRotation::new(half_turns).unwrap().into();
+                        let wire = hugr.add_load_const(value);
+                        LoadedParameter::rotation(wire)
+                    }
```
ss2165, line 670: I still think a pytket expression tree should be loaded in to hugr in one go as just a float expression tree, and then the final root of the tree converted in to a Rotation. I.e. this second non-type hinted case need not exist? I realise that messes with the recursive loading architecture here but it is much cleaner and avoids confusing cases like this. We can leave this for a future refactor since I don't think it actually changes behaviour (other than maybe adding/removing some needless conversions back and forth)

tket/src/serialize/pytket/tests.rs
```diff
@@ -411,6 +411,7 @@ fn json_roundtrip(#[case] circ_s: &str, #[case] num_commands: usize, #[case] num
     assert_eq!(ser.commands.len(), num_commands);
 
     let circ: Circuit = ser.decode().unwrap();
+    println!("{}", circ.mermaid_string());
```
ss2165, line 414: remove print

tket/src/serialize/pytket/decoder.rs
```diff
@@ -605,7 +601,19 @@ impl<'h> PytketDecoderContext<'h> {
...
+                .load_parameter(&mut self.builder, param, None),
+        )
+    }
+
+    /// Loads the given parameter expression as a [`LoadedParameter`] in the hugr, and converts it to the requested type and unit.
+    ///
+    /// See [`PytketDecoderContext::load_parameter`] for more details.
+    pub fn load_parameter_with_type(&mut self, param: &str, typ: ParameterType) -> LoadedParameter {
+        self.wire_tracker
+            .load_parameter(&mut self.builder, param, Some(typ))
```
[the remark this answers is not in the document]
aborgna-q: As mentioned in the `load_parameter` docs, that is a type hint that lets us decide for example between loading a rotation const or a float const. The output type is not guaranteed to have that type.

I guess it'd be less confusing if we change the _hint_ for an assurance and do the `with_type` internally. I'll make the change.

tket/src/serialize/pytket/decoder/param/parser.rs
aborgna-q: The set of operations we translate is quite restricted, so hopefully this should be enough 🤞

tket/src/serialize/pytket/decoder/wires.rs
aborgna-q: I guess we could change the SympyOp to return a float / add a new op to avoid breaking `-exts`.

But input params will still be rotations, so we'd be adding unnecessary conversions.
And same for consts.

I realize adding `RotationOp::to_halfturns`/`from_halfturns` is not expensive, but neither is checking the type here 🤷

PR #1076, opened 2025-08-26
chore: Dummy change to trigger python release

aborgna-q:
So we can make a release with #1075.

BEGIN_COMMIT_OVERRIDE

feat: Capture pytket's output permutation explicitly in the hugr connectivity (#1075)

END_COMMIT_OVERRIDE


ss2165 (approved):
cursed automation


why did we let the robots win

PR #1087, opened 2025-09-01
docs: add Python API docs

CalMacCQ:


CalMacCQ:
Need access to the repo settings to configure GH pages deployment I think.

tket-py/docs/examples/1-Getting-Started.ipynb
CalMacCQ, line 1: The only modification that I made to the notebooks was to modify the headings to resolve some sphinx warnings.

CalMacCQ:
> I enabled Pages and temporarily added this branch as a valid source :) https://cqcl.github.io/tket2/

Thanks :) 

Seems like there are also some mypy failures?

CalMacCQ:
> `main` seems fine. What would be the workflow with a separate branch? Would we have to manually push there each time we want a new docs build?

I think having `main` is fine. I think @lmondada said on slack that it might be preferable to deploy from another branch?


Also thanks for the C.I. fix :)

aborgna-q:
One more thing, we re-export the `tket-exts` extension definitions in `tket.extensions`
https://github.com/CQCL/tket2/blob/9a1b555e83343c769f2800f4876026355ee81534/tket-py/tket/extensions/__init__.py

Is it possible to re-export their docs too?

lmondada:
> I think @lmondada said on slack that it might be preferable to deploy from another branch?

No no, this seems to be a misunderstanding. I was talking about a different thing :)

uv.lock
```diff
@@ -1,6 +1,10 @@
 version = 1
-revision = 3
+revision = 2
```
lmondada, line 2: Could it be that you are running an old version of `uv` and that is downgrading the file format? I am guessing here but I think I had this issue myself at some point.

Try updating your `uv`, then resetting the uv lock file to `main` and then rerun the `uv` commands... It shouldn't change 2.8k lines.

justfile
```diff
@@ -69,3 +69,17 @@ gen-extensions:
...
+build-pydocs:
+    cd tket-py/docs && uv run --group docs sphinx-build -b html . build
+
+# Serve the docs html pages locally
+serve-docs: build-pydocs
+    npm exec serve tket-py/docs/build
+
+# Clean up all files generated by the sphinx build
+cleanup:
+    rm -rf tket-py/docs/build
```
aborgna-q, line 83: The name is generic, so we should run all the clean ops
```suggestion
# Clean up all generated files
clean:
    uv clean
    cargo clean
    rm -rf tket-py/docs/build
```

justfile
```diff
@@ -69,3 +69,20 @@ gen-extensions:
...
+serve-docs: build-pydocs
+    npm exec serve tket-py/docs/build
+
+# Clean up all generated files
+clean-docs:
+    rm -rf tket-py/docs/build
+    rm -rf tket-py/docs/generated
+
+clean-env:
+    uv clean
```
CalMacCQ, line 87: These commands take a long time to run so I'd rather have a separate recipe for them given that I run `clean-docs` a lot.

PR #1122, opened 2025-09-18
feat: Decode CircBoxes as DFG nodes

aborgna-q:
Adds support for decoding circuit boxes.

Draft since it requires a fix for decoding circuits with output-only bools.


tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,79 @@
...
+                let circ_inputs: Vec<Type> = itertools::repeat_n(qb_t(), qubits.len())
+                    .chain(itertools::repeat_n(rotation_type(), params.len()))
+                    .collect_vec();
+                let circ_outputs: Vec<Type> = itertools::repeat_n(qb_t(), qubits.len())
+                    .chain(itertools::repeat_n(bool_t(), bits.len()))
+                    .collect_vec();
+                let circ_signature = Signature::new(circ_inputs, circ_outputs);
+
+                // Decode the boxed circuit into a new Hugr
+                let config = decoder.config().clone();
```
acl-cqc, line 57: This is fine, but I would consider inlining `config` or  outlining DecodeOptions, also perhaps outlining `DecodeInsertionTarget::Region` (you're calling it insertion_target after all)

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,79 @@
...
+        qubits: &[TrackedQubit],
+        bits: &[TrackedBit],
+        params: &[LoadedParameter],
+        _opgroup: Option<&str>,
+        decoder: &mut PytketDecoderContext<'h>,
+    ) -> Result<DecodeStatus, PytketDecodeError> {
+        match op.op_type {
+            PytketOptype::CircBox => {
+                let Some(OpBox::CircBox { id: _id, circuit }) = &op.op_box else {
+                    return Ok(DecodeStatus::Unsupported);
```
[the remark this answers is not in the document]
acl-cqc: You might consider `match (op.op_type, &op.op_box)` so you can combine both `DecodeStatus::Unsupported`'s

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,79 @@
...
+                    .collect_vec();
+                let circ_outputs: Vec<Type> = itertools::repeat_n(qb_t(), qubits.len())
+                    .chain(itertools::repeat_n(bool_t(), bits.len()))
+                    .collect_vec();
+                let circ_signature = Signature::new(circ_inputs, circ_outputs);
+
+                // Decode the boxed circuit into a new Hugr
+                let config = decoder.config().clone();
+                let insertion_target = decoder.builder.container_node();
+                let internal = circuit.decode_inplace(
```
acl-cqc, line 59: I'm a little puzzled here that this all appears to be saying where to decode into (target), but not *what* to decode (source). Is this implicit in `circuit` ?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -442,20 +442,17 @@ impl<'h> PytketDecoderContext<'h> {
     ///   arguments from the input wire set.
     /// - Returns an error if the parameter wires do not match the expected
     ///   types.
-    pub fn add_node_with_wires(
+    pub fn wire_up_node(
```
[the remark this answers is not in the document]
acl-cqc: I mean, previous docs for `add_node_with_wires` appear to have been copied, so same for that and the new `wire_up_node` (I think). Suggest you split the old docs into two, and then make `add_node_with_wires` reference docs for `wire_up_node` as being included in what's done by `add_`.

acl-cqc, line 445: Also, does this actually need to be `pub`? If it's only used for hugr-core stuff, that's not really an extension point...

acl-cqc (commented):
Thanks @aborgna-q. Looks good, makes the framework look lovely and simple 😃, the only thing stopping me from approving right now is not understanding what's going on with https://github.com/CQCL/tket2/issues/1124 ....and that might just be three words changed here. (Or it might be, do it here!)

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,79 @@
...
+    ) -> Result<DecodeStatus, PytketDecodeError> {
+        match op.op_type {
+            PytketOptype::CircBox => {
+                let Some(OpBox::CircBox { id: _id, circuit }) = &op.op_box else {
+                    return Ok(DecodeStatus::Unsupported);
+                };
+
+                // We have no way to distinguish between input and output bits in the circuit box, so we assume all bits are outputs here.
+                //
+                // TODO: Pass the registers both as inputs and outputs once this is implemented
```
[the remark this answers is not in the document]
aborgna-q: It's implemented here: https://github.com/CQCL/tket2/pull/1127
I didn't want to include all the changes together to simplify the reviews

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,79 @@
...
+        _opgroup: Option<&str>,
+        decoder: &mut PytketDecoderContext<'h>,
+    ) -> Result<DecodeStatus, PytketDecodeError> {
+        match op.op_type {
+            PytketOptype::CircBox => {
+                let Some(OpBox::CircBox { id: _id, circuit }) = &op.op_box else {
+                    return Ok(DecodeStatus::Unsupported);
+                };
+
+                // We have no way to distinguish between input and output bits in the circuit box, so we assume all bits are outputs here.
```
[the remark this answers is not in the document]
aborgna-q: That's always the case (angle parameters to a pytket command cannot depend on another operation, so we cannot link variables to node outputs)

tket/src/serialize/pytket/decoder.rs
aborgna-q: It's a tool offered by the decoder, useful when we want to create the node ourselves and let the decoder wire it up.

It's currently used by the core decoder, which creates a DFG node manually before calling this.
If an internal extension decoder can use it, it should also be usable by some user-implemented ones.

aborgna-q:
Addressed the comments.

> the only thing stopping me from approving right now is not understanding what's going on with https://github.com/CQCL/tket2/issues/1124 ....and that might just be three words changed here. (Or it might be, do it here!)

I'm not sure what you meant by this?

At this point the `wire_up_node` call uses a simple heuristic for matching qubit/bit ids to the input/output;

- For qubits, use the passed ids sequentially in the inputs. And re-use them in the same order on the outputs.
  - This means that an operation `Qubit -> Qubit` consumes `Qubit#1` and we label its output with `Qubit#1` again.
- For bits, we use the passed ids sequentially in the inputs, and don't reuse them on the outputs.
  - So an op `Bit, Bit -> Bit, Bit` gets assigned inputs `Bit#1, Bit#2` and outputs `Bit#3, Bit#4`.
  
This works alright for simple operations, but when encoding a circuit box `Qubit, Bit -> Qubit, Bit`, we want to keep the same label in both the input and output bits (to follow how pytket uses them). This cannot be done with that simple heuristic.

The workaround in this PR is only temporary until we remove that heuristic. That's happening in the next PR on the stack: #1127. It's a bit noisy, so I didn't want to include it here.

tket/src/serialize/pytket/decoder.rs
```diff
@@ -416,11 +416,8 @@ impl<'h> PytketDecoderContext<'h> {
             .find_typed_wires(&self.config, types, qubit_args, bit_args, params)
     }
 
-    /// Add a new node to the HUGR using the provided wire set as input and
-    /// output.
-    ///
-    /// Inserts the new node into the HUGR, connects its input ports and
-    /// registers the node's output wires in the wire tracker.
+    /// Connects the input ports of a node using a list of input qubits, bits,
+    /// and parameters. Registers the node's output wires in the wire tracker.
```
acl-cqc, line 420: like the docs :). One nit: I think these "parameters" are something special, rather than just all inputs being parameters (/arguments)....is there some more precise term like "circuit parameters" or something?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -435,34 +432,35 @@ impl<'h> PytketDecoderContext<'h> {
     /// returned if the parameter does not match the expected wire type, but the
     /// unit (radians or half-turns) cannot be checked automatically.
     ///
+    /// Use [`PytketDecoderContext::add_node_with_wires`] to insert a new node
```
acl-cqc, line 435:
```suggestion
    /// Use [`Self::add_node_with_wires`] to insert a new node
```
feel free not to!

tket/src/serialize/pytket/decoder.rs
```diff
@@ -533,10 +528,57 @@ impl<'h> PytketDecoderContext<'h> {
...
+    /// # Arguments
+    ///
+    /// - `op`: The operation to add.
+    /// - `qubits`: The list of tracked qubits to use as input.
+    /// - `bits`: The list of tracked bits to use as input.
+    /// - `params`: The list of parameters to use as input.
+    ///
+    /// # Errors
+    ///
+    /// - Returns an error if the optype signature contains any complex type.
```
acl-cqc, line 554: Should this be an error for `wire_up_node` too? I don't see anything specific to that here.

tket/src/serialize/pytket/decoder.rs
```diff
@@ -533,10 +528,57 @@ impl<'h> PytketDecoderContext<'h> {
...
+    /// - Returns an error if the input wire set cannot be mapped to the node's
```
acl-cqc, line 555: If you keep `wire_up_node` as `pub`, then maybe just link to that rather than repeat its errors here?

acl-cqc (approved):
Thanks, @aborgna-q - looks good. These all seem minor suggestions / comment fixes.

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -0,0 +1,74 @@
...
+
+                // Decode the boxed circuit into a new Hugr
+                let options = DecodeOptions::new()
+                    .with_config(decoder.config().clone())
+                    .with_signature(circ_signature);
+                let target = DecodeInsertionTarget::Region {
+                    parent: decoder.builder.container_node(),
+                };
+                let internal =
+                    circuit.decode_inplace(decoder.builder.hugr_mut(), target, options)?;
```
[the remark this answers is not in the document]
aborgna-q: Fair enough, but I don't think we'll be using a "decode a pytket circ into a DFG" primitive much. So I'll leave it here for now, we can add it to the decoder context if needed again.
I'm not doing breaking changes now, but will leave a comment about `decode_into`.

PR #1127, opened 2025-09-22
feat(pytket-decoder)!: Allow specifying qubit/bit reuse

### LINKED ISSUE #1124: (pytket-decoder)!: Allow specifying input and output bit registers separatedly
`PytketDecoderContext::add_node_with_wires` and the `::wire_up_node` added in #1122 assume that bit registers are either used as _inputs_ or _outputs_, never both.

This is annoying when writing a custom decoder were we know more about the specific operation behaviour (e.g. when decoding `CircBox`es we may want to just pass the registers along).

This is a breaking change, but it shouldn't affect much.

aborgna-q:
Expands support for defining which qubit and bit ids get used for the outputs when encoding and decoding pytket operations.

On the decoder side, splits the `qubits: &[TrackedQubit]` and `bits: &[TrackedBit]`  parameters of the wiring-up operation into `input_` and `output_` lists, so both can be set explicitly.

On the encoder site, since the behaviour is a bit more complex (the input ids are not immediately known without some extra computation) we add a `EmitCommandOptions` struct with functions that mark which ids to reuse, and methods to easily set commonly used options.

Blocked by
- #1120
- #1121 
  - (Requires Hugr `0.22.4` / `0.23` release)
- #1123
- #1122

Closes #1124 

BREAKING CHANGE: `PytketEncoderContext::emit_node` and `PytketDecoderContext::add_node_with_wires` now explicitly control the output bit/qubit ids.

### DIFF
--- a/tket/src/serialize/pytket/extension/core.rs
+++ b/tket/src/serialize/pytket/extension/core.rs
@@ -40,10 +40,8 @@ impl PytketDecoder for CoreDecoder {
         match (op.op_type, &op.op_box) {
             (PytketOptype::CircBox, Some(OpBox::CircBox { id: _id, circuit })) => {
                 // We have no way to distinguish between input and output bits in the circuit box, so we assume all bits are outputs here.
-                //
-                // TODO: Pass the registers both as inputs and outputs once this is implemented
-                // <https://github.com/CQCL/tket2/issues/1124>
                 let circ_inputs: Vec<Type> = itertools::repeat_n(qb_t(), qubits.len())
+                    .chain(itertools::repeat_n(bool_t(), bits.len()))
                     .chain(itertools::repeat_n(rotation_type(), params.len()))
                     .collect_vec();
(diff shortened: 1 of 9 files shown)


kmk142789 (approved):

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -40,10 +40,8 @@ impl PytketDecoder for CoreDecoder {
         match (op.op_type, &op.op_box) {
             (PytketOptype::CircBox, Some(OpBox::CircBox { id: _id, circuit })) => {
                 // We have no way to distinguish between input and output bits in the circuit box, so we assume all bits are outputs here.
```
acl-cqc, line 42: Appears that after this PR the correct comment would be:
```suggestion
                // We have no way to distinguish between input and output bits in the circuit box, so we assume all bits are both inputs and outputs here.
```

Right??

Issue #1130, opened 2025-09-24
guppy_to_circuit always returns num_operations = 0


mariagg-quantinuum:
Converting a Guppy function definition to a `Tk2Circuit` through `compile_function()`-> `to_str(json,f_name)` -> `from_str(json,f_name)` always leads to a num_operations() = 0, regardless of the contents of the function.
This showed up whilst testing the `guppy_to_circuit(func_def) `defined in the `__init__.py` file of the tket-py examples.

```
def guppy_to_circuit(func_def: Any) -> Tk2Circuit:
    """Convert a Guppy function definition to a `Tk2Circuit`."""

    pkg = func_def.compile_function()

    f_name = pkg.modules[0].entrypoint_op().f_name

    json = pkg.to_str(EnvelopeConfig.TEXT)
    circ = Tk2Circuit.from_str(json, f_name)

    return lower_to_pytket(circ)
```

`guppy_to_circuit()` consistently results in zero `TketOp` across various tested functions.

In this example:
```
from hugr import Hugr
from hugr.envelope import EnvelopeConfig
from tket.passes import lower_to_pytket
from tket.circuit import Tk2Circuit
from typing import Any
from guppylang import guppy
from guppylang.std.quantum import qubit, rz
from guppylang.std.angles import angle
from guppylang.std.builtins import owned

@guppy
def one_op(q0: qubit @owned, angle: angle) -> qubit:
    rz(q0, angle)
    return (q0)

def guppy_to_circuit(func_def: Any) -> Tk2Circuit:
    """Convert a Guppy function definition to a `Tk2Circuit`."""

    pkg = func_def.compile_function()

    f_name = pkg.modules[0].entrypoint_op().f_name

    json = pkg.to_str(EnvelopeConfig.TEXT)
    circ = Tk2Circuit.from_str(json, f_name)

    return lower_to_pytket(circ)

def main():
     one_op = guppy_to_circuit(one_op)
     print(one_op.num_operations())
```

it also returns 0 even though `circ' is non-empty:

[one_op.pdf](https://github.com/user-attachments/files/22515235/one_op.pdf)

PR #1131, opened 2025-09-24
docs: Tket-py notebook cleanup

mariagg-quantinuum:
Notebook 1 is now working.
Notebook 2 has been updated, but currently fails due to Issue #1130 .
`__init__` has been updated accordingly.


tket-py/examples/utils/__init__.py
```diff
@@ -18,7 +18,7 @@ def _repr_hugr(
     def _repr_tk2circ(
         circ: Tk2Circuit, include=None, exclude=None, **kwargs
     ) -> dict[str, bytes | str]:
-        h = Hugr.load_json(circ.to_hugr_json())
+        h = Hugr.from_bytes(circ.to_bytes(EnvelopeConfig.TEXT))
```
ss2165, line 21: does this not work without setting the config to TEXT?

tket-py/examples/1-Getting-Started.ipynb
ss2165, line 1: We should avoid putting requirement installations in the notebooks I think, the repo environment is a better source of truth

tket-py/examples/2-Rewriting-Circuits.ipynb
ss2165, line 1: `print(type(cnots_to_zzphase_lhs))` seems leftover?

ss2165 (commented):
Thanks, this is a lot of fixes

tket-py/examples/utils/__init__.py
mariagg-quantinuum: it does, just removed it

tket-py/examples/utils/__init__.py
```diff
@@ -27,14 +27,15 @@ def _repr_tk2circ(
...
 
-    pkg = module.compile()
+    pkg = func_def.compile_function()
 
-    json = pkg.package.to_str(EnvelopeConfig.TEXT)
-    circ = Tk2Circuit.from_str(json, func_def.name)
+    f_name = pkg.modules[0].entrypoint_op().f_name
+
+    json = pkg.to_str(EnvelopeConfig.TEXT)
+    circ = Tk2Circuit.from_str(json, f_name)
```
ss2165, line 39: I think this similarly can be 


```suggestion
    byt = pkg.to_bytes()
    circ = Tk2Circuit.from_bytes(byt, f_name)
```

tket-py/examples/2-Rewriting-Circuits.ipynb
ss2165, line 1: are the contents of `cnots_to_zzphase_lhs` supposed to be commented out?

tket-py/examples/utils/__init__.py
mariagg-quantinuum: yeah that also works, committed now

ss2165 (approved):
🚀 

I think linking to your issue at the top of the broken notebook for any casual visitors is probably a good idea

CalMacCQ:
Once the notebooks are up to date again I could potentially build them as html pages as part of the docs as I originally done in this PR #1087

tket-py/examples/utils/__init__.py
```diff
@@ -18,7 +18,7 @@ def _repr_hugr(
     def _repr_tk2circ(
         circ: Tk2Circuit, include=None, exclude=None, **kwargs
     ) -> dict[str, bytes | str]:
-        h = Hugr.load_json(circ.to_hugr_json())
+        h = Hugr.from_bytes(circ.to_bytes(EnvelopeConfig))
```
aborgna-q, line 21: `mypy` is complaining because `EnvelopeConfig` is the class, rather than a concrete configuration.
Either pass `EnvelopeConfig.BINARY` or
```suggestion
        h = Hugr.from_bytes(circ.to_bytes())
```
since it defaults to `BINARY`.

PR #1137, opened 2025-09-29
feat: Definition of extension ops for modifiers and global phases

hkengo-qtnm:
This PR defines new ExtensionOps for modifiers (control/dagger/power) and global phases in a new module `rich_circuit`.
Modifiers are defined as higher order functions that takes a circuit and returns the modified circuit.
Global phases are defined as an Extension op that takes a rotation angle and returns nothing.
The resolver (Hugr to Hugr function that eliminates those) are to be implemented in this module.

# TODOs
Three todos are left in the code:
- I haven't yet understood what `post_opdef` should do, so I left them as default. But I am not sure this is right.
- The current `contain_qubit_term` function just cares `Runtime`, `ListType` and `TupleType`, which might not be exhaustive enough. This todo comment is left as it may be implemented in the future.


tket-exts/src/tket_exts/tket/modifier.py
```diff
@@ -0,0 +1,25 @@
...
+        return load_extension("tket.modifier")
+
+    def TYPES(self) -> List[TypeDef]:
+        """Return the types defined by this extension"""
+        return []
+
+    def OPS(self) -> List[OpDef]:
+        """Return the operations defined by this extension"""
+        # todo
+        return []
```
hkengo-qtnm, line 25: Here as well.

PR #1145, opened 2025-10-01
feat: Deal with non-local edges in pytket encoder

aborgna-q:
When encoding a hugr into a pytket circuit, we use a toposort over the given region to process nodes.
This skips over non-local edges, and caused the encoding to fail due to the input wires not being registered.

This PR adds a check before processing each node that looks at any incoming non-local edge and
- If it is a module-level definition of a float/rotation parameter, loads it.
- If it is a module-level function definition/declaration, we know function-typed wires don't carry pytket values so we mark them as such.
- Anything else (e.g. a non-local edge from a parent DF region) causes the node itself to be marked as unsupported.

### DIFF
--- a/tket/src/serialize/pytket/encoder.rs
+++ b/tket/src/serialize/pytket/encoder.rs
@@ -807,6 +807,11 @@ impl<H: HugrView> PytketEncoderContext<H> {
     ) -> Result<EncodeStatus, PytketEncodeError<H::Node>> {
         let optype = circ.hugr().get_optype(node);
 
+        if self.encode_nonlocal_inputs(node, optype, circ)? == EncodeStatus::Unsupported {
+            self.unsupported.record_node(node, circ);
+            return Ok(EncodeStatus::Unsupported);
+        }
+
         // Try to encode the operation using each of the registered encoders.
         //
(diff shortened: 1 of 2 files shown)


tket/src/serialize/pytket/encoder.rs
```diff
@@ -851,6 +858,70 @@ impl<H: HugrView> PytketEncoderContext<H> {
...
+        optype: &OpType,
+        circ: &Circuit<H>,
+    ) -> Result<EncodeStatus, PytketEncodeError<H::Node>> {
+        let node_parent = circ.hugr().get_parent(node);
+
+        // Explore the dataflow value and static inputs, but not the _other inputs_.
+        let input_ports = circ
+            .hugr()
+            .node_inputs(node)
+            .take(optype.value_input_count() + optype.static_input_port().is_some() as usize);
```
lmondada, line 886: I find this quite ugly and have thought multiple times that we should have a better API for specifying the kind of ports that we want to iterate over.

Maybe `node_ports_of_kind([Kind::ValuePort, Kind::StaticPort])`? But not for this PR :)

lmondada (approved):

PR #1159, opened 2025-10-06
feat: BorrowSquashPass to elide redundant return-borrows

### LINKED ISSUE #1074: BorrowSquash part II: Reordering borrow intervals
Once pairs of matching borrows and returns have been found (forming "borrow intervals" that represent the lifetime of the borrow; see https://github.com/CQCL/tket2/pull/1072), this pass should reorder them in such a way that multiple borrows of the same resource are grouped together.

This can be achieved with a toposort of the following graph G:

- nodes of G are the nodes in the DFG
- we distinguish three types of nodes:
   - borrow node: a node that is the borrow op of one of the borrow intervals
   - return node: a node that is the return op of one of the borrow intervals
   - opaque node: any node that is neither the borrow nor the return of one of the intervals (note that borrow/returns of non-constant indices might also end in this category)
 - edges depend on the node type:
   - for each opaque node $n$: add all incoming and outgoing dataflow edges of n to G
   - for each borrow node $b$: let $r_{pre}$ and $r_{aft}$ be the nearest return nodes of a borrow interval borrowing the same index as $b$ in the past resp. future of $b$. Add an outgoing edge from $b$ to $r_{aft}$ and add further incoming (resp. outgoing) edges to all opaque node between $r_{pre}$ and $b$, (resp. between $b$ and $r_{aft}$).
   - for each return node $r$: let $b_{pre}$ and $b_{aft}$ be the nearest borrow nodes of a borrow interval borrowing the same index as $r$ in the past resp. future of $r$. Add an outgoing edge from $r$ to $b_{aft}$ and add further incoming (resp. outgoing) edges to all opaque node between $b_{pre}$ and $r$, (resp. between $r$ and $b_{aft}$).

_Claim:_ any topological ordering of this graph leaves the ordering relation unchanged on the following subsets of nodes:
- ordering between opaque nodes and any other nodes,
- ordering within borrow nodes and return nodes on the same borrow index.

(proof an exercise)

As a result, all HUGRs obtained by reordering the borrow and return nodes according to a topological ordering of $G$ are equivalent.

Finally, we would like to get the toposort that groups together borrow intervals on the same index. This is achieved by the following tie break between nodes in the sort:
> 1. pick borrow and return nodes before opaque nodes
> 2. pick borrows and returns with lower borrow indices first

There is no need to order between borrows and returns of the same index: they will always have an edge between them and thus never tie.

As a result, all borrows of the same index will be reordered to be in sequence.

**Implementation note:** instead of hand rolling out the 100th version of toposort, I was thinking of using `petgraph::Acylic` to build `G` and then add as many tie-breaking edges as I can without introducing cycles. The effect should be the same.

acl-cqc:
closes #1072, closes #1074

Bails out if any dynamic indexing


tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
+//! Reorder and squash pairs of borrow and return nodes where possible.
```
ss2165:
```suggestion
//! Reorder and squash pairs of borrow and return operations on [`BorrowArray`] where possible.
```

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+        hugr: &H,
+    ) -> Result<Option<BorrowReturnPorts>, BorrowAnalysisError>;
+
+    /// If the specified type is an array that can be borrowed from,
+    /// return `Some` of:
+    ///   * `Some(size)` if the array's size is statically known.
+    ///   * `None` if the array's size is not statically known.
+    ///
+    /// Otherwise (if the type is not such an array), return `None`.
+    fn get_array_size(&self, ty: &Type) -> Option<Option<u64>>;
```
ss2165: The nested option is a bit annoying for readability, could be worth a dedicated enum

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+                .filter(|n| OpTag::DataflowParent.is_superset(hugr.get_optype(*n).tag()))
+                .collect();
+            &temp
+        } else {
+            &self.regions
+        };
+        let mut results = Vec::new();
+        let mut seen = HashSet::new();
+        for region in regions {
+            // Start with all nodes not reachable along dataflow edges from other nodes. (INcludes Input.)
```
ss2165:
```suggestion
            // Start with all nodes not reachable along dataflow edges from other nodes. (Includes Input.)
```

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+
+        let Some(ext_op) = op.as_extension_op() else {
+            return Ok(None);
+        };
+        Ok(match BArrayUnsafeOpDef::from_extension_op(ext_op) {
+            Ok(BArrayUnsafeOpDef::borrow) => {
+                let op = hugr.get_optype(node);
+                let sig = op
+                    .dataflow_signature()
+                    .ok_or_else(|| BorrowAnalysisError::NodeNotDataflow { op: op.clone() })?;
```
ss2165: You've already matched on BArrayUnsafeOpDef::borrow so this should be an `expect` imo

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+                if sig.input_count() != 2 || sig.output_count() != 2 {
```
ss2165: likewise an expect here

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+        .find(|(p, _)| *p == w.source())
+        .unwrap()
+        .1
+}
+
+fn all_outs(h: &impl HugrView<Node = Node>, n: Node) -> impl Iterator<Item = Wire> + '_ {
+    h.out_value_types(n).map(move |(p, _)| Wire::new(n, p))
+}
+
+fn find_const<H: HugrView>(hugr: &H, n: H::Node, inp: IncomingPort) -> Option<u64> {
```
ss2165: how hard would it be to extend to indices coming from the same output port?

```python
def foo(array: BArray, i: int) -> BArray:
    x = array.borrow(i)
    array.return(i, x)
    return array
```

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+    let (mut curr_node, mut outp) = hugr.single_linked_output(n, inp).unwrap();
+
+    fn is_const_conversion_op(op: &OpType) -> bool {
+        matches!(op, OpType::LoadConstant(..))
+            || op
+                .as_extension_op()
+                .is_some_and(|op| ConvertOpDef::from_extension_op(op) == Ok(ConvertOpDef::itousize))
```
ss2165: this feels flaky - future conversion ops wouldn't be covered. Would it not be better to rely on constant folding eliminating these beforehand?

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+            1
+        );
+        // CX's should still be there (in same place), but now directly connected:
+        for cx in [cx1, cx2] {
+            assert!(h
+                .get_optype(cx)
+                .as_extension_op()
+                .is_some_and(|eop| eop.qualified_id() == "tket.quantum.CX"));
+        }
+        assert!(h.output_neighbours(cx1).all(|n| n == cx2));
```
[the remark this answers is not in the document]
acl-cqc: does assert that the output neighbours of both CX's were returns?

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+        return Ok(None);
+    };
+    if matches!(is_br.action, BRAction::Clobber(_)) {
+        // We'll process the borrow-from-out port here, but the others are potentially fresh arrays.
+        candidates.extend(all_outs(hugr, node).filter(|w| w.source() != is_br.borrow_from.out));
+    }
+    Ok(Some((node, idx, is_br.action, is_br.borrow_from)))
+}
+
+fn wire_type(h: &impl HugrView<Node = Node>, w: Wire) -> Type {
```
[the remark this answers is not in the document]
ss2165: ah woops that's annoying, still the implementation is on a hugr and saves iteration to find the port https://docs.rs/hugr-core/0.24.0/src/hugr_core/builder/build_traits.rs.html#555

tket/src/passes/squash_borrow.rs
```diff
@@ -0,0 +1,708 @@
...
+            assert_eq!(rv, inport); // Nested array being returned into outer array.
+        }
+        return Ok(None);
+    }
+
+    {
+        let out_ty = wire_type(hugr, Wire::new(node, is_br.borrow_from.out));
+        if *array_ty != out_ty {
+            let array_ty = array_ty.clone();
+            return Err(BorrowAnalysisError::InconsistentArrayType { array_ty, out_ty });
```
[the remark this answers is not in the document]
acl-cqc: Yeah, basically all the errors were about invalid hugrs, so I'm left with a non-exhaustive but empty error enum

PR #1162, opened 2025-10-06
feat: modifier resolver for various `OpTypes`

hkengo-qtnm:
This PR implements a modifier resolver for default `OpType`s including Conditional, Call, etc.

## Contents
- New module `modifier_resolver` with some tests
  - This resolver tries to eliminate `Modifier`s from Hugr by rewriting circuits to ones with modifiers applied.
  - Most of the description is written at the head of `modifier_resolver.rs`.

## TODOs
- A lot of TODOs are left. The summary is also written at the head of `modifier_resolver.rs`.

## Not included
Modifier implementation for `TketOp`s will be done in a succeeding PR.


mark-koch (commented):
Nice work, especially appreciate the detailed docs!

I had a quick look at the code and tried to reply to all the questions. The overall structure looks good to me, but I don't think I'll have time for a detailed review. Would be good to maybe have a second pair of eyes on this

PR #1180, opened 2025-10-15
chore: tket1 passes nix fixes

jake-arkinstall:
I couldn't get `cargo build --workspace` to work, and it was annoying me.

Here are the changes I needed to make. Merge at your discretion:
- Adding libclang
- Removing Nix Hardening when building inside of the devenv - see https://github.com/tikv/jemallocator/issues/108
- Adding a conan profile for gcc14, which is the one provided by devenv (on linux, at least)


ss2165:
@cqc-alec do you want to check the conan config works for you?

cqc-alec (approved):

PR #1182, opened 2025-10-16
feat(pytket-decoder): Decode unsupported subgraphs from barriers

aborgna-q:
Re-inserts the unsupported subgraphs into the new HUGR when decoding a pytket circuit.

As of this draft creation, this should work correctly when decoding single pytket circuits (no nested circuits inside unsupported regions) that only have flat unsupported regions (with no nested regions nor external function calls).

Depends on
- #1164
- #1175

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2679,11 +2679,14 @@ dependencies = [
  "bindgen",
  "cc",
  "conan2",
+ "hugr",
+ "itertools 0.14.0",
  "libc",
  "rstest 0.26.1",
  "serde",
  "serde_json",
  "thiserror 2.0.17",
+ "tket",
  "tket-json-rs",
 ]
 
--- a/tket1-passes/Cargo.toml
+++ b/tket1-passes/Cargo.toml
@@ -25,6 +25,11 @@ conan2 = "0.1.8"
 
 [dev-dependencies]
 rstest.workspace = true
+itertools.workspace = true
+
+# Used for testing
+hugr = { workspace = true }
+tket = { path = "../tket", version = "0.15.0" }
 
 [lints]
 workspace = true
--- a/tket1-passes/tests/tket1-on-hugr.rs
+++ b/tket1-passes/tests/tket1-on-hugr.rs
@@ -1,0 +1,82 @@
+//! Test running tket1 passes on hugr circuit.
+
+use tket1_passes::Tket1Circuit;
+
+use hugr::builder::{BuildError, Dataflow, DataflowHugr, FunctionBuilder};
+use hugr::extension::prelude::qb_t;
+use hugr::types::Signature;
+use hugr::{HugrView, Node};
+use rstest::{fixture, rstest};
+use tket::extension::{TKET1_EXTENSION_ID, TKET_EXTENSION_ID};
+use tket::serialize::pytket::{EncodeOptions, EncodedCircuit};
+use tket::{Circuit, TketOp};
+
+/// A flat quantum circuit inside a function.
+///
+/// This should optimize to the identity.
+#[fixture]
+fn circ_flat_quantum() -> Circuit {
+    fn build() -> Result<Circuit, BuildError> {
+        let input_t = vec![qb_t(), qb_t()];
+        let output_t = vec![qb_t(), qb_t()];
+        let mut h =
+            FunctionBuilder::new("preset_qubits", Signature::new(input_t, output_t)).unwrap();
+
+        let mut circ = h.as_circuit(h.input_wires());
+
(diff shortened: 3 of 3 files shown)


tket/src/serialize/pytket/decoder.rs
```diff
@@ -107,12 +107,16 @@ impl<'h, H: HugrView> PytketDecoderContext<'h, H> {
...
-        signature: Option<Signature>,
-        input_params: impl IntoIterator<Item = String>,
-        config: impl Into<Arc<PytketDecoderConfig<H>>>,
+        mut options: DecodeOptions<H>,
     ) -> Result<Self, PytketDecodeError> {
-        let config: Arc<PytketDecoderConfig<H>> = config.into();
-        let signature = signature.unwrap_or_else(|| {
+        // Ensure that the set of decoders is present, use a default one if not.
+        if options.config.is_none() {
+            options.with_default_config();
```
acl-cqc: This is a mutator, should be called `set` not `with` (unless it's is a builder, but I don't think it is)

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -97,3 +102,194 @@ impl<H: HugrView> PytketDecoder<H> for CoreDecoder {
...
+        }
+    })
+    .last()
+    .unwrap();
+    let module_children = opaque_hugr
+        .children(opaque_hugr.module_root())
+        .filter(|c| *c != entrypoint_function)
+        .map(|c| (c, module));
+
+    // Insert the hugr's entrypoint region directly into the region being built,
```
acl-cqc: Ok so we need a big TODO here or something, when linking comes along that'll do the lifting here, moreover this approach will duplicate the non-entrypoint functions for each subgraph (when there are multiple subgraphs), which would be invalid if any were public. (Also, hang on, if it's an external subgraph, any such functions won't be in the `opaque_hugr` anyway as that was called by `extract_subgraph`....)

It'd be great if we could hold off doing this a bit longer (we inline the qubit taking functions first anyway right) and use linking the first time...

One approach would be to record which source (opaque_hugr) functions had gone to what Node in the target Hugr, and then use `link_by_node` to use the copies already made in the new Hugr.

But this is also a place where, if we're using External subgraphs, doing this in-place on a single "host" hugr would work much better (then the things you insert are only nodes straight from pytket), and maybe it makes sense to only try to give decent results (/support multiple subcircuits, etc.) in that situation.

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -97,3 +102,194 @@ impl<H: HugrView> PytketDecoder<H> for CoreDecoder {
...
+            &mut input_qubits,
+            &mut input_bits,
+            &mut input_params,
+            Some(edge_id),
+        )?;
+
+        let wire = match found_wire {
+            FindTypedWireResult::Register(wire_data) => wire_data.wire(),
+            FindTypedWireResult::Parameter(param) => param.wire(),
+            FindTypedWireResult::Unsupported { wire, id } => {
```
acl-cqc:
```suggestion
            FindTypedWireResult::Unsupported { wire, _ } => wire
```

tket/src/serialize/pytket/extension/core.rs
acl-cqc: Ok, thinking about this a bit more.
* When you are using standalone/inline Hugrs, `insert_forest` is probably appropriate: we have to add the whole of the standalone Hugr, including any module-level functions; and there is no sharing between them. In the future we can switch `insert_forest` to `insert_link_hugr` (taking shape in https://github.com/CQCL/hugr/pull/2555) to allow sharing of public functions.
   * Alternatively we can use `link_by_node` with `NodeLinkingDirective::add()` for each toplevel, that has the same effect.
* When you are using external Hugrs, we should use `link_by_node`...
   * The first time we insert a subgraph of the original host hugr, add all the toplevel functions (as you are doing with `insert_forest`), but record the node in the target (/new/being-constructed) Hugr to which each function in the source corresponds.
   * On subsequent calls, add all the functions in the source Hugr but with `NodeLinkingDirective::UseExisting` of the node recorded the first time.

That'll keep all the right links but avoid duplicating functions (in fact, it's better than the still-to-come linking-by-name routines as it'll avoid duplicating even private functions).

However note this means that you can't have a method on UnsupportedSubgraph returning the Hugr to insert (also you can't use `extract_subgraph`). You'll have to **either** return a Hugr *and* a map of NodeLinkingDirectives, **or** have a method to perform insertion of the UnsupportedSubgraph into the target.

PR #1190, opened 2025-10-20
ci: Avoid cyclic dependency on release-checks

aborgna-q:
`qis-compiler` added a dev dependency on guppy, which depends on `tket-exts`.

This caused the release-check to fail due to conflicting package resolution
https://github.com/CQCL/tket2/actions/runs/18656728359/job/53187575800?pr=1142

I think removing the `uv lock` line here should work

[Test run](https://github.com/CQCL/tket2/actions/runs/18657807507/job/53191097888)

### DIFF
--- a/.github/workflows/release-checks.yml
+++ b/.github/workflows/release-checks.yml
@@ -53,7 +53,8 @@ jobs:
           uv sync --no-sources --no-install-workspace
           uv pip install --no-sources tket-exts
           uv pip install --no-sources tket-eccs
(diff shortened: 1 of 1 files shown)


maximilianruesch:
The release checks currently work and they still contain `uv lock`: https://github.com/Quantinuum/tket2/blob/00cf6ec025f48a3bbc9ecacb0e913fcfc8c669c0/.github/workflows/release-checks.yml#L51-L61

I deem this PR outdated. Future problems should be cause for new issues / PRs.

PR #1191, opened 2025-10-20
fix: clean-up test dependency in qis compiler release workflow

ss2165:
Release-As: 0.2.8


aborgna-q (approved):

ss2165:
> Worth ensuring tests pass locally with uvx cibuildwheel before making a new release.

yes, ideally I think we should run in CI for PRs that change `pyproject.toml`

PR #1200, opened 2025-10-23
fix: guppy_to_circuit always returns num_operations = 0

ankurmagdum:
Addresses #1130 

Fixes the bug where the `num_operations` method of `tket::circuit::Circuit` always returned `0` for circuits whose top-level `FuncDef` contained only `Input`, `Output`, and `Cfg` nodes. Additionally, the existing implementation never visited sibling `FuncDef`s.

**Root cause**
* The traversal only descended into nodes tagged as `DataflowParent`, but `Cfg` and `Conditional` nodes (which also contain `ExtensionOp` and `OpaqueOp` ops) do not have that tag.
* The traversal started from `self.parent()`, which is the entry-point of the underlying hugr, without any explicit handling of `Call` ops.

**Fix**
* Treat `Cfg` and `Conditional` nodes as traversal parents in addition to `DataflowParent` nodes.
* Start traversal from `self.hugr().module_root()` to cover all `FuncDef`s.

**Testing**
The script [test_count_ops.py](https://github.com/user-attachments/files/23099903/test_count_ops.py) extends the script in the issue by creating a variety of Guppy circuits and checks that `num_operations()` returns nonzero and scales with circuit complexity.

I’d appreciate guidance on:

* whether this traversal logic matches the intended semantics and behaviour,
* and where to best write the tests; in `tket` or `tket-py`.

PS: The error in the example `tket-py/examples/2-Rewriting-Circuits.ipynb` is resolved, however the rendered `merged_circuit` doesn't show the intended rewrite.

### DIFF
--- a/tket/src/circuit.rs
+++ b/tket/src/circuit.rs
@@ -202,9 +202,7 @@ impl<T: HugrView> Circuit<T> {
                     && !IGNORED_EXTENSION_OPS.contains(&optype.to_smolstr())
                 {
                     count += 1;
-                } else if OpTag::DataflowParent.is_superset(optype.tag())
-                    | matches!(optype, OpType::CFG(_) | OpType::Conditional(_))
-                {
+                } else if self.hugr().first_child(node).is_some() {
                     roots.push(child);
                 }
             }


ankurmagdum:
I have added the tests using `pytest.mark.parameterize`. Please take a look.

aborgna-q:
I moved your tests to pre-compiled hugrs, to avoid the guppylang dependency.

Thanks again for the PR!

PR #1205, opened 2025-10-27
feat!: Interval is independent of resource IDs and scope position

lmondada:
Previously, `Interval`s were storing `ResourceId`s and `Position`s. This made them only valid for a specific `ResourceScope`. This data is no longer stored, so that `Interval`s are now valid independently of the choice of `ResourceId`s and `Position`s. See the docstring of `Interval` and the added test.

BREAKING CHANGE: most methods of resource::Interval now take a `ResourceScope` argument; Interval::singleton constructor renamed to Interval::new_singleton


lmondada:
@mariagg-quantinuum Could you review this please? I cannot requests from you atm 🫠

mariagg-quantinuum:
Checked this PR, all seems to work and make sense to me 👍 @lmondada 
also I think you might be able to set me as a reviewer now so I can tick "review done"

Issue #801, opened 2025-03-07
`Rule` and `RuleMatcher` only exist in `tket2-py`


lmondada:
As mentioned today in the chat with @ABorgna and @willsimmons1465, currently the option to build pattern matchers and rewriters from rewrite rules only exists in the Python bindings.

It should probably be moved to the core `tket2` repo, unless someone sees a reason why they have to be within the bindings code.

Might also be of interest to @CalMacCQ

PR #1208, opened 2025-10-27
refactor: move Rule and RuleMatcher from tket-py to core tket

ankurmagdum:
Addresses #801 

I wasn't entirely sure about where exactly to include `Rule` and `RuleMatcher` in `tket`, so have currently placed them in `portmatching::matcher`. Feedback on code correctness and/or documentation would be helpful. Meanwhile, I shall continue to add tests.

### DIFF
--- a/tket-py/src/pattern.rs
+++ b/tket-py/src/pattern.rs
@@ -5,17 +5,16 @@ pub mod portmatching;
 use crate::circuit::Tk2Circuit;
 use crate::rewrite::PyCircuitRewrite;
 use crate::utils::{create_py_exception, ConvertPyErr};
+use derive_more::From;
 
-use hugr::{HugrView, Node};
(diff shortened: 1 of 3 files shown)


ankurmagdum:
I have discovered an incompatibility issue with guppy-based `Rule` and the `RuleMatcher`. 

In [test_rule_matcher.py](https://github.com/user-attachments/files/23600440/test_rule_matcher.py), two `RuleMatcher`s are created. One with `guppy` and the other with `pytket`. Only the latter is able to enforce the replacement rule.

The difference between `guppy` and `pytket` is that the circuit corresponding to the guppy-based `Rule` has the pattern contained within a `CFG` node, which seems to be preventing the matcher from recognizing the `lhs` pattern in a target circuit.

It seems like this issue should be addressed in `guppy` instead of here. Please advice.

PR #1211, opened 2025-10-31
feat!: pytket EncodedCircuit struct for in-place pytket optimisation

aborgna-q:
For previous reviews, see
- #1164 
- #1182
Comments from those have ben addressed already.

This PR adds an `EncoderCircuit` in between the encoding/decoding process `Hugr <-> tket_json_rs::SerialCircuit`.

This new instruction carries multiple `SerialCircuits` associated with the Hugr region they encode. This lets us replace the old regions with modified ones in the original hugr, after optimising the SerialCircuits.

The opaque barriers that represent sections of the original hugr not encodable as pytket commands now have two payload variants:
- `OpaqueSubgraphPayload::Inline` as before encodes the unsupported hugr envelope, plus some boundary edge ids that we can use to reconstruct links between different inline payloads.
  This variant does not support all circuits. Subgraphs must be flat and contain no non-local edges (just as before, but now we detect and error out on those cases).
- `OpaqueSubgraphPayload::External`. This references a subgraph in the original hugr with a simple ID. This allows us to re-use the nodes from the initial hugr, keeping all the hierarchy and edges.
The second case only works for `Hugr`s due to `HugrBuilder` limitations, but it's the usecase we'll use for tket1 optimisation.
The first one is kept for cases where we need to extract standalone pytket circuits.

For testing, this includes
- Several new roundtrip tests in `serialize::pytket::tests`, including error tests
- An integration test in `tket1-passes` that encodes a flat hugr, applies a pass from pytket, and reassembles it inline.

BREAKING CHANGE: `OpConvertError` renamed to `PytketEncodeOpError`
BREAKING CHANGE: Some minor changes to `PytketDecodeError` variant attributes.
BREAKING CHANGE: `fn_name` moved from `DecodeOptions` to `DecodeInsertionTarget::Function`

### DIFF
--- a/tket/src/serialize/pytket/circuit.rs
+++ b/tket/src/serialize/pytket/circuit.rs
@@ -56,7 +56,8 @@ pub(super) struct EncodedCircuitInfo {
     /// The serial circuit encoded from the region.
     pub serial_circuit: SerialCircuit,
     /// A subgraph of the region that does not contain any operation encodable
-    /// as a pytket command, and hence was not encoded in [`serial_circuit`].
+    /// as a pytket command, and has no qubit/bits in its boundary that could be
+    /// used to emit an opaque barrier command in the [`serial_circuit`].
     pub extra_subgraph: Option<SubgraphId>,
     /// List of parameters in the pytket circuit in the order they appear in the
     /// hugr input.
@@ -193,7 +194,7 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
     ///
     /// When encoding a `Hugr`, prefer using [`EncodedCircuit::new`] instead to
     /// avoid unnecessary copying of the opaque subgraphs and preserve non-local
-    /// edges and function calls.
+    /// edges (like function references).
     ///
     /// See [`EncodeOptions`] for the options used by the encoder.
     pub fn new_standalone<H: HugrView<Node = Node>>(
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -5,6 +5,7 @@ mod subgraph;
 mod tracked_elem;
 mod wires;
 
+use hugr::extension::ExtensionRegistry;
 use hugr::hugr::hugrmut::HugrMut;
 pub use param::{LoadedParameter, ParameterType};
 pub use tracked_elem::{TrackedBit, TrackedQubit};
@@ -55,16 +56,16 @@ pub struct PytketDecoderContext<'h> {
     pub builder: DFGBuilder<&'h mut Hugr>,
     /// A tracker keeping track of the generated wires and their corresponding types.
     pub(super) wire_tracker: Box<WireTracker>,
-    /// Options used when decoding the circuit.
-    ///
-    /// This contains the decoding parameters specific to the current circuit
-    /// being decoded.
-    options: DecodeOptions,
-    // A register containing custom operation decoders.
+    // A registry containing custom operation decoders.
     ///
     /// This is a copy of the configuration in `options`, if present, or
     /// [`default_decoder_config`][super::default_decoder_config] if not.
     config: Arc<PytketDecoderConfig>,
+    /// The extensions to use when loading the HUGR envelope.
+    ///
+    /// When `None`, we will use a default registry that includes the prelude,
+    /// std, TKET1, and TketOps extensions.
+    pub extensions: Option<ExtensionRegistry>,
     /// A registry of opaque subgraphs from the original Hugr, that may be referenced by opaque barriers in the pytket circuit
     /// via their [`SubgraphId`].
     pub(super) opaque_subgraphs: Option<&'h OpaqueSubgraphs<Node>>,
@@ -88,7 +89,7 @@ impl<'h> PytketDecoderContext<'h> {
     ///
     /// # Defining the function signature
     ///
-    /// If the options do not define a `signature`, we default to a sequence of
+    /// If `options` do not define a `signature`, we default to a sequence of
     /// qubit types followed by bool types, according to the qubit and bit
     /// counts in the circuit.
     ///
@@ -98,7 +99,7 @@ impl<'h> PytketDecoderContext<'h> {
     ///
     /// The signature may include bare parameter wires (e.g. `float64` or
     /// `rotation`) mixed between the value types. These will be associated with
-    /// the `input_params` names in the options if possible. Any remaining
+    /// the [`DecodeOptions::input_params`] names, if possible. Any remaining
     /// parameters will be added as additional inputs with type
     /// [`rotation_type`]. Additional parameter inputs may be added during
     /// runtime, as new free variables are found in the command arguments.
@@ -112,7 +113,6 @@ impl<'h> PytketDecoderContext<'h> {
         // Ensure that the set of decoders is present, use a default one if not.
         let config = options
             .config
-            .clone()
             .unwrap_or_else(|| Arc::new(default_decoder_config()));
 
         // Compute the signature of the decoded region, if not provided, and
@@ -162,8 +162,8 @@ impl<'h> PytketDecoderContext<'h> {
         Ok(PytketDecoderContext {
             builder: dfg,
             wire_tracker: Box::new(wire_tracker),
-            options,
             config,
+            extensions: options.extensions,
             opaque_subgraphs,
         })
     }
@@ -361,6 +361,12 @@ impl<'h> PytketDecoderContext<'h> {
                     // Disconnected port with an unsupported type. We just skip
                     // it, since it must have been disconnected in the original
                     // hugr too.
+                    debug_assert!(self
+                        .builder
+                        .hugr()
+                        .get_optype(output_node)
+                        .port_kind(port)
+                        .is_none_or(|kind| !kind.is_value()));
                     continue;
                 }
             };
@@ -795,9 +801,14 @@ impl<'h> PytketDecoderContext<'h> {
         &self.config
     }
 
-    /// Returns the options used by the decoder.
-    pub fn options(&self) -> &DecodeOptions {
-        &self.options
+    /// Returns the extensions to use when decoding HUGR envelopes.
+    ///
+    /// If the option is `None`, we will use a default registry that includes
+    /// the prelude, std, TKET1, and TketOps extensions.
+    pub fn extension_registry(&self) -> &ExtensionRegistry {
+        self.extensions
+            .as_ref()
+            .unwrap_or(&crate::extension::REGISTRY)
     }
 }
 
--- a/tket/src/serialize/pytket/decoder/subgraph.rs
+++ b/tket/src/serialize/pytket/decoder/subgraph.rs
@@ -213,7 +213,8 @@ impl<'h> PytketDecoderContext<'h> {
                     .collect_vec()
                 {
                     if tgt == old_output {
-                        // TODO: We should only disconnect the specific edge here, but HugrMut doesn't have a method for that?
+                        // We only need to disconnect the specific edge here,
+                        // but there should only be one incoming value edge.
                         self.builder.hugr_mut().disconnect(old_output, tgt_port);
                         self.builder
                             .hugr_mut()
@@ -236,7 +237,7 @@ impl<'h> PytketDecoderContext<'h> {
         bits: &[TrackedBit],
         params: &[LoadedParameter],
     ) -> Result<DecodeStatus, PytketDecodeError> {
-        let to_insert_hugr = Hugr::load_str(hugr_envelope, Some(self.options.extension_registry()))
+        let to_insert_hugr = Hugr::load_str(hugr_envelope, Some(self.extension_registry()))
             .map_err(|e| PytketDecodeErrorInner::UnsupportedSubgraphInlinePayload { source: e })?;
         let to_insert_signature = to_insert_hugr.inner_function_type().unwrap();
 
--- a/tket/src/serialize/pytket/decoder/wires.rs
+++ b/tket/src/serialize/pytket/decoder/wires.rs
@@ -354,14 +354,29 @@ pub(crate) struct WireTracker {
     output_qubit_permutation: Vec<usize>,
     /// Wires with unsupported types, created from the input node or from decoded opaque barriers.
     ///
-    /// See [`EncodedEdgeID`].
-    unsupported_wires: IndexMap<EncodedEdgeID, Wire>,
-    /// Pending edge targets for an `EncodedEdgeID` that hasn't been associated
-    /// to a [`Wire`] yet.
+    /// See [`EncodedEdgeID`], [`UnsupportedWireState`]
+    unsupported_wires: IndexMap<EncodedEdgeID, UnsupportedWireState>,
+}
+
+/// Possible states for the entries in [`WireTracker::unsupported_wires`].
+#[derive(Debug, Clone)]
+enum UnsupportedWireState {
+    /// The wire has been associated with a [`Wire`].
+    Associated(Wire),
+    /// The wire has not been associated with a [`Wire`] yet.
+    ///
+    /// We store target ports that need to be connected once the source is
+    /// added.
     ///
     /// This is used when decoding unsupported inline subgraphs out-of-order,
     /// where we may see the inputs before the outputs.
-    pending_edge_targets: IndexMap<EncodedEdgeID, Vec<(Node, IncomingPort)>>,
+    Pending(Vec<(Node, IncomingPort)>),
+}
+
+impl Default for UnsupportedWireState {
+    fn default() -> Self {
+        Self::Pending(Vec::new())
+    }
 }
 
 impl WireTracker {
@@ -380,7 +395,6 @@ impl WireTracker {
             parameter_vars: IndexSet::new(),
             output_qubit_permutation: Vec::with_capacity(qubit_count),
             unsupported_wires: IndexMap::new(),
-            pending_edge_targets: IndexMap::new(),
         }
     }
 
@@ -998,17 +1012,14 @@ impl WireTracker {
         targets: impl IntoIterator<Item = (Node, IncomingPort)>,
         hugr: &mut Hugr,
     ) {
-        match self.unsupported_wires.get(&id) {
-            Some(wire) => {
+        match self.unsupported_wires.entry(id).or_default() {
+            UnsupportedWireState::Associated(wire) => {
                 for (node, port) in targets {
                     hugr.connect(wire.node(), wire.source(), node, port);
                 }
             }
-            None => {
-                self.pending_edge_targets
-                    .entry(id)
-                    .or_default()
-                    .extend(targets);
+            UnsupportedWireState::Pending(existing_targets) => {
+                existing_targets.extend(targets);
             }
         }
     }
@@ -1024,11 +1035,20 @@ impl WireTracker {
         wire: Wire,
         hugr: &mut Hugr,
     ) {
-        self.unsupported_wires.insert(id, wire);
-
-        if let Some(targets) = self.pending_edge_targets.swap_remove(&id) {
-            for (node, port) in targets {
-                hugr.connect(wire.node(), wire.source(), node, port);
+        match self
+            .unsupported_wires
+            .insert(id, UnsupportedWireState::Associated(wire))
+        {
+            None => {}
+            Some(UnsupportedWireState::Pending(targets)) => {
+                for (node, port) in targets {
+                    hugr.connect(wire.node(), wire.source(), node, port);
+                }
+            }
+            Some(UnsupportedWireState::Associated(existing_wire)) => {
+                panic!(
+                    "Tried to associate unsupported wire {id} with {wire}, but it has already been associated with {existing_wire}"
+                );
             }
         }
     }
--- a/tket/src/serialize/pytket/extension/core.rs
+++ b/tket/src/serialize/pytket/extension/core.rs
@@ -45,10 +45,8 @@ impl PytketDecoder for CoreDecoder {
                 data: Some(payload),
                 ..
             } if opgroup == Some(OPGROUP_OPAQUE_HUGR) => {
-                let payload = OpaqueSubgraphPayload::load_str(
-                    payload,
-                    decoder.options().extension_registry(),
-                )?;
+                let payload =
+                    OpaqueSubgraphPayload::load_str(payload, decoder.extension_registry())?;
                 decoder.insert_subgraph_from_payload(qubits, bits, params, &payload)
             }
             PytketOperation {
--- a/tket/src/serialize/pytket/tests.rs
+++ b/tket/src/serialize/pytket/tests.rs
@@ -754,17 +754,13 @@ fn encoded_circuit_attributes(circ_measure_ancilla: Circuit) {
 /// Standalone circuit do not currently support unsupported subgraphs with
 /// nested structure or non-local edges.
 #[rstest]
-#[case::meas_ancilla(circ_measure_ancilla(), 1, CircuitRoundtripTestConfig::Default)]
-#[case::preset_qubits(circ_preset_qubits(), 1, CircuitRoundtripTestConfig::Default)]
-#[case::preset_parameterized(circ_parameterized(), 1, CircuitRoundtripTestConfig::Default)]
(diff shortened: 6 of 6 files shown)


tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+}
+
+/// Information stored about a pytket circuit encoded from a HUGR region.
+#[derive(Debug, Clone)]
+pub(super) struct EncodedCircuitInfo {
+    /// The serial circuit encoded from the region.
+    pub serial_circuit: SerialCircuit,
+    /// A subgraph of the region that does not contain any operation encodable
+    /// as a pytket command, and hence was not encoded in [`serial_circuit`].
+    #[expect(unused)]
```
acl-cqc: Is there a TODO associated here?

tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+    /// since parameters in pytket are unordered.
+    pub input_params: Vec<String>,
+    /// List of output parameter expressions found at the end of the encoded region.
+    //
+    // TODO: The decoder does not currently connect these.
+    pub output_params: Vec<String>,
+}
+
+impl EncodedCircuit<Node> {
+    /// Encode a HugrView into a [`EncodedCircuit`].
```
acl-cqc:
```suggestion
    /// Encode a Hugr into a [`EncodedCircuit`].
```
I mean, Node=Node and AsRef and AsMut is close enough to Hugr, right. (But I'm a little surprised by the need for AsMut - presumably some trait needs this? I mean, you can't actually mutate through the `&Circuit<H>`, can you?)

tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+impl<Node: HugrNode> EncodedCircuit<Node> {
+    /// Encode a HugrView into a [`EncodedCircuit`].
+    ///
+    /// The HUGR's entrypoint must be a dataflow region that will be encoded as
+    /// the main circuit. Additional circuits may be encoded if
+    /// [`EncodeOptions::encode_subcircuits`] is set.
+    ///
+    /// The circuit may contain opaque barriers encoding opaque subgraphs in the
+    /// original HUGR. These are encoded completely as Hugr envelopes in the
+    /// barrier operations metadata.
```
acl-cqc:
```suggestion
    /// barrier operations' metadata.
```
(disambiguating from `operation's` i.e. being clear, the metadata belongs to multiple operations)

tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+        circuit: &Circuit<H>,
+        mut options: EncodeOptions<H>,
+    ) -> Result<(), PytketEncodeError<H::Node>> {
+        // List of nodes to check for subcircuits.
+        //
+        // These may be either dataflow region parents that we can encode, or
+        // any node with children that we should traverse recursively until we
+        // find a dataflow region.
+        let mut candidate_nodes = VecDeque::from([circuit.parent()]);
+        let config = match options.config.take() {
```
acl-cqc: nit: use `.take().unwrap_or_else` ?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -34,23 +37,33 @@ use crate::extension::rotation::rotation_type;
...
-    /// Contains custom operation decoders, that define translation of legacy tket
-    /// commands into HUGR operations.
-    config: Arc<PytketDecoderConfig>,
+    pub(super) wire_tracker: Box<WireTracker>,
+    /// Options used when decoding the circuit.
+    ///
+    /// This contains the decoding parameters specific to the current circuit
+    /// being decoded.
+    options: DecodeOptions,
+    /// A registry of opaque subgraphs from `original_hugr`, that are referenced by opaque barriers in the pytket circuit
```
acl-cqc:
```suggestion
    /// A registry of opaque subgraphs from the original Hugr, that may be referenced by opaque barriers in the pytket circuit
```

tket/src/serialize/pytket/decoder.rs
```diff
@@ -332,12 +353,52 @@ impl<'h> PytketDecoderContext<'h> {
...
+        }
+    }
+
+    /// Register the set of opaque subgraphs that are present in the HUGR being decoded.
+    ///
+    /// # Arguments
+    /// - `opaque_subgraphs`: A registry of opaque subgraphs from
+    ///   `self.builder.hugr()`, that are referenced by opaque barriers in the
+    ///   pytket circuit via their [`SubgraphId`].
+    pub(super) fn register_opaque_subgraphs(
```
acl-cqc: I reckon (fairly strongly I think) that this should be an extra Option parameter to `::new` - which is only `pub(super)` not `pub` after all

tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+    ///
+    /// Returns an error if a circuit being decoded is invalid. See
+    /// [`PytketDecodeErrorInner`][super::error::PytketDecodeErrorInner] for
+    /// more details.
+    pub fn reassemble_inline(
+        &self,
+        hugr: &mut Hugr,
+        config: Option<Arc<PytketDecoderConfig>>,
+    ) -> Result<Vec<hugr::Node>, PytketDecodeError> {
+        let options = match &config {
```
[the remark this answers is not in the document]
acl-cqc: Or even have `DecoderOptions::with_config(Option<Config>)`....ok that name isn't right

tket/src/serialize/pytket/decoder.rs
```diff
@@ -314,8 +331,12 @@ impl<'h> PytketDecoderContext<'h> {
                 };
                 e.hugr_op("Output")
             })?;
+        let output_wire_count = output_wires.register_count();
         let output_wires = output_wires.wires();
 
+        // Qubits not in the output need to be freed.
+        self.add_implicit_qfree_operations(&qubits[output_wire_count.qubits..]);
```
acl-cqc: Oooh, yikes. Is this standard pytket? If the `qfree`s were in the original hugr, did we drop them rather than turn them into unsupported subgraph?

tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -561,6 +567,151 @@ impl WireTracker {
...
+            let Some(param) = split_off_first(params) else {
+                return Err(
+                    PytketDecodeErrorInner::NoMatchingParameter { ty: ty.to_string() }.wrap(),
+                );
+            };
+            return Ok(FoundWire::Parameter(*param));
+        }
+
+        // Translate the wire type to a pytket register count.
+        let reg_count = match (config.type_to_pytket(ty), unsupported_wire) {
```
acl-cqc: nit: how about
```
let Some(reg_count) = config.type_to_pytket(ty) else {
   return unsupported_wire
      .map(|id| FoundWire::Unsupported { id })
      .ok_or_else(PytketDecodeError::UnexpectedInputType {.....}.wrap())
};
```

tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -875,6 +983,25 @@ impl WireTracker {
...
+    /// are used to track connections between unsupported subgraphs and HUGR
+    /// input/output nodes that got encoded as opaque barriers in the pytket
+    /// circuit.
+    ///
+    /// See [`EncodedEdgeID`].
+    pub fn get_unsupported_wire(&self, id: EncodedEdgeID) -> Option<&Wire> {
+        self.unsupported_wires.get(&id)
+    }
+
+    /// Register a new unsupported wire.
```
acl-cqc:
```suggestion
    /// Starts tracking a new unsupported wire.
```
might say that this happens when you traverse the outport, and then `get_unsupported_wire` when you traverse an inport that needs to be connected

tket/src/serialize/pytket/opaque.rs
```diff
@@ -0,0 +1,198 @@
...
+
+/// Parse an external payload from a string payload.
+///
+/// Returns `None` if the payload is inline. We avoid fully decoding the payload
+/// in this case to avoid allocating a new String for the encoded envelope.
+///
+/// # Errors
+///
+/// Returns an error if the payload is invalid.
+fn parse_external_payload<N: HugrNode>(
```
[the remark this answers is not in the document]
acl-cqc: If you update the doc, the "We avoid fully decoding" (if the payload is inline) then seems a bit superfluous

tket/src/serialize/pytket/opaque/payload.rs
```diff
@@ -0,0 +1,206 @@
...
+            let edge_id = EncodedEdgeID::new(input_wire);
+            inputs.extend(itertools::repeat_n(edge_id, subgraph_inputs.len()));
+        }
+
+        let outputs = subgraph
+            .outgoing_ports()
+            .iter()
+            .map(|(n, p)| EncodedEdgeID::new(Wire::new(*n, *p)));
+
+        // TODO: This should include descendants of the subgraph. It doesn't.
```
acl-cqc: I think you've checked against this case above (and documented it) - if you still feel it's a TODO, file an issue in Hugr ;-)

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+                ty,
+                &mut input_qubits,
+                &mut input_bits,
+                &mut input_params,
+                None,
+            ) {
+                Ok(FoundWire::Register(wire_data)) => wire_data.wire(),
+                Ok(FoundWire::Parameter(param)) => param.wire(),
+                Ok(FoundWire::Unsupported { .. }) => {
+                    unreachable!("`unsupported_wire` not passed to `find_typed_wire`.");
```
acl-cqc: neat - and because this one is external

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+        // Register the output wires that should be connected to nodes in the new region.
+        //
+        // Re-wire wires from the subgraph to the old region's outputs.
+        let mut output_qubits = qubits;
+        let mut output_bits = bits;
+        for (ty, (src, src_port)) in signature.output().iter().zip_eq(subgraph.outgoing_ports()) {
+            let wire = Wire::new(*src, *src_port);
+            match self.config().type_to_pytket(ty) {
+                Some(counts) => {
+                    // Make sure to disconnect the old wire.
```
acl-cqc: I think we can see what you are doing ;-), but a comment like "wire from subgraph payload back into pytket circuit" could be good

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+                self.builder
+                    .hugr_mut()
+                    .connect(wire.node(), wire.source(), *tgt, *port);
+            }
+        }
+
+        // Register the output wires that should be connected to nodes in the new region.
+        //
+        // Re-wire wires from the subgraph to the old region's outputs.
+        let mut output_qubits = qubits;
```
acl-cqc: just declare the params `mut`?!

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+        let Some([to_insert_input, to_insert_output]) =
+            to_insert_hugr.get_io(to_insert_hugr.entrypoint())
+        else {
+            return Err(PytketDecodeError::custom(
+                "Opaque subgraph payload has a non-dataflow parent as entrypoint",
+            ));
+        };
+        let entrypoint_children = to_insert_hugr
+            .children(to_insert_hugr.entrypoint())
+            .filter(|c| *c != to_insert_input && *c != to_insert_output)
```
acl-cqc: nit: feels like it might be easier to get the `children` iterator, `take` the first two, then `map` the iterator

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+                            self.builder
+                                .hugr_mut()
+                                .connect(*src, *src_port, tgt, tgt_port);
+                        }
+                    }
+                }
+            }
+        }
+
+        // Mark the used qubits and bits as outdated.
```
acl-cqc: Isn't this marking the *original* `qubits`/`bits` outdated, regardless of whether either of the corresponding  `input_qubits` or `output_qubits` were used?

If this is right, I think you could do this in the caller `insert_subgraph_from_payload`, common to both cases

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+
+            for (to_insert_node, port) in targets {
+                let node = *insertion_result.node_map.get(&to_insert_node).unwrap();
+                self.builder
+                    .hugr_mut()
+                    .connect(wire.node(), wire.source(), node, port);
+            }
+        }
+
+        // Register the subgraph outputs in the wire tracker.
```
acl-cqc: Idea of breaking this up into 3 (insert_forest, connecting inputs, connecting outputs) strikes again, in particular the connecting outputs seems very similar to that for external subgraphs (only the unsupported-wire case is different really, I mean you already have `make_unexpected_node_out_error` common to both but there's a lot more)

tket/src/serialize/pytket/encoder.rs
```diff
@@ -198,22 +212,44 @@ impl<H: HugrView> PytketEncoderContext<H> {
...
 
-    /// Finish building and return the final [`SerialCircuit`],
-    /// as well as any parameter expressions at the circuit's output.
+    /// Finish building the pytket circuit
+    ///
+    /// # Returns
+    ///
+    /// * An [`EncodedCircuitInfo`] containing the final [`SerialCircuit`] and some additional metadata.
+    /// * The set of opaque subgraphs that were referenced (from/inside) pytket barriers.
+    #[allow(clippy::type_complexity)]
```
acl-cqc: I presume `expect` is not, erm, allowed here? But fwiw I think the two-element tuple is fine (I had been concerned about the three but not now)

tket/src/serialize/pytket/encoder.rs
```diff
@@ -198,22 +212,44 @@ impl<H: HugrView> PytketEncoderContext<H> {
...
         circ: &Circuit<H>,
         region: H::Node,
-    ) -> Result<(SerialCircuit, Vec<String>), PytketEncodeError<H::Node>> {
+    ) -> Result<(EncodedCircuitInfo, OpaqueSubgraphs<H::Node>), PytketEncodeError<H::Node>> {
         // Add any remaining unsupported nodes
         //
-        // TODO: Test that unsupported subgraphs that don't affect any qubit/bit registers
+        // TODO: Test that opaque subgraphs that don't affect any qubit/bit registers
         // are correctly encoded in pytket commands.
+        let mut extra_subgraph: Option<BTreeSet<H::Node>> = None;
```
acl-cqc: I think this might be simpler just to use a non-Option BTreeSet and check whether it's empty

acl-cqc (commented):
Well, massive PR...but great work @aborgna-q, looks like it was worth it! :-)
Corresponding mass of comments, but to summarize
* A few issues in subgraph.rs where I'm not sure what you're trying to do with edges
   * That `pending_encoded_edge_connections` thing needs addressing but I think is straightforward to move into WireTracker (?)
* Other than that, everything is minor, although it'd be nice to see
  * Something nicer than the mutable `Option<PytketDecoderConfig>` + `expect`
  * PytketDecoderContext::new() taking an Option<OpaqueSubgraphs> (more like the encoder!!) rather than having a separate `fn register_opaque_subgraphs`
  * OpaqueSubgraphs::new(encoder_count) instead using `Node::index()`

I'm not very familiar with `encoder.rs` so assume that's all ok 😉 😆 and haven't looked at tests yet, but definitely feels like we are nearly there.

tket/src/serialize/pytket/circuit.rs
aborgna-q: Yeah, AsMut<Hugr> is pretty much a Hugr.

The limitation are the Hugr builder traits, that require AsMut/AsRef. That's why decoding is only done on Hugrs.

tket/src/serialize/pytket/circuit.rs
```diff
@@ -0,0 +1,430 @@
...
+
+    /// Encode the circuits for the entrypoint region to the hugr, and if [`EncodeOptions::encode_subcircuits`] is set,
+    /// for the descendants of any unsupported node in the main circuit.
+    ///
+    /// Auxiliary method for [`Self::new`] and [`Self::new_standalone`].
+    ///
+    /// TODO: Add an option in [EncodeOptions] to run the subcircuit encoders in parallel.
+    fn encode_circuits<H: HugrView<Node = Node>>(
+        &mut self,
+        // This is already in [`self.hugr`], but we pass it since wrapping it
```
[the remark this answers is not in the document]
aborgna-q: Yeah, the hugr is gone from `self`

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+    ) -> Result<DecodeStatus, PytketDecodeError> {
+        fn mk_subgraph_error(id: SubgraphId, context: impl ToString) -> PytketDecodeError {
+            PytketDecodeErrorInner::InvalidExternalSubgraph {
+                id,
+                context: context.to_string(),
+            }
+            .wrap()
+        }
+
+        let Some(subgraph) = self.opaque_subgraphs.and_then(|s| s.get(id)) else {
```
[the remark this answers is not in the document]
aborgna-q: It'd be better to error out than to proceed as if nothing happened and move the nodes again.
(This is responding to user input after all, they may duplicate a barrier...)

`opaque_subgraphs` is a read-only reference though. So changing that would need some new piping.
I'll leave a TODO

tket/src/serialize/pytket/decoder/subgraph.rs
aborgna-q: Or `skip(2)` :P

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -0,0 +1,407 @@
...
+                    if wire_qubits.is_none() || wire_bits.is_none() {
+                        return Err(make_unexpected_node_out_error(
+                            self.config(),
+                            signature.output().iter(),
+                            qubits.len(),
+                            bits.len(),
+                        ));
+                    }
+                    self.wire_tracker.track_wire(
+                        wire,
```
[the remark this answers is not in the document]
aborgna-q: `track_wire`? It is used elsewhere

tket/src/serialize/pytket/circuit.rs
```diff
@@ -186,10 +189,11 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
     ///
     /// The circuit may contain opaque barriers encoding opaque subgraphs in the
     /// original HUGR. These are encoded completely as Hugr envelopes in the
-    /// barrier operations metadata.
+    /// barrier operations' metadata.
     ///
     /// When encoding a `Hugr`, prefer using [`EncodedCircuit::new`] instead to
-    /// avoid unnecessary copying of the opaque subgraphs.
+    /// avoid unnecessary copying of the opaque subgraphs and preserve non-local
+    /// edges and function calls.
```
acl-cqc: nit: thinking about this....actually function calls would be preserved (because of `insert_forest`), but the called functions would be duplicated (potentially making an invalid-for-linking Hugr)?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -293,49 +295,82 @@ impl<'h> PytketDecoderContext<'h> {
...
+                    e.hugr_op("Output")
+                })?;
+
+            let wire = match found_wire {
+                FoundWire::Register(wire) => wire.wire(),
+
+                FoundWire::Parameter(param) => param.wire(),
+                FoundWire::Unsupported { .. } => {
+                    // Disconnected port with an unsupported type. We just skip
+                    // it, since it must have been disconnected in the original
```
acl-cqc: can we `assert` that perhaps? `debug_assert` if it might occur in practice with users mutating the Hugr behind our backs and we think (but don't check in tests) that that is "ok"

tket/src/serialize/pytket/decoder.rs
```diff
@@ -76,43 +80,40 @@ impl<'h> PytketDecoderContext<'h> {
...
+        options: DecodeOptions,
+        opaque_subgraphs: Option<&'h OpaqueSubgraphs<Node>>,
     ) -> Result<Self, PytketDecodeError> {
         // Ensure that the set of decoders is present, use a default one if not.
-        if options.config.is_none() {
-            options = options.with_default_config();
-        }
+        let config = options
+            .config
+            .clone()
```
acl-cqc: or `take`!

tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -984,23 +984,53 @@ impl WireTracker {
...
+    /// [`Self::connect_unsupported_wire_target`], make the connections.
+    pub fn connect_unsupported_wire_source(
+        &mut self,
+        id: EncodedEdgeID,
+        wire: Wire,
+        hugr: &mut Hugr,
+    ) {
         self.unsupported_wires.insert(id, wire);
+
+        if let Some(targets) = self.pending_edge_targets.swap_remove(&id) {
```
acl-cqc: You'd probably get more protected against redeclaring the source of the same ID if you combined the maps (as above)

tket/src/serialize/pytket/opaque.rs
```diff
@@ -181,6 +185,8 @@ fn parse_external_payload<N: HugrNode>(
         pub id: Option<SubgraphId>,
     }
 
+    // Don't do the full deserialization of the payload to avoid allocating a new String for the
```
acl-cqc: :+1 yes this is good as an implementation comment

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -66,57 +70,77 @@ impl<'h> PytketDecoderContext<'h> {
...
                         continue;
                     };
                     if neigh != old_input {
+                        // If it was linked to some node, that wasn't the old
+                        // region's input, we can keep the link.
                         continue;
                     }
-                    Wire::new(neigh, neigh_port)
+                    // If it was linked to the old circuit's input, we need to
+                    // re-wire it to the new region's input.
```
acl-cqc: 👍 this all looks good and sensible now ;-)

tket/src/serialize/pytket/decoder/subgraph.rs
```diff
@@ -126,65 +150,80 @@ impl<'h> PytketDecoderContext<'h> {
...
+                // region's output, rewire it to the new region's output.
+                // Otherwise leave it connected.
+                for (tgt, tgt_port) in self
+                    .builder
+                    .hugr()
+                    .linked_inputs(*src, *src_port)
+                    .collect_vec()
+                {
+                    if tgt == old_output {
+                        // TODO: We should only disconnect the specific edge here, but HugrMut doesn't have a method for that?
```
acl-cqc: ah now you are disconnected `old_output` and reconnecting `new_output`, makes sense 👍 

You might write this as "We should only disconnect the specific edge here, but there should only be one incoming value edge"

or if you really care, disconnect `(*src, *src_port)` and reconnect all the other edges (so connect every one, but change the target from old_output to new_output) - not that hard, but I think fine not to.

tket/src/serialize/pytket/circuit.rs
```diff
@@ -56,7 +57,6 @@ pub(super) struct EncodedCircuitInfo {
     pub serial_circuit: SerialCircuit,
     /// A subgraph of the region that does not contain any operation encodable
     /// as a pytket command, and hence was not encoded in [`serial_circuit`].
-    #[expect(unused)]
     pub extra_subgraph: Option<SubgraphId>,
```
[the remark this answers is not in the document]
aborgna-q: This is set by `EncoderContext::finish`, when creating the `EncodedCircuitInfo`.

I extended the comment a bit.

PR #1212, opened 2025-11-01
test(qis-compiler): pre-generate test hugrs and remove guppylang dependency

### LINKED ISSUE #1210: Replace guppy dev-dependency with pre-compiled examples
`qis-compiler` contains a guppylang dev-dependency, used to define test examples.
To avoid open dependency cycles, the `guppy` version is pinned to a commit there.

While this works, it requires the CI to install LLVM and whatnot to be able to compile guppy and related dependencies from source when interacting with `qis-compiler`.
I tried to move that dependency to the workspace root in #1200, and was stopped by having to infect every CI with the new requirements.

We should instead remove the guppy dev dependency altogether and include pre-compiled example hugrs in the same way we defined the ones in `test_files/guppy_optimisation`. In particular,
- Guppy example files should be self-contained _uv scripts_ defining the guppy dependency in their header. These do not need to be updated with new guppy versions unless required.
- Add a `just` directive to re-compile the hugrs from the scripts, and ensure it's up-to-date in CI.
- Tests should include the `.hugr` envelopes directly, without requiring the guppy library to run.

ss2165:
Closes #1210

- copy the optimisation examples with a generation script that pins a guppylang version for hugr generation
- compress tests by parametrising over test file
- delete unused snaps
- snaps renamed to match new test set up, but otherwise unchanged
- remove guppylang dev dependency

### DIFF
--- a/qis-compiler/README.md
+++ b/qis-compiler/README.md
@@ -50,3 +50,21 @@ bitcode_apple_silicon = compile_to_bitcode(
     target_triple="x86_64-windows-msvc",
 )
 ```
+
+## Development
+
+### Snapshot testing
+This package uses snapshot testing for its compiler output.
+Sample programs are defined using guppy in the script `python/tests/generate_hugrs.py`.
+To add a new test case add a new function to generate the desired HUGR program and
+call it from the `if __name__ == "__main__":` block.
+
+The script has self contained dependencies with a pinned guppy version
+defined in comments at the top of the file. `uv` will create a virtual environment
+and install the dependencies when you run the script with `uv run python
+tests/generate_hugrs.py`. This can also be done using `just regenerate` in the
+`python/tests` directory.
+
+HUGR regeneration should not be done in conjunction with other changes to the compiler,
+to isolate possible changes to the snapshots. It should only be required if guppylang
+updates are required for testing the compiler.


jake-arkinstall (commented):
Can you briefly outline the approach required for making a breaking change in Hugr or guppylang?

It isn't clear to me how we would do such an update in a clean manner.

One approach is to:
- use the PR'd guppylang branch to submit a PR to tket2
- update guppylang to use that PR branch
- verify green lights on both PRs
- merge tket2 (and release?)
- update guppylang to use latest tket2
- merge guppylang.

Which could be automated with some actions hacking.

But when there are two such breaking guppylang PRs things get less clear in my mind. I am probably over complicating what needs to be done.

ss2165:
This PR doesn't really change the update workfloq, it really just simplifies the uv pattern so such workflows are easier to manage (fewer surprises in dependency resolution failure), because the guppylang dependencies don't interact with the workspace/project dependencies.

In the case where there is a change that requires compilation support here, the workflow does indeed look like what you describe. This needs a bigger solution to simplify - one possibility is an integration repository where all the integration tests are moved to.

qartik:
LGTM, but not entirely clear when we need to do `just regenerate`, might be worth documenting in README.

PR #1218, opened 2025-11-03
chore: Include tket1-passes into tket-py wheels

aborgna-q:
This PR bumps various dependencies to try and get `tket1-passes` shipped inside the `tket-py` wheels.
It does not modify the python API itself.

- Adds `tket1-passes` as a dependency of `tket-py`, so it gets built into the wheels
- Bumps the `tket-c-api` conan dependency, and updates the conan profiles appropriately.
- Updates the pytket-wheels CI. We cannot use `maturin-action` anymore since that runs in an isolated container (without the tools to build tket-c-api), so we hand-roll it instead.

Note that we don't build freethreaded wheels since pytket doesn't export those, and we have it as a dependency :(
I removed a bunch of targets that weren't supported by pytket too.

Wheels build: https://github.com/CQCL/tket2/actions/runs/19064782589

### DIFF
--- a/.github/actions/tket-c-api/action.yml
+++ b/.github/actions/tket-c-api/action.yml
@@ -26,14 +26,8 @@ runs:
         cache_packages: true
 
     - name: Set up conan remote
-      if: steps.cache.outputs.cache-hit != 'true' && runner.os != 'Windows'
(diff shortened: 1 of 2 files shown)


cqc-alec (approved):
LGTM

PR #1221, opened 2025-11-03
feat: add some tests for guppy 1q squashing

CalMacCQ:
1. Added some simple test cases for 1q squashing... See (the  `guppy_optimization/1q_squash` directory)
2. I've also updated the minimal guppy version to 0.21.6 and updated the syntax accordingly (`.compile()` -> `.compile_function()`).
3. Finally I've regenerated the hugrs for all files with the new Guppy version. Let me know if I should break up these changes.


CalMacCQ:
Had some issues generating the HUGRs in my local env. Will try again tomorrow.

PR #1223, opened 2025-11-05
feat!: Replace Subcircuit // take 3

lmondada:
Hey Alan,

As promised, here is the subcircuit PR, rewritten to use CopyableExpr. This way subcircuits can be used to represent any valid SiblingSubgraph (indeed, a superset of SiblingSubgraph, as certain non-convex subgraphs are expressible too).

The convexity check will come in a separate PR.

### DIFF
--- a/tket/src/circuit.rs
+++ b/tket/src/circuit.rs
@@ -76,9 +76,10 @@ lazy_static! {
         set
     };
 }
-/// The [IGNORED_EXTENSION_OPS] definition depends on the buggy behaviour of [`NamedOp::name`], which returns bare names instead of scoped names on some cases.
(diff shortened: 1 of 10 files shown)


lmondada:
Hi, is there a specific reason why this was closed? Or maybe this was automatic with the deletion of the `lm/interval-pr`...

In any case, @doug-q are you able to re-open this PR (I can't)? Otherwise I'll create a new one :)

cqc-alec:
> Hi, is there a specific reason why this was closed? Or maybe this was automatic with the deletion of the `lm/interval-pr`...

Probably the latter. It isn't possible to reopen this PR because the target branch has gone. I suggest creating a new one!

PR #1224, opened 2025-11-05
fix(encoded-circ): Track unsupported wires between input and output

aborgna-q:
Depends on #1211

These are not represented in the pytket circuit, and cannot be encoded into a SiblingSubgraph.

We just track them as an additional item in the `EncodedCircuit`'s `EncodedCircuitInfo`.

Note that the info is only store for circuits we encode directly, and not for nested regions inside circuit boxes.
This means that the info is lost when encoding those.
I added a commented-out test with a TODO to fix that (we'll need some extra plumbing to match circ boxes to external metadata).

drive-by: Make sure the encoder's WireTracker stores the input parameter names, and passes it along.

I'm ignoring breaking changes since they only affect unpublished code.

### DIFF
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -444,6 +444,7 @@ impl<'h> PytketDecoderContext<'h> {
     /// - `straight_through_wires`: A list of wires that directly connected the
     ///   input node to the output node in the original region, and were not
     ///   encoded in the pytket circuit or unsupported graphs.
+    ///   (They cannot be encoded in `extra_subgraph`).
     pub(super) fn run_decoder(
         &mut self,
         commands: &[circuit_json::Command],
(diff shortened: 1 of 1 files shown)


acl-cqc (commented):
At first glance this looks like it ought to work, but leaves me wondering - can we not encode the straight-through-wires as (part of) an unsupported subgraph? (Maybe in that `extra_subgraph`? Not entirely understanding the latter but it feels like these might be similar concepts?)

aborgna-q:
Ideally we'd do that, but subgraphs cannot be empty :/

tket/src/serialize/pytket/decoder.rs
```diff
@@ -428,21 +441,39 @@ impl<'h> PytketDecoderContext<'h> {
     /// - `commands`: The list of pytket commands to decode.
     /// - `extra_subgraph`: An additional subgraph of the original Hugr that was
     ///   not encoded as a pytket command, and must be decoded independently.
+    /// - `straight_through_wires`: A list of wires that directly connected the
+    ///   input node to the output node in the original region, and were not
```
acl-cqc: They never would be in an unsupported graph, but would they be in a pytket circuit? I think they could be if they were a supported type?

If that's correct, maybe delete the "or unsupported graphs" as that makes it sound like they could be. If they can never be in pytket circuit *or* unsupported graph, change to "....the original region. (These are not encoded in.....)".

If they can be in pytket circuit, or in unsupported graphs, and these are ones that happen not to be in either....then ok

PR #1225, opened 2025-11-05
feat: Add bindings for tket1-passes from python

aborgna-q:
Adds some pass definitions to `tket.passes`:

- `normalize_guppy(Tk2Circuit, ...) -> Tk2Circuit`
- `clifford_simp(Tk2Circuit, *, allow_swaps = True, traverse_subcircuits = True) -> Tk2Circuit`
- `squash_phased_rz(Tk2Circuit, *, traverse_subcircuits = True) -> Tk2Circuit`

The last two use the new bridge to the old tket C++ codebase to run optimisation on regions of the hugr.

The API here not too ergonomic, we just expose it as a MVP while working on the API refresh.

To call this from a guppy program output, we'll need to convert the hugr to the rust representation and back,
```py
from guppylang import guppy
from guppylang.std.angles import angle
from guppylang.std.builtins import result
from guppylang.std.quantum import h, rx, rz, measure, qubit
from hugr import Hugr
from tket.circuit import Tk2Circuit
from tket.passes import normalize_guppy, squash_phasedx_rz

@guppy
def main() -> None:
    q = qubit()

    rz(q, angle(0.1))
    h(q)
    rx(q, angle(-0.1))
    h(q)

    b = measure(q)
    result("b", b)

program = main.compile()
compiler_state = Tk2Circuit.from_bytes(program.to_bytes())
compiler_state = normalize_guppy(compiler_state)
compiler_state = squash_phasedx_rz(compiler_state)
hugr = Hugr.from_str(compiler_state.to_str())
print(hugr.render_dot())
```

drive-by: Add `Tk2Circuit.render_mermaid` method

BEGIN_COMMIT_OVERRIDE

feat: Add bindings for tket1-passes from python ([#1225](https://github.com/CQCL/tket2/pull/1225))

END_COMMIT_OVERRIDE

### DIFF
--- a/tket-py/src/passes.rs
+++ b/tket-py/src/passes.rs
@@ -60,29 +60,29 @@ create_py_exception!(
 ///
 /// Parameters:
 /// - simplify_cfgs: Whether to simplify CFG control flow.
-/// - untuple: Whether to remove tuple/untuple operations.
-/// - constant_fold: Whether to constant fold the program.
-/// - dead_funcs: Whether to remove dead functions.
-/// - inline: Whether to inline DFG operations.
+/// - remove_tuple_untuple: Whether to remove tuple/untuple operations.
+/// - constant_folding: Whether to constant fold the program.
+/// - remove_dead_funcs: Whether to remove dead functions.
+/// - inline_dfgs: Whether to inline DFG operations.
 #[pyfunction]
-#[pyo3(signature = (circ, *, simplify_cfgs = true, untuple = true, constant_fold = false, dead_funcs = true, inline = true))]
+#[pyo3(signature = (circ, *, simplify_cfgs = true, remove_tuple_untuple = true, constant_folding = false, remove_dead_funcs = true, inline_dfgs = true))]
 fn normalize_guppy<'py>(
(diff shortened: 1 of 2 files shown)


tket-py/src/passes/tket1.rs
```diff
@@ -0,0 +1,102 @@
...
+/// (https://arxiv.org/abs/1901.10114). Produces a circuit comprising TK1 gates
+/// and the two-qubit gate specified as the target.
+///
+/// Parameters:
+/// - allow_swaps: whether the rewriting may introduce implicit wire swaps.
+/// - traverse_subcircuits: Whether to apply the optimisation to nested
+///   subregions in the hugr too, rather than just the top-level region.
+//
+// TODO: We should also expose `target_gate` here, but the most appropriate
+// parameter type [`crate::ops::PyTketOp`] doesn't include `TK2` -.-
```
aborgna-q: We only call `clifford_simp` with `target_gate = CX` for now.

We don't have a `tket` optype with both `CX` and `TK2` variants, but if we really wanted to have this we could either accept a string or a pytket optype and convert it.

PR #1226, opened 2025-11-06
fix: Multiple fixes to the pytket encoder

aborgna-q:
This PR is a compilation of fixes to the pytket encoder/decoder. With these implemented, the example from https://github.com/CQCL/hugr/pull/2667 should work correctly.

The fixes include (one per commit):

- Marking wires that carry no value as unsupported, since they cannot be attached to a pytket operation. (They will instead be included in an opaque subgraph / barrier op).
- Don't use `SiblingSubgraph` to define the unsupported subgraphs. That structure is quite opinionated and brittle;
  - it doesn't support non-value edges in the boundaries
  - it has a complex definition for mapping between boundary ports and actual node ports (unnecessary in this case)
  - requires running an expensive convexity checker on each instantiation, even when we know the set of nodes is valid
  - and other annoyances
- Mark any node with order edges as unsupported (so we don't lose those the edges the pytket roundtrip)
- Fix issue with bits not appearing in the input/output missing from the pytket circuit's list
- Add an initial hardcoded translator between native booleans and tket booleans, with a TODO to make the system extensible.
- Avoid `QAlloc`ing qubits that don't get consumed afterwards.
- Ensure we give loaded output parameters the correct type.
- Don't try to encode wires with complex types that include parameters like `[float]` / `[float, rotation]`.
- Store input parameters to `extra_subgraph` (that cannot be encoded in the pytket circuit)

The main result is a bunch of tests now passing in `tests.rs`.

### DIFF
--- a/tket/src/serialize/pytket/tests.rs
+++ b/tket/src/serialize/pytket/tests.rs
@@ -996,9 +996,6 @@ fn fail_on_modified_hugr(circ_tk1_ops: Circuit) {
     CircuitRoundtripTestConfig::Default
 )]
 #[case::output_parameter_wire(circ_output_parameter_wire(), 1, CircuitRoundtripTestConfig::Default)]
-// TODO: fix edge case: non-local edge from an unsupported node inside a nested CircBox
-// to/from the input of the head region being encoded...
-#[should_panic(expected = "has an unconnected port")]
 #[case::non_local(circ_non_local(), 2, CircuitRoundtripTestConfig::Default)]
 
 fn encoded_circuit_roundtrip(
--- a/tket/src/serialize/pytket/circuit.rs
+++ b/tket/src/serialize/pytket/circuit.rs
@@ -8,6 +8,7 @@ use hugr::core::{HugrNode, IncomingPort, OutgoingPort};
 use hugr::hugr::hugrmut::HugrMut;
 use hugr::ops::handle::NodeHandle;
 use hugr::ops::{OpParent, OpTag, OpTrait};
+use hugr::types::EdgeKind;
 use hugr::{Hugr, HugrView, Node};
 use hugr_core::hugr::internal::HugrMutInternals;
 use itertools::Itertools;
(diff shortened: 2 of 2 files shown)


tket/src/serialize/pytket/opaque/subgraph.rs
```diff
@@ -0,0 +1,177 @@
...
+use hugr::types::Signature;
+use hugr::{Direction, Hugr, HugrView, IncomingPort, OutgoingPort};
+
+/// A subgraph of nodes in the Hugr that could not be encoded as TKET1
+/// operations.
+///
+/// of subgraphs it can represent; in particular const edges and order edge are
+/// not allowed in the boundary.
+#[derive(Debug, Clone)]
+pub struct OpaqueSubgraph<N> {
```
mark-koch: Should we consider upstreaming this? In general, `SiblingSubgraph` doesn't seem production ready 😅

tket/src/serialize/pytket/tests.rs
```diff
@@ -648,6 +648,28 @@ fn order_edge() -> Circuit {
     hugr.into()
 }
 
+// Nodes with order edges should be marked as unsupported to preserve the connection.
```
mark-koch:
```suggestion
```

mark-koch (approved):
Nice! Not an expert on your pytket conversion work, but looks good from the surface 👍

Mostly just some doc nits and questions. Also I'm wondering why this doesn't count as a breaking change due to signature changes/the switch to `OpaqueSubgraph`?

tket/src/serialize/pytket/tests.rs
```diff
@@ -930,8 +982,8 @@ fn fail_on_modified_hugr(circ_tk1_ops: Circuit) {
 #[case::output_parameter_wire(circ_output_parameter_wire(), 1, CircuitRoundtripTestConfig::Default)]
 // TODO: fix edge case: non-local edge from an unsupported node inside a nested CircBox
 // to/from the input of the head region being encoded...
-#[should_panic(expected = "Unsupported edge kind")]
-#[case::non_local(circ_non_local(), 1, CircuitRoundtripTestConfig::Default)]
+#[should_panic(expected = "has an unconnected port")]
+#[case::non_local(circ_non_local(), 2, CircuitRoundtripTestConfig::Default)]
```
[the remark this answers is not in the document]
aborgna-q: It was a problem with non-local edges from input nodes... It's fixed now :)

PR #1227, opened 2025-11-06
ci: Try to fix linking in the wheels building script

aborgna-q:
Makes sure we use host profiles for cross-compilation in conan.

I removed a bunch of wheel building targets that don't [have pytket wheels](https://pypi.org/project/pytket/#files), but left the conan profiles available for local builds.

I also temporarily disabled `aarch64-unknown-linux-*` builds. Although those are supported with pytket, CI was failing to use zig to compile them...

Passing CI: https://github.com/CQCL/tket2/actions/runs/19238776474?pr=1227

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -71,7 +71,7 @@ jobs:
         # - test-wheels: Whether to test installing the wheels.
         #    This is only done when the target platform matches the runner platform.
         #
-        # Note that we are platform support is limited by the pytket wheels availability.
+        # Note that platform support is limited by the pytket wheels' availability.
         # Since we have it as a dependency, wheels built for other platforms will be unusable.
         platform:
           # ---------------- Linux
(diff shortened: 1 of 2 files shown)


tket1-passes/conan-profiles/linux-armv7-musl-gcc14
cqc-alec: This is identical to `linux-arm7-gcc14`?

.github/workflows/python-wheels.yml
```diff
@@ -68,39 +68,40 @@ jobs:
         # - manylinux: (optional) If doing manylinux-compatible wheels,
         #    the manylinux version to use. Setting this option causes the wheels
         #    to be built with zig.
+        # - test-wheels: Whether to test installing the wheels.
+        #    This is only done when the target platform matches the runner platform.
+        #
+        # Note that we are platform support is limited by the pytket wheels availability.
```
cqc-alec:
```suggestion
        # Note that platform support is limited by the pytket wheels' availability.
```

PR #1239, opened 2025-11-11
ci: Fix python versions on release-checks

aborgna-q:
- Use python 3.10 when testing the lowest possible dependencies (fixes an error with pytket not being found for py3.14)
- Get the cached tket-c-api before calling `maturin build`

CI run: https://github.com/CQCL/tket2/actions/runs/19262256076 ✅

### DIFF
--- a/.github/workflows/release-checks.yml
+++ b/.github/workflows/release-checks.yml
@@ -16,13 +16,17 @@ env:
   RUSTC_WRAPPER: "sccache"
   # Pinned version for the uv package manager
   UV_VERSION: "0.9.7"
-  # The highest and lowest supported Python versions, used for testing
-  PYTHON_HIGHEST: "3.14"
-  PYTHON_LOWEST: "3.10"
+
+  # Path to the cached tket-c-api library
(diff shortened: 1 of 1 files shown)


cqc-alec:
Do you know why the "Release Checks" workflow was skipped? Looks like it should run for every PR to `main`?

cqc-alec:
But currently (since #1016 ) it looks like it should run on all PRs? Confused.

aborgna-q:
> But currently (since https://github.com/CQCL/tket2/pull/1016 ) it looks like it should run on all PRs? Confused.

The actual step has an `if` condition, because we cannot filter by _PR branch name_ on the triggers.
https://github.com/CQCL/tket2/blob/92cc39d04e8c34f4276d3adef2a3a58d490d5ba7/.github/workflows/release-checks.yml#L30

> Still puzzled why this isn't working with python 3.14.

The _lower_ check uses the minimum bounds for each direct dependency, so for pytket it was fetching `1.34.0`, which didn't have `abi3` or 3.14 wheels.

https://github.com/CQCL/tket2/actions/runs/19241911253/job/55006437818?pr=1238#step:6:57

cqc-alec:
Ah I see, thanks.

PR #1249, opened 2025-11-11
test: regenerate guppy_opt examples, and count gates

acl-cqc:
The first part here is identical to examples from #1221 but I also needed myself. The second part....well, revealed why we need the first (I think updating to include https://github.com/CQCL/guppylang/pull/1311 is particularly important) - DCE was taking out the gates because we were not doing anything to keep the `result` ops (!).

### DIFF
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -80,15 +80,13 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
     ("tket.quantum.QAlloc", 2), ("tket.quantum.CX", 2), ("tket.quantum.MeasureFree", 2)
 ])]
 #[cfg_attr(miri, ignore)] // Opening files is not supported in (isolated) miri
-fn optimise_guppy(
+fn optimise_guppy<'a>(
     #[case] mut hugr: Hugr,
-    #[case] before: impl IntoIterator<Item = (&'static str, usize)>,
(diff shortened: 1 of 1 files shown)


tket-qsystem/tests/guppy_opt.rs
```diff
@@ -70,10 +70,10 @@ fn count_gates(h: &impl HugrView) -> (usize, usize) {
 ///
 #[rstest]
 #[case::angles(guppy_angles(), (5, 1))]
-#[case::false_branch(guppy_false_branch(), (0, 0))]
+#[case::false_branch(guppy_false_branch(), (3, 1))]
 #[case::nested(guppy_nested(), (6, 3))]
 #[case::ranges(guppy_ranges(), (8, 4))]
-#[case::simple_cx(guppy_simple_cx(), (0, 0))]
+#[case::simple_cx(guppy_simple_cx(), (4, 2))]
```
aborgna-q: 👍

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -44,19 +45,39 @@ fn guppy_simple_cx() -> Hugr {
     load_guppy_circuit("simple_cx")
 }
 
+fn count_gates(h: &impl HugrView) -> (usize, usize) {
```
CalMacCQ: Could be nice to have a comment explaining what this is counting and why.

As we dicussed in person it could be nice to have a map with keys as OpType and number of instances of the optypes values

CalMacCQ (approved):

PR #1250, opened 2025-11-11
test: run pytket on guppy_opt tests, measure (very limited) success

acl-cqc:
Sadly none of the tests reaches the `.opt` hugr; we do not expect any with a `.flat` hugr to reach that step yet as we are not flattening.

### DIFF
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -10,7 +10,7 @@ use tket::extension::{TKET1_EXTENSION_ID, TKET_EXTENSION_ID};
 
 use hugr::algorithms::ComposablePass;
 use hugr::{Hugr, HugrView};
-use rstest::{fixture, rstest};
+use rstest::rstest;
 use tket::passes::NormalizeGuppy;
 use tket::serialize::pytket::{EncodeOptions, EncodedCircuit};
 use tket::Circuit;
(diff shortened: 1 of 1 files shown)


tket-qsystem/tests/guppy_opt.rs
```diff
@@ -63,37 +86,72 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
...
+#[should_panic] /// This does not yet pass for any case!
+#[rstest]
+#[case::angles("angles")]
+#[case::false_branch("false_branch")]
+#[case::simple_cx("simple_cx")]
+#[case::nested("nested")]
+#[case::ranges("ranges")]
+fn optimise_guppy_full(#[case] name: &str) {
+    let hugr = load_guppy_circuit(&format!("{name}/{name}")).unwrap();
+    let flat = load_guppy_circuit(&format!("{name}/{name}.flat")).unwrap_or(hugr.clone());
```
acl-cqc: This feels like "what we should be doing"?

acl-cqc: Yes, so I hadn't tried loading the flat circuits and optimising those. We still fail, so now including that as a separate step

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -11,40 +12,62 @@ use hugr::algorithms::ComposablePass;
...
+use tket::serialize::pytket::{EncodeOptions, EncodedCircuit};
+use tket::Circuit;
+
+use tket1_passes::Tket1Circuit;
 use tket_qsystem::QSystemPass;
 
 const GUPPY_EXAMPLES_DIR: &str = "../test_files/guppy_optimization";
 
-fn load_guppy_circuit(name: &str) -> Hugr {
-    let file = Path::new(GUPPY_EXAMPLES_DIR).join(format!("{name}/{name}.hugr"));
```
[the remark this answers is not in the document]
acl-cqc: Fair enough, I think the right way is with an enum, done that

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -63,34 +97,82 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
...
-#[case::nested(guppy_nested(), [
-    ("tket.quantum.CZ", 1), ("tket.quantum.H", 2), ("tket.quantum.QAlloc", 3), ("tket.quantum.MeasureFree", 3)
-])]
-#[case::ranges(guppy_ranges(), [
-    ("tket.quantum.QAlloc", 4), ("tket.quantum.MeasureFree", 4), ("tket.quantum.H", 2), ("tket.quantum.CX", 2)
-])]
-#[case::simple_cx(guppy_simple_cx(), [
-    ("tket.quantum.QAlloc", 2), ("tket.quantum.CX", 2), ("tket.quantum.MeasureFree", 2)
-])]
+#[should_panic = "xfail"]
```
acl-cqc: Note this homebrew xfail mechanism, which allows the test to fail but only if the gate counts are exactly as expected in the test header.

Could do the same to `flatten_guppy` too?

acl-cqc:
> The CI errors should be fixed after #1254. Be sure to rebase/merge main.

Sweet, thanks! 👍

PR #1251, opened 2025-11-12
fix: Don't use opgroup in pytket barrier encoding

aborgna-q:
We were writing `OPAQUE_HUGR` to the pytket barrier operation's opgroup when encoding opaque subgraphs, to be able to quickly check if an incoming barrier should be decoded and return errors on invalid payloads.

However, this was an invalid use of the `opgroup` field which caused pytket to panic when multiple operations with the same opgroup had different signatures.

This PR removes the use of that field, and adds some helpers to check the contents of a barrier payload instead.

### DIFF
--- a/tket/src/serialize/pytket/opaque.rs
+++ b/tket/src/serialize/pytket/opaque.rs
@@ -138,7 +138,7 @@ impl<N: HugrNode> OpaqueSubgraphs<N> {
             return Ok(());
         };
         if !self.contains(subgraph_id) {
-            return Err(PytketEncodeError::custom(format!("Barrier operation with External Subgraph payload points to an unknown subgraph: {subgraph_id}")));
+            return Err(PytketEncodeError::custom(format!("Barrier operation with external subgraph payload points to an unknown subgraph: {subgraph_id}")));
         }
 
         let payload = OpaqueSubgraphPayload::new_inline(&self[subgraph_id], hugr)?;


tket/src/serialize/pytket/opaque/payload.rs
```diff
@@ -77,11 +81,13 @@ pub enum OpaqueSubgraphPayload {
     /// A reference to a subgraph tracked by an `OpaqueSubgraphs` registry
     /// in an [`EncodedCircuit`][super::super::circuit::EncodedCircuit]
     /// structure.
+    #[serde(rename = "HugrExternal")]
```
acl-cqc: Does this mean we should declare this as a breaking PR?

acl-cqc (approved):
Yup, thanks @aborgna-q . Looks about as nice as I think it's possible to be given the tket1 serialization format...so pretty grotty 😆 but LGTM!

tket/src/serialize/pytket/extension/core.rs
```diff
@@ -36,19 +36,18 @@ impl PytketDecoder for CoreDecoder {
         qubits: &[TrackedQubit],
         bits: &[TrackedBit],
         params: &[LoadedParameter],
-        opgroup: Option<&str>,
+        _opgroup: Option<&str>,
```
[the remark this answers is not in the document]
aborgna-q: Same here, it's part of the `PytketDecoder` trait. We don't lose much by giving the info to the user (pytket commands are normally an `Operation` + arguments + `opgroup`)

tket/src/serialize/pytket/opaque/payload.rs
aborgna-q: These external payloads are meant to be short-lived, and are only valid when decoded by the same runtime that generated them.

You could make the case that we are breaking `Inline` payloads but I don't expect anyone to be storing pytket circuits with embedded Hugrs, with the aim to recover them atm

PR #1256, opened 2025-11-13
fix(pytket-decoder): Avoid QAllocating and immediately freeing qubits

aborgna-q:
When decoding a pytket circuit, when a qubit register is not present in the region's inputs we delay `QAlloc`ating it until it is consumed.

On the other side, at the end of the decoding we insert `QFree`s on each qubit that didn't get consumed and is not part of the region's outputs.

When a qubit falls in both categories (i.e. it didn't get used on pytket commands and didn't appear in the input/outputs) we were allocating it when requesting the wires to QFree.

This PR adds a simple check to avoid the extra operations.

### DIFF
--- a/tket/src/serialize/pytket/decoder/wires.rs
+++ b/tket/src/serialize/pytket/decoder/wires.rs
@@ -444,7 +444,7 @@ impl WireTracker {
 
     /// Returns `true` if the tracked qubit has been initialized.
     ///
-    /// Qubits registers that have been registered but not associated to any
+    /// Qubits that have been registered but not associated to any
     /// wires do not need to be consumed at the end of the decoding.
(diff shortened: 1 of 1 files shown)


tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -442,6 +442,16 @@ impl WireTracker {
         &self.bits[id.0]
     }
 
+    /// Returns `true` if the tracked qubit has been initialized.
+    ///
+    /// Qubits registers that have been registered but not associated to any
```
cqc-alec:
```suggestion
    /// Qubits that have been registered but not associated to any
```

PR #1270, opened 2025-11-18
feat: NormalizeGuppy uses inline_acyclic, fixes 2*opt tests

acl-cqc:
One reason that the guppy-optimization tests are doing so badly is we are typically only doing NormalizeGuppy stuff beneath the entrypoint, which is the main function. In most tests main just allocates some qubits, calls some “circuit” function outside the entrypoint, then measures and results. IIUC, EncodedCircuit might “inline” the call to the circuit (but even that is currently disabled, IIUC), and that won’t have been NormalizeGuppy'd...

Anyway, inlining everything into the entrypoint should bring all the code we need into range of NormalizeGuppy, and lo - `nested` and `ranges` now pass if we start from the `.flat` hugr. (Of course we don't expect normal to be flattened into `.flat` yet.)

Angles crashes with #1297.

The first commit shows the benefit of just adding NormalizeGuppyPass to `optimize_guppy_pytket` (should probably be renamed to `optimize_guppy_flat`), i.e., minimal without the inlining.


tket-qsystem/tests/guppy_opt.rs
```diff
@@ -74,30 +74,29 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
 ///
 
 #[rstest]
-#[should_panic = "xfail"]
+#[should_panic = "PytketDecodeError { inner: NoMatchingParameter"]
```
acl-cqc: Not sure what this is about!

aborgna-q (approved):
👍 

I guess an alternative would be to look at the call graph, and optimize all reachable functions? (That should probably be an option in the passes)

Try rebasing on #1295 & #1298 to see how the errors change.

PR #1273, opened 2025-11-23
feat!: Replace Subcircuit // take 4

lmondada:
Sorry for the delay! Mostly copying from #1223:

Hey Alan,

As promised, here is the subcircuit PR, rewritten to use CopyableExpr. This way subcircuits can be used to represent any valid SiblingSubgraph (indeed, a superset of SiblingSubgraph, as certain non-convex subgraphs are expressible too).

Note that I have replaced many of the "old" uses of Subcircuit with straight-forward `SiblingSubgraph`. The extra abstraction layer was pretty useless. It might sense to migrate those (or new versions of them) to the new `Subcircuit` at some point, but that is orthogonal to this PR.

The convexity check will come in a separate PR.

### DIFF
--- a/tket/src/resource.rs
+++ b/tket/src/resource.rs
@@ -129,8 +129,9 @@ pub(crate) mod tests {
     use hugr::{
         builder::{DFGBuilder, Dataflow, DataflowHugr},
         extension::prelude::qb_t,
+        ops::{OpType, Value},
         types::Signature,
-        CircuitUnit, Hugr,
+        Hugr, HugrView, IncomingPort, Wire,
     };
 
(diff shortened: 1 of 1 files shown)


acl-cqc:
FYI I've broken this up into #1288 (which I think can go straight in, I've made a couple of tiny modifications), #1289 (the bulk, needs reviewing), and #1290 (questions/concerns that much of the conversion-folding code should not be needed). Diffing the three together against this reveals that they are nearly identical, the only significant change being that I've skipped the parametrization of Rewriter since that wasn't used anywhere.

Will update more when I've looked at those more.

PR #1282, opened 2025-11-26
feat: Bump version of tket-c-api.

cqc-alec:
Pulls in a fix to the `GreedyPauliSimp` pass.


cqc-alec:
> Approving, but I am concerned that CI skipped python. Given that this ends up in python wheels, I don't think it should be skipped

Good catch. I'll raise an issue.

PR #1283, opened 2025-11-26
refactor: Remove contain_qubits, use TypeUnpacker

acl-cqc:
Use instead `TypeUnpacker`, both to avoid duplication of logic, and to benefit from TypeUnpacker's caching.

(It does seem a little heavyweight, but there.)

`rs-semver-checks` seems to think this wasn't pub so we're all good?

### DIFF
--- a/tket/src/modifier/modifier_resolver.rs
+++ b/tket/src/modifier/modifier_resolver.rs
@@ -111,13 +111,11 @@ pub mod dfg_modify;
 pub mod global_phase_modify;
 pub mod tket_op_modify;
 
-use super::qubit_types_utils::contain_qubits;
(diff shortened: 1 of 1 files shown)


ss2165:
> It does seem a little heavyweight, but there

it makes sense for its original use case, we could add a wrapper around it that discards the cache per request?

acl-cqc:
> it makes sense for its original use case, we could add a wrapper around it that discards the cache per request?

I think we are more concerned with speed than memory? In which case the cache is a good counter to the extra work in building the unneeded replacement-type-row-thing. (I had been wondering whether there'd be a good place to keep the cache and might not have gone with the refactor if there hadn't been that handy `struct ModifierResolver`!...that could be a wrong intuition though)

PR #1289, opened 2025-11-28
feat: add new Subcircuit

acl-cqc:
Part two of Luca's #1273, following #1288.

I've dropped the `as_const_value`, `as_const_f64` and `get_const_inputs` methods for now as they are only used in a test, so all that will be in the third part

### DIFF
--- a/tket/src/rewrite/strategy.rs
+++ b/tket/src/rewrite/strategy.rs
@@ -500,9 +500,7 @@ mod tests {
     fn rw_to_empty(circ: &Circuit, cx_nodes: impl IntoIterator<Item = Node>) -> CircuitRewrite {
         let circ: ResourceScope<_> = circ.into();
         let subcirc = Subcircuit::try_from_resource_nodes(cx_nodes, &circ).unwrap();
-        subcirc
-            .create_rewrite(n_cx(0), &circ)
-            .unwrap()
+        subcirc.create_rewrite(n_cx(0), &circ).unwrap()
     }
 
     /// Rewrite cx_nodes -> two_qb_repl (or 10x CX if None)


tket/src/resource/types.rs
```diff
@@ -21,7 +22,8 @@ pub struct ResourceId(usize);
 impl ResourceId {
     /// Create a new ResourceId.
     ///
-    /// This method should only be called by ResourceAllocator and tests.
+    /// ResourceIds should typically be obtained from [`ResourceAllocator`].
```
acl-cqc: I'll put this in some `#[cfg(test)]`

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    pub fn nodes<'a>(&'a self, circuit: &'a ResourceScope<impl HugrView<Node = N>>) -> IndexSet<N> {
+        let mut nodes: IndexSet<N> = self.nodes_on_resource_paths(circuit).collect();
+
+        // Add any nodes and function calls that are part of copyable expressions
+        for n in self.nodes_on_resource_paths(circuit) {
+            for p in circuit
+                .get_copyable_ports(n, Direction::Incoming)
+                .map(|p| p.as_incoming().expect("incoming port"))
+            {
+                let Some(expr) = self.get_copyable_expression(n, p, circuit) else {
```
acl-cqc:
```suggestion
                if let Some(expr) = self.get_copyable_expression(n, p, circuit).and_then(CopyableExpr::as_subgraph()) { nodes.extend(subgraph.nodes()) }
```

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+
+    fn extend_copyable_io(
+        &mut self,
+        node: N,
+        circuit: &ResourceScope<impl HugrView<Node = N>>,
+    ) -> bool {
+        let mut was_changed = false;
+
+        for dir in Direction::BOTH {
+            let copyable_ports = circuit.get_copyable_ports(node, dir);
```
acl-cqc: The `for dir in Direction::BOTH` and `match dir` together both enable you to reuse....this one function call. Not worth it - I'll change to do incoming then outgoing ports "separately"....

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+
+            match dir {
+                Direction::Incoming => {
+                    let new_inputs = copyable_ports
+                        .map(|p| vec![(node, p.as_incoming().expect("incoming port"))]);
+
+                    let len = self.input_copyable.len();
+                    extend_unique(&mut self.input_copyable, new_inputs);
+                    was_changed |= self.input_copyable.len() > len;
```
acl-cqc: and we could make extend_unique return a bool so combining these two to just `was_changed |= extend_unique(....)`

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+        return Err(InvalidSubcircuit::UnsupportedBoundary(
+            "resource outputs must precede copyable outputs".to_string(),
+        ));
+    }
+
+    Ok((resource_outputs, other_outputs))
+}
+
+fn extend_unique<E: PartialEq>(vec: &mut Vec<E>, other: impl IntoIterator<Item = E>) {
+    let other_unique = other.into_iter().filter(|x| !vec.contains(x)).collect_vec();
```
acl-cqc: Eeeek. This is O(n^2) de-duping....

So this seems to happen on every call to `try_add_node`, or `try_extend`. Adding a single node at a time - if ppl want to do that, they'll do it a bunch of times; I guess we would want a builder (that maintains a `Set` as well as `Vec`, and then discards the Set when `finish()`ing into a Subcircuit).

When adding a whole `ResourceScope` at a time, seems there'll be quite a few things to add, so a reasonable time to amortize construction of that `Set` (which is only O(n) time after all!) for the adding of the new RS and then discard the Set (equivalent to `.into_builder().extend(rs).finish()`).

Or, maybe there's an easier way - do we need the `input_copyable` and `output_coypable` to be in this order? (I suspect we probably do want that, but would e.g. a BTreeMap do?)

Do you think the builder is a good plan? I'm certainly happy to make `try_extend` use a Set internally, I could try the builder thing but it does make the API more complex so maybe will live with the inefficiency in `add_node` until it hurts. (Or just add a Set to Subcircuit?)

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    vec.extend(other_unique);
+}
+
+impl<H: HugrView> ResourceScope<H> {
+    fn is_convex(&self, _subcircuit: &Subcircuit<H::Node>) -> bool {
+        unimplemented!("is_convex is not yet implemented")
```
acl-cqc: The "yet" suggests that this is a `todo!`, as in, we intend it to be implemented in the future?

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+}
+
+fn parse_output_boundary<N: HugrNode>(
+    subgraph: &SiblingSubgraph<N>,
+    circuit: &ResourceScope<impl HugrView<Node = N>>,
+) -> Result<(OutgoingPorts<N>, OutgoingPorts<N>), InvalidSubcircuit<N>> {
+    let mut out_iter = subgraph.outgoing_ports().iter().copied().peekable();
+
+    let resource_outputs = out_iter
+        .peeking_take_while(|&(node, port)| {
```
acl-cqc: I might
1. common up the `get_circuit_unit().is_some_and(is_resource)` bit into a closure
2. use `iter().position()` on the Vec to get the first index
3. then `Vec::split_off` to break off the copyable ones, finally using the closure to check they're ok

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    /// Convert the subcircuit to a [`SiblingSubgraph`].
+    ///
+    /// You may use [`Self::validate_subgraph`] to check whether converting the
+    /// subcircuit to a [`SiblingSubgraph`] will succeed.
+    pub fn try_to_subgraph(
+        &self,
+        circuit: &ResourceScope<impl HugrView<Node = N>>,
+    ) -> Result<SiblingSubgraph<N>, InvalidSubgraph<N>> {
+        self.validate_subgraph(circuit)?;
+
```
acl-cqc: Should we have some 'TODO: convexity check' here?

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+        // Add copyable inputs/outputs to the subcircuit where required
+        was_changed |= self.extend_copyable_io(node, circuit);
+
+        Ok(was_changed)
+    }
+
+    /// Extend the subcircuit by including the intervals of `other`.
+    ///
+    /// The two subcircuits may not have any resources in common. If they do,
```
acl-cqc: returning `false` if nothing would have happened anyway seems fine, but this is a special case - how about a Result where the error is the common ResourceId?

Also, how about calling this `(try_)extend_parallel`?

tket/src/rewrite/strategy.rs
```diff
@@ -472,30 +475,15 @@ impl GammaStrategyCost<fn(&OpType) -> usize> {
...
     use super::*;
-    use hugr::hugr::views::SiblingSubgraph;
     use hugr::Node;
     use itertools::Itertools;
 
-    use crate::rewrite::trace::REWRITE_TRACING_ENABLED;
-    use crate::{circuit::Circuit, rewrite::CircuitRewrite, utils::build_simple_circuit};
-
-    /// Create a rewrite rule to replace the subcircuit with a new circuit.
-    /// TODO: this should use the new Subcircuit; TEMP TEST WORKAROUND until that arrives.
```
acl-cqc: This I added in #1288 just to get tests to pass

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+
+        if !circuit.is_convex(self) {
+            return Err(InvalidSubgraph::NotConvex);
+        }
+
+        Ok(())
+    }
+
+    /// Whether the subcircuit is a valid subcircuit.
+    pub fn validate_subcircuit(&self, circuit: &ResourceScope<impl HugrView<Node = N>>) {
```
acl-cqc: I haven't followed through the constructor yet, but do these checks relate to those there? Do they relate to the variants of  `InvalidSubcircuit`? It'd be nice to return some kind of error but would these be new variants?

(My feeling is that this is more about testing whether the `circuit` has changed in the background)

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    subgraph: &SiblingSubgraph<N>,
+    circuit: &ResourceScope<impl HugrView<Node = N>>,
+) -> Result<(OutgoingPorts<N>, OutgoingPorts<N>), InvalidSubcircuit<N>> {
+    let mut out_iter = subgraph.outgoing_ports().iter().copied().peekable();
+
+    let resource_outputs = out_iter
+        .peeking_take_while(|&(node, port)| {
+            circuit
+                .get_circuit_unit(node, port)
+                .is_some_and(|unit| unit.is_resource())
```
acl-cqc: here and elsewhere - change `circ.get_circuit_unit(n, p).is_some_and(|unit| unit.is_resource())` to `circ.get_resource_id(n,p),is_some()` ?

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    ) -> Result<Self, InvalidSubcircuit<N>> {
+        let resource_nodes = subgraph
+            .nodes()
+            .iter()
+            .filter(|&&n| !is_pure_copyable(n, circuit.hugr()))
+            .copied();
+        let mut subcircuit = Self::try_from_resource_nodes(resource_nodes, circuit)?;
+
+        // Now adjust the boundary of subcircuit to match the subgraph
+        let (resource_inputs, copyable_inputs) = parse_input_boundary(subgraph, circuit)?;
```
acl-cqc: hang on - do we expect linears to precede copyables for an arbitrary SiblingSubgraph??

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
 
 mod expression;
 pub use expression::CopyableExpr;
 
-mod interval;
+/// A subgraph within a [`ResourceScope`].
+///
+/// Just like [`SiblingSubgraph`], [`Subcircuit`] represents connected subgraphs
+/// within hugr dataflow regions. Unlike [`SiblingSubgraph`], the convexity
+/// check is not performed at construction time, but instead defered until
```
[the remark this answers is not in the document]
lmondada: yes you are right. You can indeed simplify to:

> Unlike [`SiblingSubgraph`], no convexity check is performed at construction time. [`Subcircuit`] instances may thus represent non-convex subgraphs, though many such cases are ruled out by the resource intervals-based representation (further details on the representation follow).

(Or similar)

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+/// a [`SiblingSubgraph`] needs to be constructed (see
+/// [`Subcircuit::try_to_subgraph`] and [`Subcircuit::validate_subgraph`]).
+///
+/// [`Subcircuit`] distinguishes between "pure copyable" nodes, which has
+/// exclusively copyable inputs and outputs, and "resource" nodes, which have at
+/// least one linear input or output.
+///
+/// ## Subcircuit representation: resource nodes
+///
+/// The subgraph composed of resource nodes is represented as a vector of
```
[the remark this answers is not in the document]
lmondada: yes! That's more precise.

tket/src/subcircuit.rs
lmondada: Yes -- `parse_input_boundary` will return an error if that is not the case. This is for round-tripping: any SiblingSubgraph that is constructed from a Subcircuit will be of this form. By only accepting SiblingSubgraphs that satisfy this constraint, we make sure we never shuffle the boundary and thus surprise the user.

Admittedly this behaviour must be added to the docstring!! Good spot, thank you :)

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+        extend_unique(&mut self.output_copyable, other.output_copyable);
+
+        true
+    }
+
+    /// Remove a copyable input from the subcircuit.
+    ///
+    /// This is possible when the input can be expressed as a function of other
+    /// inputs or computations within the subcircuit.
+    pub fn try_remove_copyable_input(
```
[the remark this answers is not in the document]
lmondada: Your intuition is correct. More specifically, the behaviour here is close to your first bullet point. Let me explain.

First of all, note that this only makes sense for **classical** inputs. In that case, it might be that a port is an input AND the input P is "covered" by other inputs, i.e. any path from a node outside of the subcircuit to P contains another input port (in the past of P).

This is the only case where removing the input P will be successful---and is what this method attempts to do.

Should this be public? Probably not, as I would not expect that users would have to modify the boundary with this level of detail. At the same time, it should never do something invalid, so the user shouldn't be able to "shoot themselves in the foot" if they used it.

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+        let remove_inputs = subcircuit
+            .input_copyable
+            .iter()
+            .flatten()
+            .filter(|&np| !copyable_inputs.iter().flatten().contains(np))
+            .copied()
+            .collect_vec();
+        for (node, port) in remove_inputs {
+            subcircuit
+                .try_remove_copyable_input(node, port, circuit)
```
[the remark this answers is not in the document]
lmondada: I think the comment I added at the definition site of `try_remove_copyable_input` should help. This is safe because all ports in `remove_inputs` should be covered by other inputs (added when `subcircuit.input_copyable` was extended above). These other ports are not removed (and would fail to be removed if attempted). So the order in which these "redundant" (covered) inputs are removed is not important.

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+
+    /// All nodes in the subcircuit.
+    ///
+    /// This includes both nodes on resource paths and within copyable
+    /// expressions.
+    pub fn nodes<'a>(&'a self, circuit: &'a ResourceScope<impl HugrView<Node = N>>) -> IndexSet<N> {
+        let mut nodes: IndexSet<N> = self.nodes_on_resource_paths(circuit).collect();
+
+        // Add any nodes and function calls that are part of copyable expressions
+        for n in self.nodes_on_resource_paths(circuit) {
```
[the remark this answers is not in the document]
lmondada: Yes, we could rename nodes to `resource_nodes` and then have a separate `classical_nodes` set.

The resulting code might be slightly clearer, though note that calling `nodes_on_resource_paths` is pretty cheap, so perf-wise having a single index set might be the right choice.

Your call, either is fine with me.

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    /// Get the copyable output ports of the subcircuit.
+    pub fn copyable_outputs(
+        &self,
+        _scope: &ResourceScope<impl HugrView<Node = N>>,
+    ) -> impl Iterator<Item = (N, OutgoingPort)> + '_ {
+        self.output_copyable.iter().copied()
+    }
+
+    /// Get the copyable expression for the given input port, if it is a
+    /// copyable port of the subcircuit.
```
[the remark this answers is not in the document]
lmondada: > a copyable input to a resource node

Yes, that's correct (and important to specify).

The subcircuit contains all resource nodes, as well as classical nodes that are "between inputs and outputs" (see the comment I added on the `struct` docstring). This will return the expression that is in the past of `(node, port)` but within the subcircuit, i.e. spans from the subcircuit inputs in the past of `(node, port)` up to `(node, port)`

tket/src/subcircuit.rs
```diff
@@ -1,7 +1,1382 @@
...
+    }) {
+        return Err(InvalidSubcircuit::UnsupportedBoundary(
+            "resource inputs must precede copyable inputs".to_string(),
+        ));
+    }
+
+    Ok((resource_inputs, other_inputs))
+}
+
+fn parse_output_boundary<N: HugrNode>(
```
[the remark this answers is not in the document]
lmondada: Yes, that's correct!

tket/src/subcircuit.rs
lmondada: yes, it's a bit shorter, so might be a good idea. I find the current version quite readable, but choose the one that you prefer :)

lmondada (commented):
Hey Alan!

Thank you so much for your work!! I think I should have answered all your comments and help clarify things where it was unclear.

Note that I cannot "apply" any of your change suggestions, as I only have read access to the repo. So please go ahead and apply them yourself.

Do you have a clear idea how you want to proceed? If you estimate that making the changes required to this PR is a lot of work, I'm happy to pick up a part of that. Just let me know exactly what you'd like me to fix.

Ping me here or in email whenever!

PR #1295, opened 2025-12-02
fix: Encoding of opaque subgraphs with no associated qubit/bit

### LINKED ISSUE #1294: `EncoderContext::get_wire_values` should avoid extracting unsupported subgraphs from parameter wires.

aborgna-q:
Fixes #1294.

This PR addresses a problem when emitting subgraphs that cannot be put in a pytket barrier since they were not associated with registers.

Given the following graph:
```mermaid
graph LR
    subgraph 0 ["(0) Module"]
        direction LR
        subgraph 1 ["(1) [**FuncDefn: #quot;test#quot;**]"]
            direction LR
            2["(2) Input"]
            3["(3) Output"]
            16["(16) Unsupported OP"]
            17["(17) tket.rotation.from_halfturns_unchecked"]
            16--"0:0<br>float64"-->17
            17--"0:0<br>rotation"-->3
        end
    end
```
The encoding of `from_halfturns_unchecked` would request the unsupported op to be emitted in an opaque subgraph, but we are unable to put that operation alone in a barrier.

While #1294 proposes doing a lookahead on the wires and backtracking/marking `from_halfturns_unchecked` as unsupported as well, the solution here is simpler.
We just let the `EncodedCircuit` metadata carry multiple subgraphs not linked to barriers, so we register them there.

The downside is that now the circuit above is not encodable as a standalone pytket circuit (since the metadata stays on the rust in-memory object), but we can change it in the future if needed.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,7 @@
 # for syntax see https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-syntax
 
 # base rule, tket team is codeowner for whole repo
-* @CQCL/tket2
+* @quantinuum/tket2
 
 # Add specific rules below.
 # Lower the rule, higher the precedence
--- a/.github/workflows/drop-cache.yml
+++ b/.github/workflows/drop-cache.yml
@@ -5,5 +5,5 @@ on:
       - closed
 
 jobs:
-    drop-cache:
-        uses: CQCL/hugrverse-actions/.github/workflows/drop-cache.yml@main
+  drop-cache:
+    uses: quantinuum/hugrverse-actions/.github/workflows/drop-cache.yml@main
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -13,8 +13,8 @@ on:
     types: [checks_requested]
 
 jobs:
-    check-title:
-        name: check-title
-        uses: CQCL/hugrverse-actions/.github/workflows/pr-title.yml@main
-        secrets:
-            GITHUB_PAT: ${{ secrets.HUGRBOT_PAT }}
+  check-title:
+    name: check-title
+    uses: quantinuum/hugrverse-actions/.github/workflows/pr-title.yml@main
+    secrets:
+      GITHUB_PAT: ${{ secrets.HUGRBOT_PAT }}
--- a/.github/workflows/semver-checks.yml
+++ b/.github/workflows/semver-checks.yml
@@ -45,7 +45,7 @@ jobs:
         run: |
           echo "Installing apt dependencies: llvm-14"
           sudo apt-get install -y llvm-14
-      - uses: CQCL/hugrverse-actions/rs-semver-checks@main
+      - uses: quantinuum/hugrverse-actions/rs-semver-checks@main
         env:
           TKET_C_API_PATH: ${{ env.TKET_C_API_PATH }}
         with:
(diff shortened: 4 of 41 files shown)


aborgna-q:
Github seems to be [returning `null` for `merge_commit_sha`](https://github.com/CQCL/tket2/actions/runs/19868824556/job/56940658221?pr=1296#step:5:55), which causes the rs-semver-checks action to misbehave.

This happened once before [in an incident](https://github.com/orgs/community/discussions/148372#discussioncomment-11721594), so it may be same this time?

Anyways, it's safe to ignore the bot diagnostics.

tket/src/serialize/pytket/encoder.rs
```diff
@@ -56,6 +58,11 @@ pub struct PytketEncoderContext<H: HugrView> {
     unsupported: UnsupportedTracker<H::Node>,
     /// A registry of already-encoded opaque subgraphs.
     opaque_subgraphs: OpaqueSubgraphs<H::Node>,
+    /// Subgraphs in `opaque_subgraph` that could not be emitted as an opaque
```
acl-cqc:
```suggestion
    /// Subgraphs in `opaque_subgraphs` that could not be emitted a opaque barriers
```
Unless we can make this just one - you'd still need to add the Option here tho

tket/src/serialize/pytket/encoder.rs
```diff
@@ -653,24 +641,28 @@ impl<H: HugrView> PytketEncoderContext<H> {
...
-            return Err(PytketEncodeError::UnsupportedSubgraphHasNoRegisters {
+            // We cannot associate this subgraph to any qubit or bit register in the pytket circuit,
+            // so we'll store it in the [`AdditionalSubgraph`]s instead when finishing the encoding.
+            self.non_emitted_subgraphs.push(AdditionalSubgraph {
+                id: subgraph_id,
                 params: input_param_exprs.clone(),
             });
-        }
+        } else {
+            // If there's registers to attach the barrier command to, emit it as a regular command.
```
acl-cqc:
```suggestion
            // If there are registers to which to attach, emit it as a barrier command.
```
("regular barrier command" if you like)

tket/src/serialize/pytket/encoder.rs
```diff
@@ -646,31 +634,35 @@ impl<H: HugrView> PytketEncoderContext<H> {
...
+            // so we'll store it in the [`AdditionalSubgraph`]s instead when finishing the encoding.
+            self.non_emitted_subgraphs.push(AdditionalSubgraph {
+                id: subgraph_id,
                 params: input_param_exprs.clone(),
             });
-        }
+        } else {
+            // If there are registers to which to attach, emit it as a barrier command.
+
+            // Create pytket operation, and add the subcircuit as hugr
```
acl-cqc: I think `payload` is a reference to an external subgraphid here, not a `hugr`, right?

tket/src/serialize/pytket/tests.rs
```diff
@@ -752,6 +752,57 @@ fn circ_complex_param_type() -> Circuit {
...
+    let input_t = vec![qb_t()];
+    let output_t = vec![qb_t(), rotation_type()];
+    let mut h = FunctionBuilder::new(
+        "unsupported_subgraph_no_registers",
+        Signature::new(input_t, output_t),
+    )
+    .unwrap();
+    let [q] = h.input_wires_arr();
+
+    // Declare two function to calls.
```
```suggestion
    // Declare two functions to call.
```

acl-cqc (approved):
Looks good, thanks @aborgna-q, a few one-liner/nits is all I can see :)

PR #1304, opened 2025-12-08
fix: Track output qubits in CircuitInfo

### LINKED ISSUE #1303: Incorrect order of outputs when decoding circuits with partial output signature
Test case:

```rust
fn circ_discard_first_qubit() -> Circuit {
    let input_t = vec![qb_t(), qb_t()];
    let output_t = vec![qb_t()];
    let mut h =
        FunctionBuilder::new("discard_first_qubit", Signature::new(input_t, output_t)).unwrap();

    let [q1, q2] = h.input_wires_arr();

    h.add_dataflow_op(TketOp::MeasureFree, [q1]).unwrap();

    let [q2] = h.add_dataflow_op(TketOp::X, [q2]).unwrap().outputs_arr();

    let hugr = h.finish_hugr_with_outputs([q2]).unwrap();
    hugr.into()
}
```

After running `serialize::pytket::tests::encoded_circuit_roundtrip` on it, the final result is
```mermaid
graph LR
    subgraph 0 ["(0) Module"]
        direction LR
        subgraph 1 ["(1) [**FuncDefn: #quot;discard_first_qubit#quot;**]"]
            direction LR
            7["(7) Input"]
            8["(8) Output"]
            9["(9) tket.quantum.X"]
            10["(10) tket.quantum.Measure"]
            11["(11) tket.quantum.QFree"]
            7--"0:0<br>qubit"-->10
            7--"1:0<br>qubit"-->9
            9--"0:0<br>qubit"-->11
            10--"0:0<br>qubit"-->8
        end
    end
```

Notice that we're connecting `q1` to the output rather than `q2`. We need to track the ignored qubits on the pytket circuit somehow, (or again just put them on the CircuitInfo)

aborgna-q:
Fixes #1303 
Depends on #1298

Stores the output list of qubit/bit parameters when encoding a region's output, so we can use them when decoding circuits.

### DIFF
--- a/tket/src/serialize/pytket/decoder.rs
+++ b/tket/src/serialize/pytket/decoder.rs
@@ -338,6 +338,7 @@ impl<'h> PytketDecoderContext<'h> {
         let qubits: Vec<TrackedQubit> = Vec::from_iter(qubits);
         let bits: Vec<TrackedBit> = Vec::from_iter(bits);
         let mut qubits_slice: &[TrackedQubit] = &qubits;
+        let mut bits_slice: &[TrackedBit] = &bits;
 
         // Load the output parameter expressions.
         let output_params = encoded_info
@@ -375,7 +376,7 @@ impl<'h> PytketDecoderContext<'h> {
                     &mut self.builder,
                     ty,
                     &mut qubits_slice,
-                    &mut bits.as_slice(),
+                    &mut bits_slice,
                     &mut params,
                     Some(EncodedEdgeID::default()),
                 )


tket/src/serialize/pytket/decoder.rs
```diff
@@ -347,8 +375,8 @@ impl<'h> PytketDecoderContext<'h> {
                     &self.config,
                     &mut self.builder,
                     ty,
-                    &mut qubits,
-                    &mut bits,
+                    &mut qubits_slice,
+                    &mut bits_slice,
```
acl-cqc: `bits_slice` is only used here, can you avoid a bunch of type annotations by just  passing `&mut &bits` ?

PR #1305, opened 2025-12-09
test(guppy_opt.rs): `run_pytket` applies to entire hugr instead of just the entrypoint

### LINKED ISSUE #1639: [Bug]: PassScope has not improved the guppy_optimization tests
### Description

We know from https://github.com/Quantinuum/tket2/pull/1270 that applying existing optimizations across the hugr, rather than just to the `main` function (which contains no quantum gates), should fix several of the not-optimizing "xfail"s in the guppy_optimization tests.

PassScope was intended to apply those optimizations without inlining, but we haven't seen improvement on those tests, so....what went wrong?

Most likely we (/some pass(es)) are either not setting PassScope as it should, and/or not taking it into account.

### Steps to Reproduce

https://github.com/Quantinuum/tket2/blob/1bfb01c56ae5bcd803ec021fa75076016ac8b886/tket-qsystem/tests/guppy_opt.rs#L118 has several `xfail`s that PassScope was intended to fix

### Expected Behaviour

These should start optimizing, as they did with https://github.com/Quantinuum/tket2/pull/1270

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

acl-cqc:
closes #1639

`guppy_opt.rs` uses its own `run_pytket` function that builds an EncodedCircuit. EncodedCircuit looks only within the entrypoint, so....instead run it everywhere.

This fixes two of the tests, so remove xfails :-)

(So, this is a bit of a test artifact. Does this mean we should update any other tests anywhere? IIUC for python stuff the answer is no.)

Also simplify the ranges.flat.array test (standard treatment now fine).

A couple of the others improve, to the point they have no "true quantum" gates only qalloc/measure/qfree. This is just that the tket1 pass in CLIFFORD_SIMP_STR does not optimize out these totally-predictable measurements....

And there's another change to behaviour on `angles.rs`, for which I've raised #1747

### DIFF
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -94,7 +94,7 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
 
 #[rstest]
 #[case::nested_array("nested_array", None)]
-#[should_panic = "xfail"]
+#[should_panic = "PytketDecodeError { inner: DuplicatedParameter"]
 #[case::angles("angles", Some(vec![
     ("tket.quantum.Rz", 2), ("tket.quantum.Measure", 1), ("tket.quantum.H", 2), ("tket.quantum.QAlloc", 1), ("tket.quantum.QFree", 1)
 ]))]
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -50,13 +50,12 @@ fn load_guppy_example(path: &str) -> std::io::Result<Hugr> {
 
 /// Run a json-encoded pytket pass on the given HUGR.
 fn run_pytket(h: &mut Hugr, pass_json: &str) {
-    let ep = h.entrypoint();
-    h.set_entrypoint(h.module_root());
     let encode_options = EncodeOptions::new()
         .with_subcircuits(true)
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
Nice!

> A couple of the others improve, to the point they have no "true quantum" gates only qalloc/measure/qfree. I think this is just that the tket1 pass in CLIFFORD_SIMP_STR does not optimize out these totally-predictable measurements....is there one that does?

We do not translate the qallocs for tket1 anymore, so we'll need to have a separate pure-hugr alloc/free optimization.

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -49,6 +50,8 @@ fn load_guppy_example(path: &str) -> std::io::Result<Hugr> {
 
 /// Run a json-encoded pytket pass on the given HUGR.
 fn run_pytket(h: &mut Hugr, pass_json: &str) {
+    let ep = h.entrypoint();
+    h.set_entrypoint(h.module_root());
```
aborgna-q: This could use [`EncodedCircuit::new_with_entrypoint`](https://docs.rs/tket/latest/tket/serialize/pytket/struct.EncodedCircuit.html#method.new_with_entrypoint) instead.

PR #1314, opened 2025-12-11
chore: Bump MSRV to 1.89 and migrate to edition 2024

aborgna-q:
Looks like we hadn't done the edition migration here.
This PR does the MSRV bump + edition migration, including running `cargo fix --edition`, `cargo fmt` (quite noisy after edition changes), and enabling+fixing some new clippy lints.

Hugr has also bumped its MSRV, so we need to do it here before updating the dependency version.

I'll add the commit to `.git-blame-ignore-revs` after merging, so this doesn't pollute blame logs.

### DIFF
--- a/tket/src/ops.rs
+++ b/tket/src/ops.rs
@@ -41,53 +41,188 @@ use strum::{EnumIter, EnumString, IntoStaticStr};
 #[non_exhaustive]
 /// Simple enum of tket quantum operations.
 pub enum TketOp {
-    /// Hadamard gate.
+    /// Hadamard gate
+    ///
+    /// This is a single qubit gate.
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -31,6 +31,7 @@ unexpected_cfgs = { level = "warn", check-cfg = [
 missing_docs = "warn"
 
 [workspace.lints.clippy]
+allow_attributes = "warn"
```
aborgna-q: Clippy now suggests using `expect` instead of `allow`, so we detect when lint exceptions are no longer needed.

ss2165 (approved):

PR #1325, opened 2025-12-16
feat!: Bump hugr to 0.25.0

aborgna-q:
BEGIN_COMMIT_OVERRIDE
feat!: Bump hugr to 0.25.0 (#1325)

BREAKING CHANGE: Bumped hugr dependency to 0.25.0
BREAKING CHANGE: Bumped pyo3 dependency to 0.27.2
END_COMMIT_OVERRIDE

~(Currently pinned to a commit in hugr main, will mark as ready once hugr is released)~


qis-compiler/rust/lib.rs
```diff
@@ -139,9 +148,9 @@ fn get_hugr_llvm_module<'c, 'hugr, 'a: 'c>(
 
 fn process_hugr(hugr: &mut Hugr) -> Result<()> {
     QSystemPass::default().run(hugr)?;
-    // with_entrypoint(hugr, hugr.module_root(), |hugr| {
-    //     // `with_entrypoint` returns Rerooted, which the pass expects a bare Hugr.
-    // })?;
+    // TODO: We still want to run this as long as deserialized hugrs are allowed to contain Value::Function
```
acl-cqc: Note this is also done inside `QSystemPass`. Admittedly, only with feature gate "llvm", but there was a TODO about removing the gate. Do we still need it here *as well*?

tket-qsystem/src/lib.rs
```diff
@@ -122,12 +124,12 @@ impl QSystemPass {
         // We expect any Hugr will have *either* drop ops, or ValueArrays (without drops),
         // so only one of these passes will do anything; the order is thus immaterial.
         self.lower_drops().run(hugr)?;
-        self.linearize_arrays().run(hugr)?;
 
         #[cfg(feature = "llvm")]
         {
-            // TODO: Remove "llvm" feature gate once `inline_constant_functions` is moved to
-            //  `hugr-passes`. See https://github.com/quantinuum/hugr/issues/2419
+            // TODO: We still want to run this as long as deserialized hugrs are allowed to contain Value::Function
```
acl-cqc: I dunno why we only wanted to remove the "llvm" feature gate after moving inline_constant_functions to hugr-passes, can we just do that now?

tket-qsystem/src/replace_bools.rs
```diff
@@ -266,10 +258,25 @@ fn barray_get_dest(size: u64, elem_ty: Type) -> NodeTemplate {
...
+        .add_hugr(in_range.hugr_mut(), in_range_container)
+        .unwrap();
+    in_range
+        .hugr_mut()
+        .connect(elem.node(), elem.source(), copy_discard, 0);
+    let [elem1, elem2] = in_range
+        .hugr_mut()
+        .node_outputs(copy_discard)
+        .map(|p| Wire::new(copy_discard, p))
+        .take(2)
```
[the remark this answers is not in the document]
acl-cqc: But if you used NodeTemplate::add above then I think you could keep in the builder framework and just use `outputs_arr`, as previously

tket-qsystem/src/replace_bools.rs
```diff
@@ -366,57 +373,58 @@ fn lowerer() -> ReplaceTypes {
...
             .get_op(GenericArrayOpDef::<BorrowArray>::get.opdef_id().as_str())
             .unwrap(),
-        |args| {
+        |args, rt| {
             let [Term::BoundedNat(size), Term::Runtime(elem_ty)] = args else {
                 unreachable!()
             };
             if elem_ty.copyable() {
-                return None;
+                return Ok(None);
```
acl-cqc: (I mean, this looks like an `Ok((!elem_ty.copyable()).then(|| barray_get_dest(....)))` too, but/so nvm)

acl-cqc (commented):
Looks good but we need also https://github.com/Quantinuum/hugr/pull/2779 first and we should also take #1333 to avoid code size regression

tket/src/passes/unpack_container/op_function_map.rs
```diff
@@ -120,7 +120,9 @@ impl OpFunctionMap {
             let func_node = hugr
                 .insert_hugr(hugr.module_root(), func_def)
                 .inserted_entrypoint;
-            lowerer.replace_op(&op, NodeTemplate::Call(func_node, vec![]));
+            // TODO: Call is deprecated. We should use LinkedHugr instead.
+            #[expect(deprecated)]
+            lowerer.set_replace_op(&op, NodeTemplate::Call(func_node, vec![]));
         }
```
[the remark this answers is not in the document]
aborgna-q: I was able to remove [the other deprecated call](https://github.com/Quantinuum/tket2/pull/1325/changes/5479c7058116104b65e12fbec9f122ffe5451dad), but doing the same thing here fails due to untranslated temp ops.
I guess it's because the hugrs inside the templates are not being translated in time when multiple lowerers are defined together...

https://github.com/Quantinuum/tket2/commit/8fa7eb69d6e3793baefb0c20b8c00c3ed6360acd

aborgna-q:
I'm having problems tracking down some [non-determinism](https://github.com/Quantinuum/tket2/actions/runs/20580498674/job/59106814434?pr=1325) in the llvm snapshot tests in qis-compiler.

Running `cargo clean && rm -rf .venv && uv run maturin d
evelop --uv && just update-snapshots-py`
fails in [HEAD](https://github.com/Quantinuum/tket2/pull/1325/commits/a2f45b9fd4810aaef4b4135b37938a06dfa8d28e) changes the snapshots on each run.

In [`0c437bfe`](https://github.com/Quantinuum/tket2/pull/1325/changes/0c437bfefeaab4c58704463c9fb1de8533106111), the test never changes.

But all the commits in the middle seem to _sometimes_ be deterministic and _sometimes_ fail depending on what was last built in my environment (even though I'm doing a clean) 🫠

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -202,6 +200,29 @@ fn build_func(op: TketOp) -> Result<Hugr, LowerTk2Error> {
...
+fn func_as_node_template(mut func_def: Hugr) -> NodeTemplate {
+    // Create a replacement hugr for the op nodes: Add a `call` node in the `func_def` hugr and set it as entrypoint.
+    let func_node = func_def.entrypoint();
+    let func_signature = func_def.inner_function_type().unwrap().into_owned();
+    let func_id = FuncID::<true>::from(func_node);
+    func_def.set_entrypoint(func_def.module_root());
+    let mut b = ModuleBuilder::with_hugr(func_def);
+    let call = {
+        let mut f = b
+            .define_function_vis("", func_signature, Visibility::Private)
```
acl-cqc: fwiw Private is default if you do just `define_function`, but fine for explicitness

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -202,6 +200,29 @@ fn build_func(op: TketOp) -> Result<Hugr, LowerTk2Error> {
...
+            .unwrap();
+        let call = f.call(&func_id, &[], f.input_wires()).unwrap();
+        f.finish_with_outputs(call.outputs()).unwrap();
+        call
+    };
+    let mut call_hugr = b.finish_hugr().unwrap();
+    call_hugr.set_entrypoint(call.node());
+
+    NodeTemplate::LinkedHugr(Box::new(call_hugr), NameLinkingPolicy::default())
```
acl-cqc: `NameLinkingPolicy::default` will mean that if you insert the call-hugr more than once, you'll get an error that it's not sure which (existing or new) FuncDefn to use (it only wants to keep one and doesn't realize they are the same)...options are
* Use `default().on_multiple_defn(OnMultiDefn::UseTarget or UseSource)`
* Add a FuncDecl only, and then link in the FuncDefn once only either before or after ReplaceTypes

tket-qsystem/src/lib.rs
acl-cqc: OIC, current implementation is in `hugr::llvm`, right. But no LLVM dependency there, so we could move it to hugr-passes. But we're just gonna delete it instead, right....

Ok, I guess there is nothing we can really do here, we have to leave it inside the `#[cfg(feature = "llvm")]`

acl-cqc (approved):
Ok so I am nervous of merging this without the nondeterminism bug solved; at the very least, we need to backup the branch so it isn't deleted by squash-merge, noting the "last deterministic commit".

I presume you want to merge before #1333 and then that can follow afterwards, and I guess that's ok as it's only a code-size regression.

But I'd also be happy to leave this as approved+mergable, but not merged, until such time as we desperately need it, and see whether that happens or we manage to fix the nondeterminism bug first....

acl-cqc (approved):
Ok, leaving both `NodeTemplate::Call`s in, and with the tests re-enabled, looks good - please go ahead ;-)

PR #1326, opened 2025-12-17
feat: Remove order edges in NormalizeGuppy pass

aborgna-q:
Depends on the unreleased `hugr 0.25.0`

Requires #1325

### DIFF
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -82,7 +82,7 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
 #[case::nested_array("nested_array", None)]
 #[should_panic = "xfail"]
 #[case::angles("angles", Some(vec![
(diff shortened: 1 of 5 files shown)


tket-qsystem/tests/guppy_opt.rs
```diff
@@ -98,7 +98,7 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
 ]))]
 #[should_panic = "xfail"]
 #[case::false_branch("false_branch", Some(vec![
-    ("TKET1.tk1op", 1), ("tket.quantum.H", 1), ("tket.quantum.QAlloc", 1), ("tket.quantum.MeasureFree", 1)
+    ("tket.quantum.Measure", 1), ("tket.quantum.QFree", 1), ("tket.quantum.QAlloc", 1)
```
acl-cqc: And here....same MeasureFree->Measure+QFree, but also....we see `TKET1.tk1op` + `H` -> nothing; so was the `tk1op` a second H gate that then cancelled with the first? But why would "an H gate with an order edge" be represented as "tk1op" if "MeasureFree with an order edge" was still "MeasureFree"??

PR #1328, opened 2025-12-19
refactor(metadata)!: Migrate all metadata keys onto the new metadata traits

maximilianruesch:
Migrates all metadata keys (that I know of) in `tket` to use the new metadata traits introduced in HUGR, and unifies them to be in one file. This includes some debugging metadata, unitary flags, etc. Finally, some trivial fixes / simplifications are added that come as a convenience of the new metadata system.

I would still argue that this is a refactor, even though a breaking refactor seems contradictional.

Related to https://github.com/Quantinuum/guppylang/pull/1378

BREAKING CHANGE: Renamed the `"unitary"` key to contain the `TKET` prefix as is standard with all other metadata keys.

### DIFF
--- a/tket-py/docs/examples/Guppy-opt-example.ipynb
+++ b/tket-py/docs/examples/Guppy-opt-example.ipynb
@@ -19,17 +19,17 @@
    ]
   },
   {
-   "cell_type": "markdown",
+   "cell_type": "code",
+   "execution_count": null,
    "id": "bdf94fdd",
    "metadata": {},
+   "outputs": [],
    "source": [
-    "```sh\n",
     "!uv pip install \\\n",
     "    \"pytket~=2.11.0\" \\\n",
     "    \"guppylang~=0.21.6\" \\\n",
     "    \"hugr~=0.14.4\" \\\n",
-    "    \"tket~=0.12.13\"\n",
-    "```"
+    "    \"tket~=0.12.14\""
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


tket/src/metadata.rs
```diff
@@ -0,0 +1,75 @@
+//! (Incomplete) Collection of metadata keys used throughout tket.
```
ss2165: what's missing?

tket/src/metadata.rs
```diff
@@ -0,0 +1,75 @@
...
+use hugr_core::metadata::Metadata;
+use tket_json_rs::register::{Bit, Qubit};
+
+/// Metadata key for the number of qubits that a HUGR node expects to be required for execution.
+///
+/// This value is only valid when set at the entrypoint function node. TODO discuss this
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct MaxQubits;
+impl Metadata for MaxQubits {
+    const KEY: &'static str = "TKET.expected_qubits"; // TODO think about this name really hard
```
ss2165: this doesn't line up with the name of the struct or what is used in released guppy

tket/src/metadata.rs
```diff
@@ -0,0 +1,75 @@
...
+    type Type<'hugr> = u32;
+}
+
+/// Metadata key for traced rewrites that were applied during circuit transformation.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct CircuitRewriteTraces;
+impl Metadata for CircuitRewriteTraces {
+    const KEY: &'static str = "TKET.rewrites";
```
ss2165: use lowercase `"tket."`, this is the convention used by the extensions

tket/src/metadata.rs
maximilianruesch: Seems to be an oversight, the collection was previously incomplete, but now every key should be in here.

PR #1333, opened 2025-12-24
feat: hide new public funcs introduced by linearization

acl-cqc:
So following #1325 I was finally able to (/finally succeeded) in evaluating the effect of https://github.com/Quantinuum/hugr/pull/2749 on guppy tests. (There was some question on that PR as to whether we needed to add new facility, see #2766; here is the data.)

Three tests were affected (indeed the same ones as were broken and then fixed by https://github.com/Quantinuum/hugr/pull/2779 but that is a separate issue). The sizes of the LLVM output as follows:

| | Original (hugr 0.24.3) | Hugr 0.25+[fix](https://github.com/Quantinuum/hugr/pull/2779) | Hugr 0.25+[fix](https://github.com/Quantinuum/hugr/pull/2779) + this PR |
|----|----|----|----|
| [new test](https://github.com/Quantinuum/guppylang/pull/1411) | 9568 | 11856 | 9888 |
| basic_type | 10224 | 11616 | 10400 |
| notebook2 | 15680 | 17184 | 15936 |
| notebook5 (Hugr 1) | 10832 | 12304 | 11072 |
| notebook5 (Hugr 2) | 10480 | 11840 | 10656 |

Inspection of the hugrs in the middle row revealed the helper functions from https://github.com/Quantinuum/hugr/pull/2749 were indeed present in the LLVM output as `define` (not `define private`). Hence, this PR, following which LLVM is able to remove the `define`s. (I have not looked into the remaining +2-3% increase. (The numbers include many other changes between hugr-0.24.3 and hugr-0.25.0.)

### DIFF
--- a/tket-qsystem/src/lib.rs
+++ b/tket-qsystem/src/lib.rs
@@ -396,20 +396,20 @@ mod test {
             dfb.finish_hugr_with_outputs([arr2]).unwrap()
         };
 
-        let pub_funcs = |hugr: &Hugr| {
+        let count_pub_funcs = |hugr: &Hugr| {
             hugr.children(hugr.module_root())
                 .filter(|n| match hugr.get_optype(*n) {
                     OpType::FuncDefn(fd) => fd.visibility() == &Visibility::Public,
                     OpType::FuncDecl(fd) => fd.visibility() == &Visibility::Public,
                     _ => false,
                 })
-                .collect_vec()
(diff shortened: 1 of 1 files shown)


acl-cqc:
Test revealed (and is pending) fix in https://github.com/Quantinuum/hugr/pull/2792

PR #1346, opened 2026-01-06
feat: gridsynth pass

km-campbell:
Adding a pass that applies an open source rust implementation of the gridsynth algorithm to all Rz gates in a HUGR. Gridsynth decomposes Rz gates into the Clifford + T basis. Python bindings to enable users to directly modify Guppy-generated HUGRs in Python are included.

As part of the pass, the constant node used to load the angle inputted to the Rz and any intermediary nodes are cleaned up using a garbage collection strategy.

A related demo for the pass is available in the private https://github.com/Quantinuum/gridsynth_guppy_demo.git repo. The demo also functions as a series of manual integration tests for using the pass on Guppy generated HUGRs in Python, albeit using a Jupyter notebook.


tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,521 @@
...
+} // CONCERN: I am concerned that this approach may not clean up properly if the Guppy user
+// has redundant calls of the constant (eg, has used it to define another constant but then
+// not used that constant).
+
+impl GarbageCollector {
+    /// Add references to constant node
+    fn add_references(&mut self, node: Node, increment: usize) {
+        // if reference not in references add it with the default value 1, else increment count
+        let count = self.references.entry(node.index()).or_insert(1);
+        *count += increment;
```
PabloAndresCQ: Considering that you are incremeting by `increment` later, shouldn't you use default value `0` when the entry does not exist?

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,521 @@
...
+            let angle_node = current_node;
+            // Add references to angle node if this has not already been done
+            garbage_collector.infer_references_to_angle(hugr, load_const_node, angle_node);
+            // Remove one reference to reflect the fact that we are about to use the angle node
+            garbage_collector.remove_references(angle_node, 1);
+            // Let garbage collector know what nodes led to the angle node
+            garbage_collector
+                .path
+                .entry(angle_node.index())
+                .or_insert(path);
```
PabloAndresCQ: Nit-pick: due to the assumptions on `NormalizeGuppy` guaranteeing that each `angle_node` is only used once, we _know_ that this entry does not exist in the `garbage_collector`. Adding with `entry().or_insert()` is odd and bug-prone.

PabloAndresCQ (changes requested):

tket-py/tket/passes.py
```diff
@@ -176,3 +177,42 @@ def _normalize(self, hugr: Hugr, inplace: bool) -> PassResult:
...
+
+    # TO DO: make the NormalizeGuppy pass optional, in case it is already run
+    # before Gridsynth. Need to warn users, at least in docs that if NormalizeGuppy
+    # is not run first then Gridsynth is likely to fail. Maybe issue the warning if
+    # the option to run NormalizeGuppy is set to False. The option would be specified
+    # as a field of the dataclass (would also need to add @dataclass decorator)
+    # like for NormalizeGuppy above
+    def run(self, hugr: Hugr, *, inplace: bool = True) -> PassResult:
+        # inplace option does nothing for now but I retain for consistency of
+        # API with other passes
```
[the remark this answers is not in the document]
aborgna-q: Yeah, `implement_pass_run` takes care of dealing with the missing case.

PabloAndresCQ (commented):
I found a case that broke in the current state of the gridsynth PR. If we have something like this:

```
@guppy
def foo():
  // Interesting quantum program here

@guppy
def bar():
  foo()
```

If you run Gridsynth on `foo.compile()` it works, but if you run it on `bar.foo()` it doesn't.  It complains with a "index out of bounds: len 0 but index 0" coming from the Rust code. A workaround is to call `inline_acyclic` to inline every function. I do not know what's actually causing this, it'd be good to solve it with a less drastic approach.

Also, this should be a simple unit test to include.

chloe-quantinuum:
Comments on the updated GridSynth pass and when it succeeds or fails:

The pass is designed to succeed if all rotation angles can be known statically, and throw an `UndefinedAngleError` otherwise. For now the pass only makes a naive attempt at restructuring the HUGR to identify rotation angles by calling `InlineFunctionsPass` and `Normalize`. This allows the pass to be run on HUGRs generated from guppy with nested functions, such as:

``` 
@guppy
def apply_rz(q: qubit, theta: float) -> None:
    rz(q, angle(theta))

@guppy
def main() -> None:
    theta = 0.5
    q = qubit()
    apply_rz(q, theta)
    discard(q)

program = main.compile()
```

or guppy with certain types of classical control flow such as an angle dependent on an if statement where the rotation is applied locally to the branches:

```
@guppy
def main() -> None:
    q1 = qubit()
    q2 = qubit()
    h(q1)
    theta = angle(0.5)
    if measure(q1):
        rz(q2, theta)
    else:
        rz(q2, -theta)
    discard(q2)


program = main.compile()
```

However, if for example we have an angle dependent on an if statement where the rotation is applied outside the branches, such as:

```
@guppy
def main() -> None:
    q1 = qubit()
    q2 = qubit()
    h(q1)
    result = measure(q1)
    if result:
        theta = angle(0.5)
    else:
        theta = angle(1.0)
    rz(q2, theta)
    discard(q2)


program = main.compile()
```

the pass will not be able to determine the angle and throw an `UndefinedAngleError`.  To resolve such issues we can either:
- create more pre-processing passes to flatten the program
- feedback information to users about how they can rewrite their programs - basically make them flatten it for us :sweat_smile:

I'm not sure how much of a priority this would be (provided we get runtime gridsynth), but the work would likely have a lot of overlap with potential future work to further flatten the HUGR for optimisation purposes. The main passes that would be useful are loop unrolling, and a variant of hoisting that transforms
```
if condition:
    C1;
else:
    C2;
C3;
```
into
```
if condition:
    C1;
    C3;
else:
    C2;
    C3;
```
Unfortunately, both of these passes would make the size of the HUGR explode.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    simplify(gridsynth_gates(&mut config).gates)
+}
+
+/// Compresses a gridsynth gate sequence into a shorter normal form.
+fn simplify(gates: String) -> String {
+    let mut gates = gates;
+    let n = gates.len();
+    let mut normal_form_reached = false;
+    while !normal_form_reached {
+        let new_gates = gates
```
aborgna-q: This should use something like KMP to avoid multiple traversals.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    let _guard = GRIDSYNTH_LOCK.lock().unwrap();
+    let seed = 1234;
+    let verbose = false;
+    let up_to_phase = false;
+    let mut config = config_from_theta_epsilon(theta, epsilon, seed, verbose, up_to_phase);
+    simplify(gridsynth_gates(&mut config).gates)
+}
+
+/// Compresses a gridsynth gate sequence into a shorter normal form.
+fn simplify(gates: String) -> String {
```
aborgna-q, line 159: The `gridsynth_gates` call before this already uses `rsgrydsyth::normal_form::NormalForm`, and leaves the gates in a `(T | HT | SHT)* + Clifford` normalized sequence.

Do we need further simplification afterwards? (I believe many of the patterns here won't appear in the normalized form anyways).

_If_ we still need this function, it should be noted that each `replacen` is doing a full clone of the string so the current code will be _really_ inneficient.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    let new_nodes: Vec<Node> = gates
+        .chars()
+        .map(|gate| match gate {
+            'H' => TketOp::H,
+            'S' => TketOp::S,
+            'T' => TketOp::T,
+            'D' => TketOp::Tdg,
+            'X' => TketOp::X,
+            'Z' => TketOp::Z,
+            _ => panic!("The gate {gate} is not supported"),
```
aborgna-q, line 216: Gridsynth produces an `I` string on `Rz(0.0)` here, so this will panic.
Make sure to add a test.

Also, please return an error rather than panicking on an unknown gate.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+/// Replace an `Rz` node with the Clifford+T gates in `gates`.
+fn replace_rz_with_gates<H: HugrMut<Node = Node>>(
+    hugr: &mut H,
+    rz_node: Node,
+    gates: &str,
+) -> Result<(), GridSynthError> {
+    // W is a global phase factor so we can ignore it
+    let gates = gates.replacen('W', "", gates.len());
```
aborgna-q, line 205: - Why `replacen` instead of `replace`?

- If we take the owned `gates: String` here we can avoid cloning it by using `gates.retain(|c| c != 'W')` instead.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    type Result = ();
+
+    fn run(&self, hugr: &mut H) -> Result<(), Self::Error> {
+        InlineFunctionsPass::default().run(hugr)?;
+        Normalize::default().run(hugr)?;
+
+        let rz_nodes: Vec<Node> = hugr
+            .nodes()
+            .filter(|n| hugr.get_optype(*n).cast::<TketOp>() == Some(TketOp::Rz))
+            .collect();
```
aborgna-q, line 79: The pass does not respect the `PassScope` configuration.
That informs which nodes we are allowed to modify, rather than running over the whole hugr.

When calling other passes, this should propagate the config with `.with_scope(self.scope.clone())`.

For `rz_nodes` you could do something like
```rs
let rz_nodes = self
    .scope
    .regions(hugr)
    .flat_map(|parent| hugr.children(parent))
    .filter(|n| hugr.get_optype(*n).cast::<TketOp>() == Some(TketOp::Rz))
    .collect_vec();
```

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+
+            let theta = find_angle(hugr, const_node);
+            hugr.remove_node(source_node);
+
+            let const_out = hugr
+                .node_outputs(const_node)
+                .next()
+                .expect("Const has a static output");
+
+            if !hugr.is_linked(const_node, const_out) {
```
aborgna-q, line 112:
```suggestion
            if !hugr.is_linked(const_node, const_out) && self.scope.in_scope(hugr, const_node) == InScope::Yes {
```

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    }
+}
+
+/// Runs the gridsynth algorithm on `theta` (radians), and returns the gate string.
+fn gridsynth(theta: f64, epsilon: f64) -> String {
+    let _guard = GRIDSYNTH_LOCK.lock().unwrap();
+    let seed = 1234;
+    let verbose = false;
+    let up_to_phase = false;
+    let mut config = config_from_theta_epsilon(theta, epsilon, seed, verbose, up_to_phase);
```
aborgna-q, line 154: We are doing `up_to_phase = false` here, but then dropping the `W`s in `replace_rz_with_gates` and doing non-phase-preserving rewrites in `simplify`.

Do we want this to be phase preserving? Either way, it should be clearly stated in the pass docs.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// The angle of the Rz gate cannot be determined statically.
+    #[error(ignore)]
+    #[display("Could not determine angle of {_0} statically")]
+    UndefinedAngleError(Node),
+}
+
+/// Applies the gridsynth algorithm to synthesise arbitrary rotations to Clifford+T.
+///
+/// Applies [InlineFunctionsPass] and [Normalize] to flatten the HUGR so angles can be known
+/// statically. When this is not possible, we return a [GridSynthError::UndefinedAngleError].
```
aborgna-q, line 42: These are normally already applied when running on a program compiled from guppy.

Add an option flag to disable the pre-processing.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
+//! A pass that applies the gridsynth algorithm to synthesise arbitrary ratations to Clifford+T.
```
aborgna-q, line 1:
```suggestion
//! A pass that applies the gridsynth algorithm to synthesise arbitrary rotations to Clifford+T.
```

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+            })
+            .to_radians()
+    } else {
+        panic!("{const_node} has unexpected value type (expected ConstRotation or ConstF64)")
+    }
+}
+
+/// Runs the gridsynth algorithm on `theta` (radians), and returns the gate string.
+fn gridsynth(theta: f64, epsilon: f64) -> String {
+    let _guard = GRIDSYNTH_LOCK.lock().unwrap();
```
aborgna-q, line 150: We can recover from poison here
```suggestion
    let _guard = GRIDSYNTH_LOCK.lock().unwrap_or_else(|e| e.into_inner());
```

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+        Self {
+            scope: PassScope::default(),
+            epsilon: 1e-3,
+        }
+    }
+}
+
+impl GridSynthPass {
+    /// Sets the precision of the gridsynth approximation.
+    pub fn with_epsilon(mut self, epsilon: f64) -> Self {
```
aborgna-q, line 62: nit: Fail early on epsilon values that will make the search loop forever (`<= 0` / `NaN` / etc.)

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+        rot.to_radians()
+    } else if let Some(fl) = value.get_custom_value::<ConstF64>() {
+        ConstRotation::new(fl.value())
+            .unwrap_or_else(|_| {
+                panic!(
+                    "ConstF64 value {:?} for node {const_node} is not a valid rotation",
+                    fl.value()
+                )
+            })
+            .to_radians()
```
aborgna-q, line 142: nit: If the only allowed sequence is `Const -> LoadConst -> Rz` this branch will never be reached (it'd need a conversion operation in the middle).

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+    } else {
+        panic!("{const_node} has unexpected value type (expected ConstRotation or ConstF64)")
+    }
+}
+
+/// Runs the gridsynth algorithm on `theta` (radians), and returns the gate string.
+fn gridsynth(theta: f64, epsilon: f64) -> String {
+    let _guard = GRIDSYNTH_LOCK.lock().unwrap();
+    let seed = 1234;
```
aborgna-q, line 151: This could be a pass option (with a `with_seed` setter method)

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+}
+
+impl WithScope for GridSynthPass {
+    fn with_scope(mut self, scope: impl Into<PassScope>) -> Self {
+        self.scope = scope.into();
+        self
+    }
+}
+
+// Testing guppy-generated HUGRs from test_files/guppy_optimization/gridsynth/
```
aborgna-q, line 251: It'd be nice to test some unitaries rather than just gate counts

aborgna-q (changes requested):

tket-py/tket/passes.py
```diff
@@ -176,3 +177,42 @@ def _normalize(self, hugr: Hugr, inplace: bool) -> PassResult:
...
+    Parameters:
+    - epsilon: The allowable error tolerance.
+    """
+
+    # TO DO: make the NormalizeGuppy pass optional, in case it is already run
+    # before Gridsynth. Need to warn users, at least in docs that if NormalizeGuppy
+    # is not run first then Gridsynth is likely to fail. Maybe issue the warning if
+    # the option to run NormalizeGuppy is set to False. The option would be specified
+    # as a field of the dataclass (would also need to add @dataclass decorator)
+    # like for NormalizeGuppy above
```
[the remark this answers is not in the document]
aborgna-q: It's fine, apart from the performance hit of running everything twice.

tket/src/passes/gridsynth.rs
```diff
@@ -0,0 +1,383 @@
...
+
+        for rz_node in rz_nodes {
+            let angle_port = hugr
+                .node_inputs(rz_node)
+                .nth(1)
+                .expect("Rz should have an angle input");
+
+            let (source_node, _) = match hugr.single_linked_output(rz_node, angle_port) {
+                Some(link) => link,
+                _ => return Err(GridSynthError::UndefinedAngleError(rz_node)),
```
aborgna-q, line 89: > The pass is designed to succeed if all rotation angles can be known statically, and throw an UndefinedAngleError otherwise.

When that error is raised the hugr is left in a half-transformed state; some angles may have already been transformed while others are left untouched.

Could we do this in two passes; collecting all the `(angle, rz_node)` first?
We could either return the error early without modifying anything, or process all valid angles before returning all errors.

PR #1358, opened 2026-01-15
fix: Wrongly reused qubit IDs in pytket encoding

aborgna-q:
Fixes a bug reported by @CalMacCQ where an encoded pytket circuit would use the same qubit identifier on multiple registers.

The root cause was the encoder being too eager in marking qubits for re-use.

In the following dataflow graph, when processing `c` we first looked at the incoming connection from `a` and and de-registered the wire (since `c` is the only consumer, there's no need to track it anymore). This used to also deregister the values in the wire, so the qubit register got added to the list of unused ids.
The ID would normally have been marked as _used_ again when registering `c`'s outputs, but in this case when processing the incoming connection from `b` we would emit an opaque barrier that generates a new qubit. For that, we checked the list of unused qubits and ended up using the same identifier for both of `c`'s inputs.

```mermaid
graph LR
  a--"qubit"-->c
  b["b (unsupported)"]--"qubit"-->c
```

The here simply detaches the wire de-registration logic from the qubit freeing. We do the latter only when we see an input qubit doesn't appear in the outputs.

### DIFF
--- a/tket1-passes/tests/tket1-on-hugr.rs
+++ b/tket1-passes/tests/tket1-on-hugr.rs
@@ -52,12 +52,8 @@ fn circ_flat_quantum() -> Circuit {
     build().unwrap()
 }
 
-/// Circuit extracted from callum's bug report
-/// https://cambridgequantum.slack.com/archives/C09F952LKM4/p1768405383883349
+/// A circuit with some (unsupported) borrow array operations.
 fn circ_borrow_array() -> Circuit {
-    //let envelope: &[u8] = include_bytes!("callum-bug-report-normalized.hugr");
-    //Circuit::load(envelope, None).unwrap()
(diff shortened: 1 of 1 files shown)


tket/src/serialize/pytket/encoder/value_tracker.rs
```diff
@@ -315,6 +315,16 @@ impl<N: HugrNode> ValueTracker<N> {
         })
     }
 
+    /// Frees a tracked qubit, and allow re-using it's ID when calling [`ValueTracker::new_qubit`].
```
cqc-alec:
```suggestion
    /// Frees a tracked qubit, and allow reusing its ID when calling [`ValueTracker::new_qubit`].
```

PR #1370, opened 2026-01-21
feat!: handling multiple QSystem gatesets

jake-arkinstall:
This PR aims to explore the work required to add a second QSystem gateset within the framework of the existing gateset.

It is **not complete**, as rebasing work is required, but reviews will be extremely useful for knowing that this is going in the right direction.

The design choice is "least invasive". There is a good argument for a more invasive refactor to handle things naturally, but that is a longer term effort.

As a consequence of this least invasive approach, the `QSystemOpBuilder` trait remains a trait, and thus now requires platform arguments for the `add` and `build` functions (and for 1q operations this parameter is presently unused), which adds a lot of noise to the screen when reading and writing, and uses match blocks for when rebasing differs. I would like to explore handling this in a better manner.

Another design choice is to bundle all gates into QSystem itself. The intent is to have user-facing guppy code fetch gateset-specific functions from a submodule of qsystem (perhaps re-exporting Helios gates from the parent module to retain backwards compatibility), but on the tket definition side they're still in one big "bag". This choice is made because of the overlap between the two gatesets' 1q operations (rxy, rz) and platform functions (qalloc etc), and the intent is that the more invasive refactor will take gateset overlaps into account while hopefully introducing more separation.

BREAKING CHANGE: QSystemOpBuilder trait's `build_*` and `add_*` functions take an additional `platform` argument. LowerTketToQSystemPass now has a `platform` member and no longer satisfies Default (a platform must be provided, and there is no good 'default' platform).


ss2165:
Superceded by and #1567  #1647

PR #1376, opened 2026-01-27
Gridsynth nit-picks and gate sequence normalisation

PabloAndresCQ:
Merging into the open draft PR #1346 with some improvements

### DIFF
--- a/tket/src/passes/gridsynth.rs
+++ b/tket/src/passes/gridsynth.rs
@@ -3,10 +3,10 @@
 /// (2) that NormalizeGuppy has been run. It expects that the following is guaranteed:
 /// * That every Const node is immediately connected to a LoadConst node.
 /// * That every constant is used in a single place (i.e. that there's a single output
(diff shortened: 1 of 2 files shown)


tket/src/passes/gridsynth.rs
```diff
@@ -200,113 +206,107 @@ fn apply_gridsynth(
...
-    gates.gates
+    let mut gate_sequence = gridsynth_gates(&mut gridsynth_config).gates;
+
+    if simplify {
+        let n = gate_sequence.len();
+        let mut normal_form_reached = false;
+        while !normal_form_reached {
+            // Not the most efficient, but it's easiest to reach the normal form by doing
+            // string rewrites.
+            // TODO: Can be done with Regex, preferably by providing all LHS to the Regex
```
chloe-quantinuum: Just wondering, do you have any idea of the efficiency of using regex vs these replacements?

PR #1413, opened 2026-02-20
ci: Use pre-built tket-c-api binaries

aborgna-q:


maximilianruesch:
Has been addressed in #1471

PR #1416, opened 2026-02-23
feat: Add new tket-qec crate with Iceberg code extension.

### LINKED ISSUE #1400: [Feature]: Add extension for iceberg code
### Problem Statement

models logical operation of iceberg code

transversal_cx[N]: [Block[N],Block[N]] -> [Block[N],Block[N]}
permute_block[N]: [Block[N], N usizes ...] -> [Block[N}] # N numbers specifying the permutation. Panics on repeated indices
prepare_0[N]: [] -> [Block[N]]
prepare_+[N]: [] -> [Block[N]]
prepare_combo[N]: [N bools ...] -> [Block[N]] # false means 0, true means +. previous two could be implemented with this
prepare_magic_blah[N]: [] -> [Block[N]]
measure[N]: [Block[N]] -> array[N,future[bool]] # this consumes the block. We may also want a variant that returns the projected block

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

cqc-alec:
Closes #1400 .

I was in two minds about putting this in the tket2 repo or making a new repo for logical code extensions; still open to persuasion either way. Of course, the follow-up to this is compiling from computational to logical and from logical to physical; and that probably does belong in tket...


tket-qec/src/iceberg/ops.rs
```diff
@@ -0,0 +1,699 @@
...
+                .as_ref(),
+            &Signature::new(block_type(6), block_type(6))
+        );
+    }
+
+    #[test]
+    fn test_hugr_ops() {
+        let block = block_type(6);
+        let x3 = EXTENSION
+            .instantiate_extension_op("X", [6.into(), 3.into()])
```
PabloAndresCQ: Oh, so for every operation we need to indicate the number of logical qubits per block? I thought the conclusion was that there would be a separate extension per `k` (generated automatically).

tket-qec/src/iceberg/ops.rs
```diff
@@ -0,0 +1,699 @@
...
+fn vec_of_blocks_and_bools(n_blocks: usize, n_bools: usize) -> Vec<Type> {
+    let mut types: Vec<Type> = vec![block_tv(0); n_blocks];
+    types.extend(vec![bool_t(); n_bools]);
+    types
+}
+
+fn block_with_angles_sig(n_angles: usize) -> FuncValueType {
+    FuncValueType::new(
+        vec_of_blocks_and_angles(1, n_angles),
+        vec_of_blocks_and_angles(1, 0),
```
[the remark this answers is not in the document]
cqc-alec: Yes, it's the type of the output. It's not quite `block_tv(0)` but a vector containing a single one of those.

tket-qec/src/iceberg/ops.rs
```diff
@@ -0,0 +1,699 @@
...
+
+    fn extension(&self) -> ExtensionId {
+        EXTENSION_ID.clone()
+    }
+
+    fn extension_ref(&self) -> Weak<Extension> {
+        Arc::downgrade(&EXTENSION)
+    }
+
+    fn init_signature(&self, _extension_ref: &Weak<Extension>) -> SignatureFunc {
```
doug-q: This interface does not allow dynamic indexing in any ops. Is this intentional and Ok??

cqc-alec: That is a very good point which I hadn't considered. I'm not at all sure that it is OK.

tket-qec/src/iceberg/ops.rs
```diff
@@ -0,0 +1,699 @@
...
+    /// X gate on all qubits.
+    AllX,
+    /// Y gate on all qubits.
+    AllY,
+    /// Z gate on all qubits.
+    AllZ,
+    /// X gate on one qubit with Z on all others.
+    XWithAllButOneZ,
+    /// Z gate on one qubit with X on all others.
+    ZWithAllButOneX,
```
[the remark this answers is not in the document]
PabloAndresCQ: Better not to remove them yet. I am questioning if it makes sense to have Pauli frame tracking on Iceberg. The use case of applying QEC corrections in SW doesn't apply on a d=2 code. 

I've implemented a prototype of Pauli frame tracking and I'm not convinced about it. I'll make a draft PR to gather feedback.

PR #1422, opened 2026-02-26
feat!: Upgrade LLVM code in tket-qsystem and qis-compiler

### LINKED ISSUE #1421: [Feature]: Upgrade to LLVM 21
### Problem Statement

Need to upgrade Inkwell code in qis-compiler and tket-qsystem over API breakages between 14 and 21

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

cgh-qtnm:
BREAKING CHANGE: The hugr-llvm LLVM version is being upgraded to 21.1, and code that depends on it here needs to be updated as well. This also includes an MSRV change to 1.91 to match hugr. See [the hugr PR](https://github.com/Quantinuum/hugr/pull/2901) for details.

Closes #1421

### DIFF
--- a/.github/workflows/python-qis-wheels.yml
+++ b/.github/workflows/python-qis-wheels.yml
@@ -46,7 +46,8 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - ubuntu-24.04-arm
+          # TODO: Temporarily disabled until we create a manylinux_2_28 compatible docker image for ARM with LLVM 21.
+          #- ubuntu-24.04-arm
           - macos-15
           - macos-15-intel
           - windows-2022


Cargo.lock
```diff
@@ -1112,8 +1111,7 @@ dependencies = [
 [[package]]
 name = "hugr-cli"
 version = "0.25.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b6baee3cc2fe4d7c5635a630341e28237f12a8e5beef31b5aaaebe93bc9ba90a"
+source = "git+https://github.com/quantinuum/hugr?branch=george%2Fllvmup#7d8587c64086ea5a2a4486b23ef7b73ab2c44040"
```
cqc-alec: I'm not sure that we want to be depending on these branches from `main` here. Should we wait for an updated release of hugr?

cqc-alec:
Have given it a superficial review (and ignored all the snaps assuming those changes make sense). LGTM!

cgh-qtnm:
@aborgna-q This should be ready for you to merge into the release branch. Notes:

1. Fixed what Iooks like a bug where `BoolOp::not` had a signature with two inputs.
2. Wrap a bunch of `TypeBase<NoRV>` in `vec!` so that they will cast to a single-element `TypeRowBase<NoRV>`. Unless I'm missing something this is just due to an oversight in the type row code, we should fix it in the next HUGR release. If there's a cleaner way to fix this feel free to revert and replace.
3. Removed remnants of stack array support.

Still a few bugs. In particular, the one about failure to monomorphize `_range_comptime` in the `qis-compiler` tests might be a hugr-llvm problem, but I'm not sure exactly what's going wrong. Let me know if further attention is needed there from me.

Issue #1163, opened 2025-10-07
Rust direct callable HUGR to QIS compilation


qciaran:
Instead of only wrapping the HUGR to QIS compilation in Python/PyO3, can it be callable directly in Rust?

This function appears to be a PyO3 only function: https://github.com/CQCL/tket2/blob/0fb0ac92e61e6ec7699e1de4cb4cc3cc5a68bad1/qis-compiler/rust/lib.rs#L463

Especially since it seems only a cdylib is generated:
https://github.com/CQCL/tket2/blob/0fb0ac92e61e6ec7699e1de4cb4cc3cc5a68bad1/qis-compiler/Cargo.toml

(Unless I am being dumb... I think this can only be used from Python.)

Related to issue: https://github.com/CQCL/selene/issues/86

Reason for wanting this is to push down as much of the code in PECOS into Rust as possible. So it would be nice if only Guppy -> HUGR is in Python and the rest is handled in Rust.

qciaran:
I wonder if just adding "rlib" to `crate-type` would solve the issue

ss2165:
Potentially, might need to make the module `pub` as well

PR #1428, opened 2026-02-28
feat: Make qis-compiler callable from Rust crates

qciaran:
Currently the qis-compiler can only be called from Python. This aims to add minimal changes to make it callable from Rust crates too (e.g., in PECOS).

This aims to add no breaking changes to those already using the Python API. It only adds functionality for Rust users.

Closes Issue #1163

### DIFF
--- a/qis-compiler/Cargo.toml
+++ b/qis-compiler/Cargo.toml
@@ -11,11 +11,11 @@ edition = "2024"
 [lib]
 name = "selene_hugr_qis_compiler"
 path = "rust/lib.rs"
-crate-type = ["cdylib"]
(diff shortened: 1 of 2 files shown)


ciaranra:
This mainly just let's me remove one of the crates in PECOS and directly use tket2 for compiling to QIS. So this isn't a blocker for me. I can either keep my crate or maybe point to the commit in the fork for now (although don't love that since the commit might disappear eventually).

So if it is worth holding off, that is fine. Just in the future hopefully can call the QIS compiler from Rust

maximilianruesch:
Superseded by #1864

PR #1449, opened 2026-03-16
feat: InlineFunctionsPass supports "Always" annotation that errors if unsatisfiable

acl-cqc:
Intended as precursor to #1532: this adds the "Always" option only, but takes into account in its own pass, which *raises an error* if it cannot obey all the annotations.

Python interface is very WIP, not sure I've grokked everything there yet.


tket-py/test/test_pass.py
```diff
@@ -285,6 +288,52 @@ def test_modifier_execution() -> None:
...
+    d = Dfg(tys.Tuple(tys.Qubit, tys.Qubit))
+
+    f_id = d.module_root_builder().define_function(
+        "id",
+        [tys.Qubit],
+    )
+    f_id.set_outputs(f_id.input_node[0])
+
+    if annotate:
+        f_id.metadata["tket.inline"] = "always"
```
acl-cqc: We should export the metadata key from rust, as hugr-py/rust/metadata.rs does, but we don't have a module set up for that yet

tket-py/test/test_pass.py
```diff
@@ -285,6 +288,52 @@ def test_modifier_execution() -> None:
...
+
+    (tup,) = d.inputs()
+    (q1, q2) = d.add(ops.UnpackTuple()(tup))
+    call1 = d.call(f_id, q1)
+    call2 = d.call(f_id, q2)
+    (tup,) = d.add(ops.MakeTuple()(call1, call2))
+
+    d.set_outputs(tup)
+
+    # validate(d.hugr)
```
acl-cqc: I guess there is no way to do this here?

tket-py/tket/passes/__init__.py
```diff
@@ -12,7 +12,7 @@
 from tket import _state
 from . import inline_funcs
 from .._tket import passes as _passes, optimiser as _optimiser
-
+from .._tket.passes import InlineAlwaysError  # type: ignore
```
acl-cqc: I think the `type: ignore` means I need to declare this in `passes.pyi` but I'm not (yet) sure exactly what to add there. We should do the same for InlineFunctionsError and ModifierResolverError too (unless I'm missing something?)

tket-py/test/test_pass.py
aborgna-q:
```py
CompilationState.from_python(d.hugr).validate()
```

tket-py/test/test_pass.py
aborgna-q: Added the python exports to #1532.

tket-py/test/test_pass.py
acl-cqc: Ah, fab, thanks 🙌

tket-py/test/test_pass.py
```diff
@@ -307,6 +311,50 @@ def test_modifier_execution() -> None:
...
+    d = Dfg(tys.Tuple(tys.Qubit, tys.Qubit))
+
+    f_id = d.module_root_builder().define_function(
+        "id",
+        [tys.Qubit],
+    )
+    f_id.set_outputs(f_id.input_node[0])
+
+    if annotate:
+        f_id.metadata["tket.inline"] = "always"
```
aborgna-q:
```suggestion
        f_id.metadata[InlineAnnotation] = "always"
```

tket-py/tket/metadata.py
```diff
@@ -74,7 +74,7 @@ class MaxQubitsHint(Metadata[int]):
     KEY = _metadata.MAX_QUBITS_HINT
 
 
-InlineAnnotationValue: TypeAlias = Literal["never"] | Literal["best_effort"]
+InlineAnnotationValue: TypeAlias = Literal["never", "best_effort", "always"]
 
 
 class InlineAnnotation(Metadata[InlineAnnotationValue]):
```
aborgna-q, line 80: This needs updates in the docs and in `from_json`

tket/src/passes/inline_always.rs
```diff
@@ -0,0 +1,376 @@
...
+    #[display("Cycle detected in functions marked to Always inline: {_0:?}")]
+    AlwaysCycle(Vec<N>),
+}
+
+impl<N: std::fmt::Debug> std::error::Error for InlineAlwaysError<N> {}
+
+/// A [ComposablePass] that inlines `Call`s to functions
+/// according to [InlineAnnotation]s.
+#[derive(Default, Clone, Debug)]
+pub struct InlineAlwaysPass {
```
aborgna-q: We should explain what are the differences between this pass an `InlineFunctionsPass`, not only here but in the other pass' docs and on the python definitions too.

tket/src/passes/inline_funcs.rs
```diff
@@ -81,7 +81,8 @@ impl<H: HugrMut> ComposablePass<H> for InlineFunctionsPass {
                 match h.get_metadata::<InlineAnnotation>(func) {
                     Some(InlineAnnotation::Never) => false,
                     Some(InlineAnnotation::BestEffort) => true,
-                    None => self.heuristic.should_inline(func, h),
+                    Some(InlineAnnotation::Always) // instruction to InlineAlwaysPass
```
aborgna-q: I'm thinking we should treat `Always` as `BestEffort` here. We won't inline every occurrence, but at least we should inline the ones we can.

tket/src/metadata.rs
```diff
@@ -49,6 +49,10 @@ impl Metadata for MaxQubitsHint {
 #[serde(rename_all = "snake_case")]
 #[non_exhaustive]
 pub enum InlineAnnotation {
+    /// Always inline calls to this function.
+    ///
+    /// If this cannot be done, an error will be raised.
```
aborgna-q: Would probably merit a mention to the different inline passes.

aborgna-q (commented):
Some comments, mostly about differentiating the two passes.

Although from the technical side it makes sense to have two of them (we need to do different kinds of traversals for each), I struggle to justify having the split passes from the user point of view.
When would I want to "inline some of the annotated functions, but not the others that I've also annotated"?

tket/src/metadata.rs
acl-cqc: Now only one pass so leaving.

tket-py/tket/metadata.py
acl-cqc: Thanks! Done

tket-py/test/test_pass.py
acl-cqc: Done, thanks

acl-cqc:
> Some comments, mostly about differentiating the two passes.
> 
> Although from the technical side it makes sense to have two of them (we need to do different kinds of traversals for each), I struggle to justify having the split passes from the user point of view. When would I want to "inline some of the annotated functions, but not the others that I've also annotated"?

Yeah, totally agree. Have merged the two. I didn't really go looking for common code and I imagine there must be at least some but see how you think this looks now?

tket/src/passes/inline_funcs.rs
```diff
@@ -155,10 +168,140 @@ pub fn inline_acyclic_scoped<H: HugrMut>(
     Ok(())
 }
 
+fn inline_always_scoped<H: HugrMut>(
```
aborgna-q, line 172: Add method docs

tket/src/passes/inline_funcs.rs
```diff
@@ -155,10 +168,140 @@ pub fn inline_acyclic_scoped<H: HugrMut>(
...
+            debug_assert!(hugr.static_targets(func).unwrap().next().is_none());
+            if !funcs_to_preserve.contains(&func) {
+                hugr.remove_subtree(func);
+            }
+        }
+    }
+    Ok(())
+}
+
+fn cycles<'a, N: Copy>(
```
aborgna-q: Docs

tket/src/metadata.rs
```diff
@@ -49,6 +49,10 @@ impl Metadata for MaxQubitsHint {
...
+    Always,
```
aborgna-q: The semver checks are failing because this was added before all existing variants, and changed their inner tag id.
It also breaks the `Ord` derive.

Can you move the new variant to the end?
Alternatively, do
```rs
    Always = 2,
    BestEffort = 0,
```

tket/src/passes/inline_funcs.rs
```diff
@@ -155,10 +168,140 @@ pub fn inline_acyclic_scoped<H: HugrMut>(
...
+    let mut seen = HashSet::new();
+    while let Some(parent) = parents.pop_front() {
+        if hugr.get_optype(parent).is_func_defn() {
+            seen.insert(parent);
+        }
+        let mut to_inline = Vec::new();
+        for child in hugr.children(parent) {
+            if hugr.first_child(child).is_some() {
+                parents.push_back(child);
+            } else if hugr.get_optype(child).is_call()
```
aborgna-q: We're not handling `LoadFunction`s here, even though the `reachable_always` logic above does add them

https://github.com/Quantinuum/tket2/blob/566dc2bd3252ab8749ac1c1b3ba1120bc39249e5/tket/src/passes/inline_funcs.rs#L202-L203

acl-cqc: That's a really good point, thank you! Should ignore LoadFunction edges when looking for cycles, as they can't be inlined (yet we push them onto `to_inline`, ooops); and not delete functions with LoadFunctions remaining. Quite a few bugs there, which tests including LoadFunction would have spotted.... I've updated code but not added the tests. I think we'll have to let this one slip to a later release....it is non-breaking after all.

PR #1451, opened 2026-03-18
feat: Allow modifier use with qis-compiler

jake-arkinstall:
Presently, the use of (experimental?) guppylang modifiers, such as:
```python
@guppy
def main() -> None:
    q1 = qubit()
    q2 = qubit()
    h(q1)
    with control(q1):
        x(q2)
    
    result("q1", measure(q1))
    result("q2", measure(q2))
```

Results in
```
RuntimeError: No extension could emit extension op: (IdentList("tket.modifier"), "ControlModifier")
```

This PR adds use of the `ModifierResolverPass` before QSystemPass to resolve modifiers, enabling compilation to QIS.


aborgna-q:
Update: This will be added to NormalizeGuppy instead.
https://github.com/Quantinuum/tket2/issues/1640

PR #1463, opened 2026-03-24
docs: Fix tket README introductory example

### LINKED ISSUE #1362: Fix "Using TKET" example in README
It doesn't work as there is no `json` module anymore.

maximilianruesch:
There were a couple of more issues than just the `json` module missing. This variant has been tried and tested locally. Perhaps we should add some tests for this? I don't know whether we can.

Closes #1362


maximilianruesch:
Even uncovered a subtle bug in the readme code 👍

PR #1489, opened 2026-03-31
feat(const-fold): combine Consts for each out-port; write prepopulated inputs

acl-cqc:
Previously we made a list of *input ports* whose incoming edges to break with a Const+LoadConstant, thus duplicating the Const for any outport with multiple edges. Instead, place one Const+LoadConstant per outport (that has >=1 edge), and reroute all the old edges to come from the same LoadConstant. (These could be nonlocal, iff the original edges were.)

Also remove the `mb_root_inp` mechanism that had been used to prevent values passed to `prepopulate_wire` / `prepopulate_inputs` being written into the graph - writing these in is consistent with what we do to other ops. (There are some tests using `assert_fully_folded` that this would have broken without the first/Const-combining change; and bugs fixed in #1470 and #1474  that `mb_root_inp` was hiding; but now these are solved, removal is straightforward.)

### DIFF
--- a/tket/src/passes/const_fold.rs
+++ b/tket/src/passes/const_fold.rs
@@ -139,7 +139,7 @@ impl<H: HugrMut<Node = Node> + 'static> ComposablePass<H> for ConstantFoldPass {
 
         let results = m.run_subtree(ConstFoldContext, root);
         let mut root_descs = hugr.descendants(root);
-        assert_eq!(root_descs.next(), Some(root)); // Skip root
+        assert_eq!(root_descs.next(), Some(root), "Could not skip root");
         let wires_to_break = root_descs
             .filter(|n| !hugr.get_optype(*n).is_load_constant()) // no point in adding another Const!
             .flat_map(|n| hugr.out_value_types(n).map(move |(outp, _ty)| (n, outp)))


tket/src/passes/const_fold.rs
```diff
@@ -140,21 +138,16 @@ impl<H: HugrMut<Node = Node> + 'static> ComposablePass<H> for ConstantFoldPass {
...
-            .filter_map(|(n, ip)| {
-                let (src, outp) = hugr.single_linked_output(n, ip).unwrap();
-                // Avoid breaking edges from existing LoadConstant (we'd only add another)
-                // or from root input node (any "external inputs" provided will show up here
-                //   - potentially also in other places which this won't catch)
-                (!hugr.get_optype(src).is_load_constant() && Some(src) != mb_root_inp).then_some((
-                    n,
-                    ip,
+        let mut root_descs = hugr.descendants(root);
+        assert_eq!(root_descs.next(), Some(root)); // Skip root
```
ss2165:
```suggestion
        assert_eq!(root_descs.next(), Some(root), "unexpected non-root first descendant"); // Skip root
```

acl-cqc:
> Thanks Alan, this looks good.
> 
> I don't like the `assert_fully_folded` test cases, they are too strict. I think making this change for that reason is letting the tail wag the dog.

Yes, that's very true, good point. Rather I think placing Const nodes like this is *the right thing to do* - I'm not quite sure why we had the old approach, possibly so that we didn't insert any consts for unconnected outports, but that's easily solved by a filter.

That it means we don't need to rewrite `assert_fully_folded` is a "nice" (i.e. lazy/labour-saving) side-effect...and so if the removal of `mb_root_inp` is seen as the *second* thing we're doing here (could be another PR, but becomes a two-line change!) then it seems natural enough. Description/title rewritten, thus ;)

> I think writing prepopulated_wires values into the graph is the right thing to do, kind of surprised we weren't already.

This is pretty old, but I think in pre-entrypoint days, if you'd run on a DFG-rooted Hugr, and specified inputs, it might have tried to write them *outside the root*...

> This change will create fewer const nodes.
> 
> I suspect that this change will make fixed-point passes (constant-fold, optimise, loop until no changes) perform worse on pathalogical graphs, because we are not reducing the work for future loops anymore. I don't think this is currently relevant, and surely not a blocker.

Not sure what you mean here - we do have a special-case for not breaking out-edges from LoadConstants (😆), which we're keeping. If you ConstFold, then LocalizeEdges, then ConstFold again, you'll probably duplicate the constants deeper into the Hugr (disconnecting edges from Input), but I think that's OK.

The other case would be if you wanted to run only on a subgraph/subregion, this'll likely mean you have more nonlocal edges entering that subregion from outside. At the moment we don't support that yet (we analyse the whole Hugr, then just filter results down to the subregion), but if we do implement that (i.e. stop analysing outside the subregion), then if all those incoming nonlocals are `Top` *then* you'll suffer. That might be worth a special-case.

PR #1507, opened 2026-04-09
chore: Remove `pytket` dependency from `tket-py`

aborgna-q:
After #1482 we no longer use `pytket` from the python side other than for type-checking `PytketHugrPass` and when testing.
Moved the dependency to a `pytket` extra. This is only required to have proper typing for the pass, and to run some pytket optimisation tests.

~Now that `pytket` is not required to build and test the wheels, we can produce free-threaded wheels for python 3.13t (pytket doesn't publish those).~
Edit: Removed the free-threaded wheels build.
I'll leave it for later, once hugr publishes `3.14t` releases (https://github.com/Quantinuum/hugr/pull/3021).


tket-py/test/test_pauli_prop.py
```diff
@@ -191,13 +194,13 @@ def test_cat(propagate_matcher: RuleMatcher):
...
-    assert t2c.to_tket1() == Circuit(4).H(2).X(2).CX(2, 1).CX(2, 3).CX(1, 0)
+    assert t2c.to_tket1() == pytket.Circuit(4).H(2).X(2).CX(2, 1).CX(2, 3).CX(1, 0)
 
     assert apply_exhaustive(t2c, propagate_matcher) == 3
 
-    assert t2c.to_tket1() == Circuit(4).H(2).CX(2, 1).CX(2, 3).CX(1, 0).X(0).X(1).X(
-        2
-    ).X(3)
+    assert t2c.to_tket1() == pytket.Circuit(4).H(2).CX(2, 1).CX(2, 3).CX(1, 0).X(0).X(
+        1
```
acl-cqc: super-nit but this is the worst indenting ever ;)

.github/workflows/ci.yml
```diff
@@ -297,7 +297,7 @@ jobs:
         run: uv python install ${{ env.PYTHON_LOWEST }}
       - name: Setup dependencies
         # avoid building qis-compiler
-        run: uv sync --package=tket --exact --python ${{ env.PYTHON_LOWEST }}
+        run: uv sync --package=tket --exact --all-extras --python ${{ env.PYTHON_LOWEST }}
```
acl-cqc, line 300: Do we need this to run tests? (Ok we do for `mypy` in the one above)

tket-py/tket/_tket/state.pyi
```diff
@@ -1,9 +1,14 @@
...
-from pytket._tket.circuit import Circuit as Tk1Circuit
 
 from tket._tket.ops import TketOp
 from hugr.envelope import EnvelopeConfig
 
+try:
+    from pytket._tket.circuit import Circuit as Tk1Circuit
+except ImportError:
+    # Pytket is only available as an optional dependency under the `pytket` extra.
+    from typing import Any as Tk1Circuit  # type: ignore
```
acl-cqc: This is probably the bit I'm least keen on. So if pytket isn't installed....we have `CompilationState::from_tket1(Any) -> CompilationState`.

Is there a way to instead make the method disappear entirely if pytket is not installed? What about if one instead duplicated the rest of CompilationState? (two versions, one inside the `try` and one inside the `except`) Would that work? Or introduced a superclass which (if pytket is defined) has just that method, and (if not) then is empty??

tket-py/tket/_tket/state.pyi
```diff
@@ -1,9 +1,14 @@
 from typing import Any, Callable
-from pytket._tket.circuit import Circuit as Tk1Circuit
 
 from tket._tket.ops import TketOp
 from hugr.envelope import EnvelopeConfig
 
+try:
+    from pytket._tket.circuit import Circuit as Tk1Circuit
+except ImportError:
+    # Pytket is only available as an optional dependency under the `pytket` extra.
```
acl-cqc:
```suggestion
    # Pytket is installed as an optional dependency under the `pytket` extra, or for testing.
```

tket-py/tket/passes/__init__.py
```diff
@@ -21,6 +19,13 @@
 )
 from hugr.passes.scope import PassScope, GlobalScope
 
+if TYPE_CHECKING:
```
acl-cqc, line 22: hang on, so how come this wasn't inside an `if TYPE_CHECKING` before? Shouldn't `ruff` complain if it's only used for type-checking but isn't inside an `if TYPE_CHECKING`?

tket-py/tket/passes/__init__.py
```diff
@@ -21,6 +19,13 @@
...
+    try:
+        # Available via the `pytket` extra, and as a dev dependency for testing.
+        from pytket.passes import BasePass
+    except ImportError:
+        from typing import Any as BasePass  # type: ignore
```
acl-cqc: Hmmmm. Similar concern as before. If pytket isn't installed, never mind the typechecking, are you gonna be able to use PytketHugrPass? Where would you get your `BasePass`es from?

acl-cqc (commented):
Hmmmm. I guess this is a move in the right direction, as not having to install pytket does seem like a significant win. It does make the typechecking a bit messy/awkward though, hmmm. Wild idea - can you `from typing import Never as Foo` instead?

tket-py/test/test_pauli_prop.py
aborgna-q: 🐍

.github/workflows/ci.yml
aborgna-q: The pytket pass tests require pytket to be installed.

tket-py/tket/_tket/state.pyi
aborgna-q: We don't install this automatically for testing since it's not a dev-dependency anymore.

tket-py/tket/_tket/state.pyi
aborgna-q: This here is a typing file for the rust bindings.

The `.so` will still contain the function if `pytket` is not in the python runtime, so we need to keep some typing hint. (The call in practice is doing duck typing, it will work as long as the input has a `to_dict` method returning valid json). 

In the python-side `tket._state.CompilationState` we currently define
```py
def from_tket1(circ) -> CompilationState
```
Python doesn't really provide nice ways to implement conditional definitions.
The method will work still if pytket is not in the env (and we pass some custom thing), so might as well leave it there (perhaps we should do the same conditional import instead).

I guess we could have a local protocol that defines that ducktyping part explicitly instead.

aborgna-q:
- Added a protocol for pytket circuits and passes, duck-typing the methods that we need when loading them from the rust side (i.e. `to_dict`).

- `CompilationState.to_tket1` is the only call that requires `pytket` to be installed. I moved it to a conditional definition now.
This is only for the user-facing state class. The binding typing file always defines it, since it's reflecting what's in the compiled bindings.

tket-py/tket/passes/__init__.py
aborgna-q: Not sure why 🤷

acl-cqc (approved):
Nice. When you said "duck typing" I was worried what I'd find ;-) but I think the thing you have in `util.py` is not too bad, and makes a much nicer job of things :-). Thanks!

PR #1521, opened 2026-04-16
feat: Support debug info in qis-compiler

cgh-qtnm:
This PR adopts the debug info support being implemented for hugr-llvm in https://github.com/Quantinuum/hugr/pull/3026, by adding an `emit_debug` parameter to the qis-compiler entry points. See the hugr PR for an explanation of the test changes. This PR also updates the sample HUGRs used in `test_basic_generation.py` to include debug info.


cgh-qtnm:
> ## [Codecov](https://app.codecov.io/gh/Quantinuum/tket2/pull/1521?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=Quantinuum) Report
> 
> ❌ Patch coverage is `2.56410%` with `38 lines` in your changes missing coverage. Please review.

I believe these lines are covered by the tests in `qis-compiler/python/tests/test_basic_generation.py`.

Cargo.toml
```diff
@@ -44,10 +44,10 @@ large_enum_variant = "allow"
 [patch.crates-io]
 
 # Uncomment to use unreleased versions of hugr
-# hugr = { git = "https://github.com/quantinuum/hugr", "rev" = "de7b9f7581ae02cd80dff7b12794b55635903304" }
-# hugr-core = { git = "https://github.com/quantinuum/hugr", "rev" = "de7b9f7581ae02cd80dff7b12794b55635903304" }
-# hugr-cli = { git = "https://github.com/quantinuum/hugr", "rev" = "de7b9f7581ae02cd80dff7b12794b55635903304" }
-# hugr-llvm = { git = "https://github.com/quantinuum/hugr", "rev" = "de7b9f7581ae02cd80dff7b12794b55635903304" }
+hugr = { git = "https://github.com/quantinuum/hugr", branch = "george/debuginfo" }
```
cgh-qtnm, line 47: Need to update to a released version of HUGR and remove these patches before merging.

qis-compiler/python/tests/snapshots/test_basic_generation/test_llvm/x86_64-unknown-linux-gnu-print_current_shot/print_current_shot_x86_64-unknown-linux-gnu
```diff
@@ -5,23 +5,50 @@ target triple = "x86_64-unknown-linux-gnu"
...
+!llvm.dbg.cu = !{!1}
+!name = !{!3}
+
+!0 = !{i32 2, !"Debug Info Version", i32 3}
+!1 = distinct !DICompileUnit(language: DW_LANG_Python, file: !2, producer: "guppylang (guppylang-internals-v0.32.0)-v0.21.11", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug)
+!2 = !DIFile(filename: "/Users/george.hodgkins/proj/tket2/qis-compiler/python/tests/generate_hugrs.py", directory: "file:///Users/george.hodgkins/proj/clone-anduril")
+!3 = !{!"mainlib"}
+!4 = distinct !DISubprogram(name: "print_current_shot.<locals>.main", linkageName: "__hugr__.__main__.print_current_shot.<locals>.main.1", scope: null, file: !5, line: 194, type: !6, scopeLine: 195, spFlags: DISPFlagDefinition, unit: !1)
+!5 = !DIFile(filename: "/Users/george.hodgkins/proj/tket2/qis-compiler/python/tests/generate_hugrs.py", directory: "")
+!6 = !DISubroutineType(types: !7)
```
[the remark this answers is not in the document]
cgh-qtnm: This is a good point, not sure it's avoidable with our current frontend generation though. It doesn't affect the current tests because they just use whatever is present in the HUGR (don't actually resolve the path).

qis-compiler/python/tests/generate_hugrs.py
```diff
@@ -1,10 +1,12 @@
 # /// script
 # requires-python = ">=3.10"
 # dependencies = [
-#     "guppylang ==0.21.9",
+#     "guppylang==0.21.9,
 #     "tket",
 # ]
 # ///
+# TODO: point at re-released guppylang(-internals) once available with debug info. right
```
cgh-qtnm, line 9: Need to fix this before merging, once latest guppylang is rereleased.

Cargo.toml
```diff
@@ -44,10 +44,10 @@ large_enum_variant = "allow"
...
+hugr-core = { git = "https://github.com/quantinuum/hugr", branch = "george/debuginfo" }
+hugr-cli = { git = "https://github.com/quantinuum/hugr", branch = "george/debuginfo" }
+hugr-llvm = { git = "https://github.com/quantinuum/hugr", branch = "george/debuginfo" }
```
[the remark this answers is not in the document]
cgh-qtnm: https://github.com/Quantinuum/tket2/pull/1521#discussion_r3112530387

qis-compiler/python/tests/snapshots/test_basic_generation/test_llvm/x86_64-unknown-linux-gnu-print_current_shot/print_current_shot_x86_64-unknown-linux-gnu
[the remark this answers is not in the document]
cgh-qtnm: Looked into it and `clang` also produces DIFile entries with absolute paths, so I think we will keep that approach and live with the consequences.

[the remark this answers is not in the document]
jake-arkinstall: Recommend some anonymisation run on IR text before turning it into a snapshot. Otherwise CI will not be able to reproduce it and the snapshot test is guaranteed to fail, no?

qis-compiler/python/tests/snapshots/test_basic_generation/test_llvm/aarch64-apple-darwin-flip_some/flip_some_aarch64-apple-darwin
```diff
@@ -28,74 +96,12 @@ declare void @panic(i32, ptr) local_unnamed_addr #0
...
-  tail call void @___rxy(i64 %qalloc.i105.i, double 0x400921FB54442D18, double 0.000000e+00)
-  %lazy_measure67.i = tail call i64 @___lazy_measure(i64 %qalloc.i105.i)
-  tail call void @___qfree(i64 %qalloc.i105.i)
-  %read_bool80.i = tail call i1 @___read_future_bool(i64 %lazy_measure67.i)
-  tail call void @___dec_future_refcount(i64 %lazy_measure67.i)
-  tail call void @print_bool(ptr nonnull @res_c3.B223E16D.0, i64 12, i1 %read_bool80.i)
-  %1 = tail call i64 @teardown()
-  ret i64 %1
+  tail call void @setup(i64 %0), !dbg !24
+  tail call void @"__hugr__.__main__.flip_some.<locals>.main.1"(), !dbg !24
```
jake-arkinstall, line 102: Can we switch off tail calls for debug builds?

I don't know how we can recover DWARF metadata if we can't reliably walk the call stack. I guess this could be done manually when compiling bitcode to an object file (-fno-optimize-sibling-calls) but it feels like we shouldn't be hinting it anyway.

qis-compiler/python/tests/snapshots/test_basic_generation/test_llvm/helios-x86_64-unknown-linux-gnu-print_current_shot/print_current_shot_x86_64-unknown-linux-gnu_helios
```diff
@@ -5,23 +5,50 @@ target triple = "x86_64-unknown-linux-gnu"
...
+!llvm.dbg.cu = !{!1}
+!name = !{!3}
+
+!0 = !{i32 2, !"Debug Info Version", i32 3}
+!1 = distinct !DICompileUnit(language: DW_LANG_Python, file: !2, producer: "guppylang (guppylang-internals-v0.32.0)-v0.21.11", isOptimized: false, runtimeVersion: 0, emissionKind: FullDebug)
+!2 = !DIFile(filename: "/Users/george.hodgkins/proj/tket2/qis-compiler/python/tests/generate_hugrs.py", directory: "file:///Users/george.hodgkins/proj/clone-anduril")
+!3 = !{!"mainlib"}
+!4 = distinct !DISubprogram(name: "print_current_shot.<locals>.main", linkageName: "__hugr__.__main__.print_current_shot.<locals>.main.1", scope: null, file: !5, line: 194, type: !6, scopeLine: 195, spFlags: DISPFlagDefinition, unit: !1)
+!5 = !DIFile(filename: "/Users/george.hodgkins/proj/tket2/qis-compiler/python/tests/generate_hugrs.py", directory: "")
+!6 = !DISubroutineType(types: !7)
```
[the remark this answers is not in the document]
cgh-qtnm: The snapshot doesn't fail because this string is present in the input HUGR, and the compiler just copies it into the bitcode without interpreting it. But I will sanitize it for tidiness

qis-compiler/python/tests/snapshots/test_basic_generation/test_llvm/aarch64-apple-darwin-flip_some/flip_some_aarch64-apple-darwin
cgh-qtnm: Looked into this:

1) `tail call` by itself actually [has no effect](https://llvm.org/docs/CodeGenerator.html#tail-call-optimization) unless a) explicitly enabled with `-tailcallopt` or b) the call is immediately followed by a return (this is the case covered by `optimize-sibling-calls`). And even if enabled with `-tailcallopt`, it *still* has no effect unless the call site (or callee function?) is labeled with a non-default calling convention like `fastcc`.

2) Even when compiling a `tail call fastcc` with `llc -tailcallopt`, it seems that a stack frame is still generated (or at least the calling convention uses a normal-looking `bl`-`ret` pair) on Linux and Darwin aarch64, and on Darwin at least the debugger displays a correct backtrace at a call site that was marked with `tail`.

TLDR, if the tail call annotations currently don't propagate to codegen, I think we should just disable them in hugr-llvm and QsystemPass unless/until we want them.

cgh-qtnm: Resolving this conversation: we didn't end up removing tail calls globally because the rest of the team disliked the idea. The user guidance will be that, as with classical programs, optimization may reduce the quality of debug info.

jake-arkinstall: Yuck. But understandable. We can manage it at the compilation stage with flags I guess.

cgh-qtnm:
> This PR contains breaking changes to the public Rust API. 

The breaking changes are from the HUGR upgrade PR, not debug info support.

tket/src/llvm/rotation.rs
```diff
@@ -274,7 +274,7 @@ mod test {
     #[rstest]
     #[case(ConstRotation::new(1.0).unwrap(), ConstRotation::new(0.5).unwrap(), 1.5)]
     #[case(ConstRotation::PI, ConstRotation::new(1.5).unwrap(), 2.5)]
-    fn exec_aadd(
+    fn exec_add(
```
aborgna-q: Typo?

tket-py/pyproject.toml
```diff
@@ -52,6 +52,9 @@ requires = ["maturin~=1.7"]
 build-backend = "maturin"
 
 [dependency-groups]
+dev = [
+    "numpy" # used for tests
+]
```
aborgna-q: Add it to the root's `pyproject.toml` instead.

qis-compiler/rust/lib.rs
```diff
@@ -142,14 +142,18 @@ fn get_hugr_llvm_module<'c, 'hugr, 'a: 'c>(
     hugr: &'hugr Hugr,
     module_name: impl AsRef<str>,
     exts: Rc<CodegenExtsMap<'a, Hugr>>,
-) -> Result<Module<'c>> {
+    emit_debug: bool,
+    ptr_bits: u32,
```
aborgna-q: I added a PR in hugr (https://github.com/Quantinuum/hugr/pull/3098) to give this a better definition.

- In general, bare `bool`s as parameters are confusing in the call site (what does `foo(true)` mean?). In almost all cases is cleaner to define a one-off `Enum` with explicit options.

- `ptr_bits` is only valid here when `emit_debug` is `true`. This is another hint that we should be using an `Enum` instead.

`emit_module` now takes a `EmitDebugInfo` instead.

Python does not support tagged unions, but we can have a `kwargs`-only `bool` instead.

qis-compiler/rust/lib.rs
```diff
@@ -475,12 +507,13 @@ mod selene_hugr_qis_compiler {
 
     /// Compile HUGR package to LLVM IR string
     #[pyfunction]
-    #[pyo3(signature = (pkg_bytes, opt_level=2, target_triple="native", platform="helios"))]
+    #[pyo3(signature = (pkg_bytes, opt_level=2, target_triple="native", platform="helios", emit_debug=false))]
```
aborgna-q: Can we make these kwarg-only?
That includes `compile_to_bitcode` too.

It'll also require an update to the typing hints.


```suggestion
    #[pyo3(signature = (pkg_bytes, *, opt_level=2, target_triple="native", platform="helios", emit_debug=false))]
```

aborgna-q: Note that this is a breaking change, but we're making a breaking release anyways.

cgh-qtnm: No objection as long as @jake-arkinstall approves

cgh-qtnm: I've checked and this doesn't break Selene, so I've applied the change.

aborgna-q (approved):

PR #1524, opened 2026-04-17
feat: InlineFunctionsPass

aborgna-q:
Adds a pass definition that calls the already-defined function inlining tool.

This version uses a simple heuristic to choose when to inline:
- Function must not be on a cycle of the call-graph (i.e. it's not recursive).
- The function definition must have a limited number of nodes (defaults to <= 64 descendants).

We can add better heuristicts in the future, and check for inline hints.
This is a first version to help with simple optimization cases.

Note that the inlined funcDefs are not removed. `RemoveDeadFuncsPass` can be called afterwards to remove what's not needed.

We should also add this as a first step in `NormalizeGuppy`


tket/src/passes.rs
```diff
@@ -43,6 +43,8 @@ pub use inline_dfgs::InlineDFGsPass;
 
 // Inline function calls.
 pub mod inline_funcs;
+pub use inline_funcs::InlineFunctionsPass;
```
cqc-alec, line 46: Maybe this should be called something like `InlineSmallFunctionsPass`, since the name suggests that it inlines all functions.

tket-py/tket/_tket/passes.pyi
```diff
@@ -42,6 +42,14 @@ def normalize_guppy(
     - remove_redundant_order_edges: Whether to remove redundant order edges.
     """
 
+def inline_functions(
```
cqc-alec, line 46: Maybe rename to `inline_small_functions`? Alternatively we could have `max_inline_size` be optional and if not set inline all functions?

tket-py/test/test_pass.py
```diff
@@ -240,3 +241,11 @@ def test_modifier_resolver() -> None:
     resolved: Hugr = mr_pass(normalized)
 
     assert _count_ops(resolved, "tket.modifier.ControlModifier") == 0
+
+
+def test_inline_functions() -> None:
```
cqc-alec, line 247: This is an example where the pass does nothing (I think because there are no function calls); can we add a test where the pass does something (and also where it does nothing because the function was too large)?

tket/src/passes/inline_funcs.rs
```diff
@@ -1,30 +1,115 @@
...
+    fn with_scope(mut self, scope: impl Into<PassScope>) -> Self {
+        self.scope = scope.into();
+        self
+    }
+}
+
 /// Inline (a subset of) [Call]s whose target [FuncDefn]s are not in cycles of the call
 /// graph.
 ///
+/// Processes any calls nodes that are descendants of the entrypoint.
```
cqc-alec:
```suggestion
/// Processes any call nodes that are descendants of the entrypoint.
```

cqc-alec (commented):
Thanks, just small comments.

tket-py/test/test_pass.py
aborgna-q: The `fn_calls` example  should probably have some calls there :P
I updated the generator and made test check that we're inlining things.

tket-py/tket/_tket/passes.pyi
aborgna-q: I'd like to have the default behaviour be useful. I don't think "inline everything" is something users will want to do most of the time.

See my comment about a Heuristics enum below. I'll add an `All` variant.

tket/src/passes.rs
aborgna-q: I don't think we want to limit ourselves to only small functions here.

The `max_size` limit is the simplest of the heuristics we could apply.

We may want to add alternatives in the future that decide based on the function types or something else.

We'll also want to add the option to inline functions based on `inline` hints in their metadata.

I think I'll add an `InlineHeuristic` enum instead of the `max_size` flag.

Do you have an alternative name suggestion that also works for different heuristics?

cqc-alec: Sounds good to me. I think the name is fine if it is referring to a general heuristic.

aborgna-q:
Added the heuristic enum, and a flag for following inline hints.
(The inline hint metadata will come in a separate PR, but we can at least stabilize the pass API now).

I'm using `Protocol`s for the python heuristics, since python enums can't carry data. Not sure if I love this design, but it works.

cqc-alec (approved):
LGTM

PR #1525, opened 2026-04-17
chore: Fix rust 1.95 lints

aborgna-q:
Run `clippy --fix` with the latest release


devenv.lock
ss2165, line 1: ...surprising diff

ss2165 (approved):

devenv.lock
aborgna-q: drive-by: `just check` fixed the missing EOL

PR #1530, opened 2026-04-21
feat: HUGR extension for global variables

maximilianruesch:
A HUGR extension for setting global variables. Two operations are defined: 
- The `with` operation that acts as a context manager to set the value for the global state, run a function, and return the resulting value of the global state.
- The `map` operation that maps a function over the contents of the named global variable.


tket-qsystem/src/extension/globals.rs
```diff
@@ -0,0 +1,361 @@
+#![allow(missing_docs)]
+
```
doug-q:
```suggestion
```

tket-qsystem/src/extension/globals.rs
```diff
@@ -0,0 +1,361 @@
...
+        EXTENSION_ID
+    }
+
+    fn description(&self) -> String {
+        match self {
+            Self::with => {
+                "Run a function with the set contents of the global variable.".to_string()
+            }
+            Self::map => {
+                "Map a function over the contents of the named global variable.".to_string()
```
doug-q, line 137: Can you document what the type params mean here ?

tket-qsystem/src/extension/globals.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub static ref NAME_PARAM: TypeParam = TypeParam::StringType;
+    pub static ref TYPE_PARAM: TypeParam = TypeParam::RuntimeType(TypeBound::Linear);
+
+    /// The [TypeParam] of various types and ops specifying the input signature of a function.
+    pub static ref INPUTS_PARAM: TypeParam = TypeParam::ListType(Box::new(TypeBound::Linear.into()));
+    /// The [TypeParam] of various types and ops specifying the explicit output signature of a function.
+    pub static ref OUTPUTS_PARAM: TypeParam = TypeParam::ListType(Box::new(TypeBound::Linear.into()));
+
+    /// The [TypeParam] of various types and ops specifying the implicit output signature of a function.
+    pub static ref IMPL_OUTPUTS_PARAM: TypeParam = TypeParam::ListType(Box::new(TypeBound::Linear.into()));
```
doug-q: Not sure what this is, I suspect that this "implicit"ness comes from guppy, it does not make sense to talk about it at the hugr level

hsemenenko: Yes, this is implicit returns from Guppy functions. In order to construct the HUGR op signature correctly, so that it matches the signature from Guppy, we need to pass the explicit and implicit returns separately. Implicit types are appended to the end of the output types. In our case, the global state is the first argument and the ordering of the return types depends on whether the other inputs to the function are linear or not. We cannot check whether they are linear in HUGR, so we have to check in Guppy and pass the correct arguments to instantiate the op. 

`(global_state, non_linear_input) -> Out` becomes `(global_state, non_linear_input) -> [Out, global_state]`

However,

`(global_state, linear_input) -> Out` becomes `(global_state, linear_input) -> [Out, global_state, linear_input]`

The ordering of the outputs is inconsistent so when we lower to LLVM, we cannot be certain to where the global state is in the function return.

We instead always have the `global_state` as the first argument, and construct the function signature from the explicit and implicit return args.

tket-qsystem/src/extension/globals.rs
hsemenenko: This seems inconsistent with other extensions that have one-line descriptions.

doug-q (approved):
Looks great! Thank you

PR #1531, opened 2026-04-21
fix: Apply modifiers through function calls

### LINKED ISSUE #1519: [Bug]: Modifier are not applied to function
### Description

When we apply the modifier to a function call, e.g.:
```
@guppy
def bar(q: qubit) -> None:
    x(q)

@guppy
def foo() -> None:
    q1 = qubit()
    q2 = qubit()
    with control(q1):
        fuu(q2)
```
the modifier is not applied to the `x` in `bar`

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
close #1519 

The bug was due to an error in the metadata checking. We were checking if the function had the metadata "tket.unitary" instead of just "unitary", causing the pass to consider classical some quantum function, and thus not forwarding the modifier inside these functions.

I added a new set of example programs in `test_files/modifier_examples` that generate hugrs from guppy functions with modifiers. When running `just recompile-test-hugrs`, we generate the hugr, apply the passes on these new hugr thanks to `test_files/run_modifier_examples/apply_passes.py`, and finally run the programs and save a snapshot of the results (the results are also saved in `hugr_results.txt` in a readable format).

A new test function in `tket-py/test/test_pass.py` uses the snapshot to check that the passes solve the modifier properly, producing the right results.

A lot of renaming has been done for clarity


scripts/run_modifier_examples.sh
```diff
@@ -0,0 +1,23 @@
...
+    cd "$repo_root/test_files/modifier_examples"
+    just r
+)
+
+echo "==> Running Rust test test_saved_hugr"
+(
+    cd "$repo_root"
+    cargo test -p tket test_saved_hugr
+)
+
```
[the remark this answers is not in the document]
cqc-alec: Indeed this test does not exist.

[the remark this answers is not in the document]
cqc-alec: Where is this script invoked from?

test_files/modifier_examples/justfile
```diff
@@ -0,0 +1,19 @@
+# List the available commands
+help:
+    @just --list --justfile {{justfile()}}
+
+# Re-generate all hugr files and their mermaid diagrams.
```
cqc-alec: I don't see where mermaid diagrams are generated?

test_files/run_modifier_examples/hugr_results/ctrl_on_call2_solved.npy
cqc-alec, line 1: After running just recompile-modifiers I found this file and a few others were modified. Maybe we need to pin a numpy version?

test_files/run_modifier_examples/apply_passes.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+mr_pass = ModifierResolverPass()
+modifier_examples_dir = Path(__file__).resolve().parents[1] / "modifier_examples"
+modified_hugrs_dir = Path(__file__).resolve().parents[1] / "modified_hugrs"
+modified_hugrs_dir.mkdir(parents=True, exist_ok=True)
+
+
+input_paths = (
+    # todo: the file as input must be found in modifier_examples_dir, not in the current directory
```
cqc-alec: Is this an old TODO, looks like it is correct?

test_files/run_modifier_examples/run_hugrs.py
```diff
@@ -0,0 +1,81 @@
...
+
+        000 -> 0.7071+0.j, 111 -> 0.7071+0.j
+    """
+    n_qubits = int(np.round(np.log2(len(state))))
+    parts = []
+    for idx, amp in enumerate(state):
+        if abs(amp) > threshold:
+            label = format(idx, f"0{n_qubits}b")
+            parts.append(f"\t{label} -> {amp:.4g}")
+    return "\n".join(parts) if parts else "(all zero)"
```
cqc-alec: All zero shouldn't be possible (except for very large numbers of qubits where everything is below the threshold, in which case we should probably say "all amplitudes close to zero" or something).

test_files/utility.py
```diff
@@ -0,0 +1,6 @@
+from pathlib import Path
+
+
+hugr_pdf_directory = "../0_hugr_pdf"
+
+Path(hugr_pdf_directory).mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
cqc-alec: Did you intend to check this in?

tket-py/test/test_pass.py
```diff
@@ -238,10 +242,46 @@ def test_modifier_resolver() -> None:
...
+                    "run_hugrs.py",
+                    hugr_name,
+                    tmp_path,
+                ],
+                cwd=run_hugrs_dir,
+                check=True,
+            )
+
+            computed_statevector = np.load(tmp_path)
+            print(computed_statevector)
```
cqc-alec: Remove this `print()`.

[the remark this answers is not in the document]
cqc-alec: Seems reasonable. It does feel quite hacky to be calling `uv run` to execute Python from a subprocess within Python rather than calling Python directly. I suppose this is because we don't want to add the extra dependencies to the tests? Should we reconsider that?

tket/src/modifier/modifier_resolver.rs
```diff
@@ -981,28 +985,29 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+        } else if self.modify_array_op(h, op_node, optype, new_dfg)?
+            || self.try_array_convert(h, op_node, optype, new_dfg)?
         {
             Ok(())
         } else {
             // Some other Hugr extension operation.
             // Here, we do not know what is the modified version.
             // We try to place the original operation.
-            self.modify_dataflow_op(h, n, optype, new_dfg)
+            // NICOLA: Not sure, if we dont know what to do maybe an error is better?
```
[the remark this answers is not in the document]
cqc-alec: Agree.

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1131,18 +1148,23 @@ pub fn resolve_modifier_with_entrypoints(
...
-    // Ad hoc cleanup procedure.
+    // Ad hoc cleanup procedure: remove any dangling global-phase nodes that
+    // were produced or left behind by the resolution passes above.
     delete_phase(h, entry_points)?;
 
+    // --- Phase 4: Validation ---
+    // Confirm that the resulting hugr is still structurally valid after all rewrites.
     h.validate()
         .map_err(|e| ModifierResolverErrors::BuildError(e.into()))?;
+    // h.extensions_mut() = original_extensions;
```
cqc-alec: Delete commented line.

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1220,40 +1258,49 @@ mod tests {
...
+
             if dagger {
+                // Wrap with Dagger before Control.
                 call = func
                     .add_dataflow_op(dagger_op, vec![call])
                     .unwrap()
                     .out_wire(0);
             }
+
+            // Wrap the with the Control modifier.
```
cqc-alec:
```suggestion
            // Wrap with the Control modifier.
```

test_files/run_modifier_examples/hugr_results/ctrl_on_call2_solved.npy
nicolaassolini-qntm: numpy is installed by guppy, not sure if fixing the verison is a good idea

nicolaassolini-qntm:
`test_files/utility.py` and `scripts/run_modifier_examples.sh` were old files, I removed them. 

I also merged the main, in doing so, I uploaded the guppy version to .13 in all the tests.

test_files/run_modifier_examples/hugr_results/ctrl_on_call2_solved.npy
cqc-alec: Hmm OK. Wonder why I was getting different results locally.

nicolaassolini-qntm: do you still have that issue? Now I also see that a CI test is failing, maybe because I'm not using `uv` to run the hugr

tket-py/test/test_pass.py
```diff
@@ -238,10 +242,47 @@ def test_modifier_resolver() -> None:
...
+
+    for hugr_path in sorted(modified_hugrs_dir.glob("*.hugr")):
+        hugr_name = hugr_path.stem.removesuffix("_solved")
+        expected_statevector = expected_results[hugr_path.stem]
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            tmp_path = Path(tmp_dir) / f"{hugr_name}.npy"
+            subprocess.run(
+                [
+                    "uv",
```
cqc-alec, line 271: If we made `guppylang` a dev dependency of tket we could presumably avoid subprocess calls here, but we can leave it for now.

cqc-alec (approved):
Something seems to have gone wrong with uploading the coverage reports.

PR #1532, opened 2026-04-21
feat: InlineHint metadata

aborgna-q:
Follow up to Quantinuum/hugr#1524. Defines a metadata value for hinting the compiler on which functions to inline.

Part of https://github.com/Quantinuum/tket2/issues/1804.

Note that this definition is rust-only. We still need to add exports for all tket metadata definitions on the python bindings.

Guppy issue/PR (they'll need to be updated):
- https://github.com/Quantinuum/guppylang/pull/1513
- https://github.com/Quantinuum/guppylang/issues/1054


acl-cqc (commented):
Ok let's combine functionality (not necessarily PRs) here with #1449.
I've updated that to use a similar metadata (there without the `Never`, with `#[non_exhaustive]`, and I called it InlineAnnotation rather than hint, but at least it's in the same place ;-)).
Noting that #1524 uses `inline_acyclic` which will mean any function on a cycle will not be inlined at all (cycles are removed from graph rather than merely broken), I think #1449 handles the "Always" bit better, so that can be run first as a separate pass.

Thus, either (a) put this in first, but without the Always option (to be added in 1449); or (b) let's get #1449 in first and then this can add the Never option. I don't really mind which way around, and I think #1449 can be marked ready now (probably for a random reviewer? But you might want to check over the annotation).

WRT this PR:
* I am also unsure about "follow_hints=False" still respecting the "Never" hint; are we sure it shouldn't follow that too? (Otherwise, follow_hints will do nothing when Always is removed. One could say "Always is not a hint" but an annotation, iow it's correct for "follow_hints" to have no effect until we add some InlineAnnotation::XYZHint).
* Would be nice to add to the heuristic here that (non-preserved?) functions with only one call "are always small enough" ;)
* I haven't checked what you're doing here, but I note it's quite tricksy doing inlining respecting PassScope::EntrypointXXX :(

acl-cqc:
If we'd rather put this in with the algorithm as here, which I think is ok....
* Do we want to combine metadata for this pass and Quantinuum/hugr#1449?
   * If so, I think it should be "InlineAnnotation" and the *variants* that are hints should be called XyzHint, or something like that. (Does it live in `tket.hint....` ?) e.g. your "always" is a *hint*, not a command.
   * Otherwise, we could have both InlineAnnotation and InlineHint....
* "Always" would be better as "BestEffort" or "Best" or "Max" or something like that, with actually "always" handled by Quantinuum/hugr#1449. (Really "Always" here means "Trust me I am small enough")
* "follow"ing hints (default true) would be better as "ignore hints" (default false).
   * Should probably apply uniformly to both never and (pseudo-)always?

aborgna-q:
- Renamed `metadata::InlineHint` to `metadata::InlineAnnotation`. It's now `#[non_exhaustive]`.
- Renamed the metadata key from `tket.hint.inline` to just `tket.inline`.
- Removed `InlineAnnotation::Always` and added a `BestEffort` (encoded as `best_effort`).
- Removed the `follow_inline_hints` flag from the inline funcs pass. It's less confusing now.

I think that should combine well with your PR now?

tket-py/tket/metadata.py
```diff
@@ -72,6 +75,41 @@ class MaxQubitsHint(Metadata[int]):
...
 
+class InlineAnnotationValue(Enum):
+    """Metadata value hinting the compiler that a function declaration should be inlined at its call sites.
+
+    When a function is not annotated, we use a heuristic to determine whether to inline.
+
+    Values:
+    - "never": Never inline this function.
+    - "best_effort":
+        Inline the function if we know it won't produce an invalid Hugr.
```
acl-cqc:
```suggestion
        Inline the function if possible
 ```
 or "as much as the compiler is able."

I don't know that infinitely big Hugrs are invalid. I mean, we can't validate them, but if we had infinite memory and address space, validation won't throw an error... ;)

File: tket-py/tket/metadata.py [hunk 2]
```diff
@@ -72,6 +75,41 @@ class MaxQubitsHint(Metadata[int]):
...
+        This is a best effort option; the compiler may choose not to inline
```
acl-cqc:
```suggestion
        Whilst this is no guarantees (the compiler may choose not to inline
        functions with this annotation), the compiler should aim to inline whenever possible, ignoring any restrictions due to finite quantities
```

Or "ignoring quantitative restrictions" or something like that

tket-py/tket/metadata.py
```diff
@@ -72,6 +75,41 @@ class MaxQubitsHint(Metadata[int]):
     KEY = _metadata.MAX_QUBITS_HINT
 
 
+class InlineAnnotationValue(Enum):
```
acl-cqc: Just wondering if it's possible to use a `Literal["never","best_effort"]` instead of this Enum? That might be much simpler?

acl-cqc (approved):
Thanks Agustin, looks good to me

tket-py/tket/metadata.py
aborgna-q: I'd be cautious of promising what the compiler will do here.
We don't want to be limited by the metadata restrictions when updating the compiler inlining pass.

tket-py/tket/metadata.py
aborgna-q: Yeah, for a public API it may be simpler to do strings. Updated.

PR #1535, opened 2026-04-21
feat: Add python definitions for the TKET metadata keys

aborgna-q:
We should also add a metadata for inline hints either here or in https://github.com/Quantinuum/tket2/pull/1532, whichever gets merged second.


tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    KEY = _metadata.MAX_QUBITS
+
+
+class CircuitRewriteTraces(Metadata[list[RewriteTraceValue]]):
+    """Metadata key for rewrite traces recorded during circuit transformation."""
+
+    KEY = _metadata.CIRCUIT_REWRITE_TRACES
+
+
+class Unitary(Metadata[int]):
```
acl-cqc: Maybe UnitaryFlags? It's definitely not a unitary (i.e. matrix) itself

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    KEY = _metadata.UNITARY
+
+
+class InputParameters(Metadata[list[str]]):
+    """Metadata key for explicit names of input parameter wires."""
+
+    KEY = _metadata.INPUT_PARAMETERS
+
+
+class OpGroup(Metadata[str]):
```
acl-cqc: Perhaps `PytketOpGroup`? And elsewhere in this file for other metadata relating to pytket, but there are some that are not

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+from typing import TYPE_CHECKING, TypeAlias, TypedDict
+
+from hugr.metadata import Metadata
+from ._tket import metadata as _metadata
+
+if TYPE_CHECKING:
+    from hugr.utils import JsonType
+
+
+__all__ = [
```
acl-cqc, line 38: Did you mean to exclude `PytketBit` and `PytketQubit`? (Are those used outside of metadata? If so, I wonder if there's somewhere else they could go?)

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+
+
+class Phase(Metadata[str]):
+    """Metadata key for the serialized TKET1 global phase expression."""
+
+    KEY = _metadata.PHASE
+
+
+def _store_pytket_register(value: list[tuple[str, list[int]]]) -> JsonType:
+    return [[name, indices] for name, indices in value]  # type: ignore
```
acl-cqc: Why the `type: ignore`? Can/should you be more specific? It looks reasonable to me from my reading of `JsonType` (you remove the tuple)

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+            raise TypeError(
+                f"Expected {key} register name to be a string, but got {type(name)}"
+            )
+        if not isinstance(indices, list) or not all(
+            isinstance(index, int) for index in indices
+        ):
+            raise TypeError(
+                f"Expected {key} register indices to be a list of integers, but got {indices!r}"
+            )
+        registers.append((name, indices))  # type: ignore
```
acl-cqc: you probably could remove this `type: ignore` by iterating through `indices` and building a new list (I had a similar case recently), up to you

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    KEY = _metadata.INPUT_PARAMETERS
+
+
+class OpGroup(Metadata[str]):
+    """Metadata key for the pytket ``opgroup`` field on a decoded operation."""
+
+    KEY = _metadata.OP_GROUP
+
+
+class BitRegisters(Metadata[list[PytketBit]]):
```
acl-cqc: `PytketBitRegisterNames`?

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    @classmethod
+    def to_json(cls, value: list[PytketQubit]) -> JsonType:
+        return _store_pytket_register(value)
+
+    @classmethod
+    def from_json(cls, value: JsonType) -> list[PytketQubit]:
+        return _read_pytket_register(cls.KEY, value)
+
+
+class Phase(Metadata[str]):
```
acl-cqc: `PytketPhaseExpr`?

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+PytketBit: TypeAlias = tuple[str, list[int]]
+
+
+class RewriteTraceValue(TypedDict):
+    """Serialized rewrite trace metadata entry."""
+
+    individual_matches: int
+
+
+class MaxQubits(Metadata[int]):
```
acl-cqc: Should this be `MaxQubitsHint`?

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    >>> node.metadata[MaxQubits] = 3
+    >>> node.metadata[InputParameters] = ["theta", "phi"]
+    >>> node.metadata[QubitRegisters] = [("q", [0]), ("ancilla", [1])]
+    >>> node.metadata[Phase] = "1/2"
+    >>> node.metadata[MaxQubits]
+    3
+    >>> node.metadata.get(QubitRegisters)
+    [('q', [0]), ('ancilla', [1])]
+"""
+# Changes to this file **MUST** be reflected in `tket/src/metadata.rs`
```
acl-cqc: In the sense that they ought to match, but this doesn't mean that things will break if they don't match, right? Isn't that SHOULD ?

acl-cqc: I guess this means that all of my suggested renames should be applied on the rust side too, heh

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    "QubitRegisters",
+    "Phase",
+]
+
+
+# Identifier for a TKET1 qubit register element.
+#
+# This can be passed to `pytket.unit_id.Qubit.from_list`
+PytketQubit: TypeAlias = tuple[str, list[int]]
+# Identifier for a TKET1 bit register element.
```
acl-cqc: newline

tket-py/tket/metadata.py
```diff
@@ -0,0 +1,151 @@
...
+    "Unitary",
+    "InputParameters",
+    "OpGroup",
+    "BitRegisters",
+    "QubitRegisters",
+    "Phase",
+]
+
+
+# Identifier for a TKET1 qubit register element.
```
acl-cqc: most places in this file use `pytket`, but there are a few `TKET1`. Should we make them consistent?

acl-cqc (approved):
In #1449 I found myself wondering, does this not exist...so yes thank you :-) 👍.

Mostly just naming suggestions, nothing major, looks good!

tket-py/tket/metadata.py
aborgna-q: Mypy used to not like the recursive type. I guess it got fixed ?

tket-py/tket/metadata.py
aborgna-q: Changed to a SHOULD. I'll do the renaming on both rust and python (with some aliases for backwards compatibility)

PR #1546, opened 2026-04-27
fix!: Passes with Nested Modifier

### LINKED ISSUE #1518: [Bug]: Classical Function Inside Modifier
### Description

We got the following error: 
`PyModifierResolverError: Unreachable error: Cannot connect the wires with the same direction: DirWire(Node(5), In(0)) -> DirWire(Node(3), In(0))`
If a classical function is called under a modifier

### Steps to Reproduce

```
@guppy
def fuu(i: int) -> int:
    return i + 1

@guppy
def foo() -> None:
    q1 = qubit()
    q2 = qubit()
    h(q1)
    with dagger:
        bar(q2)
        fuu(2)
    result("q1", measure(q1))
    result("q2", measure(q2))

hugr = foo.compile().modules[0]
```

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

### LINKED ISSUE #1520: [Bug]: Error with function taking classical arguments under nested modifier
### Description

the following code:
```
def foo() -> None:
    q1 = qubit()
    q2 = qubit()

    with dagger, control(q1):
        rx(q2, angle(1/3))
    
    result("q1", measure(q1))
    result("q2", measure(q2))
```

raise: 
`PyModifierResolverError: Unreachable error: Cannot connect the wires with the same direction: DirWire(Node(11), Out(1)) -> DirWire(Node(14), Out(0))` due to the presence of the argument `angle(1/3)`.

probably related to #1518 

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1518 closes #1520

**Bug 1**
The condition in [`modify_fn_if_needed`](https://github.com/Quantinuum/tket2/blob/08d61c43f40a346998435cb5442487ceb1006039/tket/src/modifier/modifier_resolver/dfg_modify.rs#L302) was incorrect. As a result, the pass solver attempted to process classical functions, leading to an invalid inversion of non-quantum input ordering.

**Bug 2**
There was a misalignment between metadata-derived values ([`from_metadata`](https://github.com/Quantinuum/tket2/blob/08d61c43f40a346998435cb5442487ceb1006039/tket/src/modifier.rs#L64-L65)) and the corresponding function representation. This inconsistency became visible with nested modifiers, where metadata mismatches caused the modifier builder to skip the modified block, ultimately triggering errors.

**Bug 3**
In [modify_cfg](https://github.com/Quantinuum/tket2/blob/1c8bc20ee5efab15f4df5cb53c8da821b6dc7477/tket/src/modifier/modifier_resolver.rs#L1022), wire_node_inout was called with the modified signature types. The modified signature has extra control qubits prepended. Since wire_node_inout iterates the type list while incrementing old port indices starting from 0, it was misaligned producing two Out ports that can't be connected.

**Fix:**
* The check in `modify_fn_if_needed` has been corrected to properly exclude classical functions.
* Metadata handling has been aligned with function values (via updates introduced in this [guppy PR](https://github.com/Quantinuum/guppylang/pull/1691)), ensuring consistency between node values and their expected representation at the tket level.
* At [`modifier_resolver.rs#L1069`](https://github.com/Quantinuum/tket2/blob/1c8bc20ee5efab15f4df5cb53c8da821b6dc7477/tket/src/modifier/modifier_resolver.rs#L1069) we use the original signature instead of the modified one since the offset already contains the information about the added qubits.
* 

Additional tests have been introduced to improve coverage of modifier handling and HUGR construction.

* The testing pipeline has been partially refactored to avoid persisting intermediate modified HUGRs, simplifying the workflow and reducing unnecessary state.
* The [`[test_pass](https://github.com/Quantinuum/tket2/blob/52d10e19a91f46825c3eaf38f7c65e5f5a80bf81/tket-py/test/test_pass.py#L254)`](https://github.com/Quantinuum/tket2/blob/52d10e19a91f46825c3eaf38f7c65e5f5a80bf81/tket-py/test/test_pass.py#L254) test has been extended: it now validates not only execution but also correct resolution of modifiers, increasing confidence in the end-to-end behavior.

BREAKING CHANGE: In `ModifierResolver::modify_fn_if_needed` the parameter `signature` has been removed
BREAKING CHANGE: In `control.rs`, `dagger.rs`, and `power.rs`, the modifier constructors `new()` became private, and `signature()` was reduced from `pub` to `pub(crate)`.
BREAKING CHANGE: In `modifier_resolver.rs`, several internal helpers were hidden or narrowed:
BREAKING CHANGE: `DirWire::new`, `ModifierResolver::new`, `ModifierResolverErrors::{unreachable, unresolvable}`, `connect_all`, and `modify_signature` are now private.
BREAKING CHANGE: `DirWire::reverse` is now `pub(crate)`
BREAKING CHANGE: In resolver submodules: `modify_fn_if_needed`, `modify_fn`, `modify_global_phase`, `delete_phase`, and `modify_tket_op` were changed from `pub` to `pub(crate)`.

Not a python breaking release
BEGIN_COMMIT_OVERRIDE
fix: Passes with Nested Modifier (#1546)
END_COMMIT_OVERRIDE


tket/src/modifier/modifier_resolver.rs
```diff
@@ -680,23 +682,22 @@ impl<N: HugrNode> ModifierResolver<N> {
         let new_load = self.with_modifiers(modifiers, |this| {
             this.apply_modifier_chain_to_loaded_fn(hugr, modifier_node)
         })?;
-
+        // NICOLA: the fail is before here!
```
cqc-alec: Should this comment be deleted? Otherwise, needs explanation!

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -356,6 +356,7 @@ impl<N: HugrNode> ModifierResolver<N> {
         Ok(new_function_node)
     }
 
+    /// NICOLA: seemes to be useless
```
cqc-alec: If it's useless can we remove it?

tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -293,8 +299,49 @@ mod tests {
             .call(callee.handle(), &[], vec![inputs[0]])
             .unwrap()
             .out_wire(0);
+        // inputs[0] = func
+        //     .add_dataflow_op(TketOp::X, vec![inputs[0]])
+        //     .unwrap()
+        //     .out_wire(0);
```
cqc-alec: Delete commented code?

cqc-alec (approved):
Thanks, just some minor comments.

nicolaassolini-qntm:
> Thanks, just some minor comments.

I forgot to delete these comments. I will delete it!

tket/src/modifier/modifier_resolver/dfg_modify.rs
nicolaassolini-qntm: old comment, the function is needed

PR #1550, opened 2026-04-28
fix: bug in hugr building when `control` modifiers are nested

### LINKED ISSUE #1537: [Bug]: Invalid hugr with multiple nested controller
### Description

When we have a multiple controller modifier nested with other controllers, we produce an invalid hugr.

### Steps to Reproduce

given the hugr from the program: 
```python
@guppy
def main() -> None:
    t = qubit()
    c1 = qubit()
    c2 = qubit()
    c3 = qubit()
    rx(c1, angle(1 / 3))
    h(c2)
    x(c3)
    with control(c1, c2):
        with control(c3):
            # with dagger:
            h(t)
            rz(t, angle(1 / 2))
```
we obtain the error: `tket.PyModifierResolverError: The constructed HUGR is invalid: Connected ports Port(Outgoing, 0) in Node(100) and Port(Incoming, 0) in Node(102) have incompatible kinds. Cannot connect array(1, qubit) to array(2, qubit)..`

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
close #1537 

The bug was in [`modify_signature`](https://github.com/Quantinuum/tket2/blob/07a0fdbed9454b17faf60707fb7623e0b71eb5a6/tket/src/modifier/modifier_resolver.rs#L703): it prepended each accumulated control array with repeated insert(0, ...), which reversed nested control groups. For accumulated controls like [2, 1], the function signature became [array(1), array(2), ...] while the caller and generated unpack nodes expected [array(2), array(1), ...]. 

I changed it to build the full control prefix once and splice it at the front, preserving accumulation order.

Deleted old `ctrl_on_call` test outputs (.hugr, .npy)


doug-q (approved):
Nice. Thank you.

PR #1556, opened 2026-04-29
feat: expose the QSystemPass to Python

### LINKED ISSUE #1553: [Feature]: QSystemPass should be exposed to Python as a tket `ComposablePass`
### Problem Statement

The QSystemPass currently only has a rust defininition (see `tket.qsystem` in `lib.rs`) 

We should expose this to Python so users can perform a rebase manually on their guppy programs (converting from quantum to qsystem ops.)

The pass should also implement the `ComposablePass` trait just like `NormalizeGuppy` and `PytketHugrPass`

CalMacCQ:
closes #1553 

Driveby: now ncludes a fix to how we resolve extensions in Python. See [076c176](https://github.com/Quantinuum/tket2/pull/1556/commits/076c1761220fab4d067d09ed17c454259af802c6). Prior to this fix the transformed HUGR contained Custom ops rather than QSystem ops as we would expect.

Another driveby: There was a Boolean flag called `hide_funcs` which did not have a corresponding `with_*` method in Rust. I assume this was an oversight so I added a method in [739870d](https://github.com/Quantinuum/tket2/pull/1556/commits/739870dc8167dcb5ca911346a044d51c0e354b5a)[739870d](https://github.com/Quantinuum/tket2/pull/1556/commits/739870dc8167dcb5ca911346a044d51c0e354b5a).


tket-py/src/passes.rs
```diff
@@ -203,3 +212,24 @@ fn resolve_modifiers(circ: &mut CompilationState, scope: Option<PyPassScope>) ->
...
+fn qsystem_rebase_pass(
+    circ: &mut CompilationState,
+    constant_fold: bool,
+    monomorphize: bool,
+    force_order: bool,
+    lazify: bool,
+    scope: Option<PyPassScope>,
+) -> PyResult<()> {
+    let py_scope = scope.unwrap_or_default();
+    let qsystem_pass = tket_qsystem::QSystemPass::default_with_scope(py_scope.scope)
```
CalMacCQ, line 228: One flag that's missing here is `hide_funcs` which is in the `QSystemPass` rust struct. We should also expose this to Python right? There is no corresponding "with" method that I'm aware of.

tket-py/tket/passes/__init__.py
```diff
@@ -294,3 +294,49 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
             scope=self._scope,
         )
         return program
+
+
+@dataclass
+class QSystemPass(ComposablePass):
+    """A pass to convert quantum ops to qsystem ops."""
+
```
CalMacCQ, line 311: TODO: Document the boolean flags here

tket-py/tket/_tket/passes.pyi
```diff
@@ -99,3 +99,13 @@ def resolve_modifiers(
...
+
+def qsystem_rebase_pass(
+    circ: CompilationState,
+    constant_fold: bool = True,
+    monomorphize: bool = True,
+    force_order: bool = True,
+    lazify: bool = True,
+    scope: PassScope | None = None,
+) -> None:
+    """Runs a rust backed pass to convert quantum ops to qsystem ops."""
```
CalMacCQ: Docs needed here

tket-py/test/test_pass.py
```diff
@@ -297,3 +297,12 @@ def test_inline_functions() -> None:
     all = InlineFunctions(heuristic=inline_funcs.All())(hugr)
 
     assert _count_ops(all, "Call") == 0
+
+
+def test_python_qsystem_pass() -> None:
+    circ = Circuit(2).H(0).Rz(0.75, 0).CX(0, 1)
```
CalMacCQ: A better test would be to load in a guppy generated hugr, apply the qsystem pass and then assert that no `std.quantum` ops are present.

Will do that tomorrow.

tket-py/tket/passes/__init__.py
```diff
@@ -294,3 +295,56 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
             scope=self._scope,
         )
         return program
+
+
+@dataclass
```
aborgna-q: Since all parameters are boolean, let's mark them as `kw_only` to avoid someone calling `QSystemPass(True, False, False, True)`.


```suggestion
@dataclass(kw_only=True)
```

tket-py/src/passes.rs
aborgna-q: That looks like an unintended omission.

Could you add a `with_hide_funcs` method, and propagate it here?

CalMacCQ: added in [739870d](https://github.com/Quantinuum/tket2/pull/1556/commits/739870dc8167dcb5ca911346a044d51c0e354b5a)[739870d](https://github.com/Quantinuum/tket2/pull/1556/commits/739870dc8167dcb5ca911346a044d51c0e354b5a). Do the docs make sense?

tket-py/tket/passes/__init__.py
CalMacCQ: done in [a20b808](https://github.com/Quantinuum/tket2/pull/1556/commits/a20b808bda907e87d3b45c2c1075f77bbe82146e)[a20b808](https://github.com/Quantinuum/tket2/pull/1556/commits/a20b808bda907e87d3b45c2c1075f77bbe82146e)

tket-py/src/passes.rs
aborgna-q: LGTM

aborgna-q (approved):
🚀

PR #1557, opened 2026-04-29
feat: Clean Up by Deleting Unconnected Nodes after Modifier Solve

### LINKED ISSUE #1549: [Feature]: Removing old function blocks after Modifier resolving
### Problem Statement

We resolve the modifier, we "transform" the functions. However, the old functions are still in the hugr. (see Node(31) in the attached file)

[ctrl_on_x_solved.pdf](https://github.com/user-attachments/files/27159456/ctrl_on_x_solved.pdf)

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1549

When a modifier is resolved on a function, a new function node is created, leaving an unreachable function node in the hugr. This Pr solves the issue.

- Delete the consumed `LoadFunction` target during `apply_modifier_chain_to_loaded_fn`[[1]](https://github.com/Quantinuum/tket2/blob/c118aed07d0c8dbc64d8ac91a7bfdb5296956368/tket/src/modifier/modifier_resolver/call_modify.rs#L92-L93) and run dead-function after[[2]](
https://github.com/Quantinuum/tket2/blob/c118aed07d0c8dbc64d8ac91a7bfdb5296956368/tket/src/modifier/modifier_resolver.rs#L1174-L1175).
- Added in test some assertions to check that the function block and the modifier block are removed after the passes have been run.


tket/src/modifier/modifier_resolver.rs
```diff
@@ -1088,6 +1092,81 @@ impl<N: HugrNode> ModifierResolver<N> {
     }
 }
 
+fn module_child_containing<N: HugrNode>(h: &impl HugrView<Node = N>, node: N) -> Option<N> {
```
aborgna-q, line 1103: Pls add docs to functions

tket/src/modifier/modifier_resolver.rs
```diff
@@ -323,6 +323,8 @@ pub struct ModifierResolver<N = Node> {
     // ```
     // _modified_functions: HashMap<N, (CombinedModifier, Node)>,
     // ```
+    /// Original functions for which the resolver generated modified replacements.
+    modified_functions: Vec<N>,
```
aborgna-q: Could we use a BTreeSet or something similar to avoid linear searchs?

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1179,6 +1258,10 @@ pub fn resolve_modifier_with_entrypoints(
     // were produced or left behind by the resolution passes above.
     delete_phase(h, entry_points)?;
 
+    // Remove only original functions for which this resolver generated modified
+    // replacements, and only when no remaining non-obsolete function uses them.
```
aborgna-q, line 1309: This new behaviour should be documented in the public docs, both here and in the pass (for both rust and python defs).

aborgna-q (dismissed):
Nice. Just a couple comments.

nicolaassolini-qntm:
@aborgna-q I addressed your comment and the Copilot one

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1143,6 +1163,13 @@ fn remove_unused_modified_functions<N: HugrNode>(
         candidates.remove(&entrypoint_owner);
     }
 
+    // Remove public functions from candidates to preserve exported API functions.
+    candidates.retain(|func| {
+        h.get_optype(*func)
+            .as_func_defn()
+            .is_some_and(|func_defn| *func_defn.visibility() != Visibility::Public)
+    });
```
aborgna-q: We only need to preserve public functions if the scope says so.

Use `PassScope::preserve_interface` or `PassScope::in_scope` instead (and update the pass docs).

tket/src/modifier/modifier_resolver.rs
```diff
@@ -102,6 +103,7 @@
 //! - User defined extension ops: There is no way to infer modified unknown extension ops.
 //!   We currently try to insert the original optype without any modification,
 //!   but this could result in an unexpected error.
+#[cfg(test)]
 use hugr_core::Visibility;
```
aborgna-q: The import should go in the `mod tests` submodule below.

aborgna-q (approved):
LGTM

PR #1558, opened 2026-04-30
feat!: Add `measurement` extension and change `tket.quantum` / `tket.qsystem` extension measurement ops return type

### LINKED ISSUE #1484: [Feature]: Add `Measurement` type and adjust measurement ops to use it instead of `tket.bool`
### Problem Statement

Needed for changes to measurement interface in Guppy

### Proposed Solution

Steps:
- quantum extension
	- add `Measurement` type
	- `MeasureFree`: change output from opaque bool to measurement
	- `Measure`: change output (qubit, sum bool) to (qubit, measurement) 
- qsystem extension
	- `Measure`: change output from opaque bool to measurement
	- `MeasureReset` change output  (qubit, opaque bool) to (qubit, measurement)
- passes
	- new `ReplaceMeasurement` pass which replaces measurement types with future<bool> and ops with lazy qsystem versions
	- change `ReplaceBool` pass to replace tket bools to measurements and old version measurement ops with new version measurement ops for as an upgrade pass
		- check whether we want the (borrow) array replacement logic in a non-upgrade pass (even though I don't think we have other cases currently where non-linear types are replaced with linear types)
		- don't delete `tket.bool` extension for now to use in upgrade pass, but deprecate (if possible to deprecate extensions?)

### Component

Rust library, Python library

### Effort Estimate

Medium - Moderate complexity

### Additional Context

_No response_

tatiana-s:
Closes #1484

Also removes the `bool` extension and deletes all relevant passes + adds a measurement replacement pass

BREAKING CHANGE: `tket.quantum` extension - `MeasureFree` returns a `Measurement` instead of a `tket.bool`
BREAKING CHANGE: `tket.qsystem` extension - `Measure` and `MeasureReset` are removed, `FutureToMeasurement` is added 
BREAKING CHANGE: (tket-exts) `opaque_bool` and `bool` removed


tket-exts/src/tket_exts/data/tket/measurement.json
```diff
@@ -0,0 +1,102 @@
...
+            }
+          ]
+        }
+      },
+      "binary": false
+    },
+    "Free": {
+      "extension": "tket.measurement",
+      "name": "Free",
+      "description": "Consume a measurement without reading it.",
```
acl-cqc:
```suggestion
      "description": "Discard a measurement without reading it.",
```

acl-cqc:
```suggestion
      "description": "Discard (a copy of) a measurement without reading it.",
```

tket-exts/src/tket_exts/data/tket/measurement.json
```diff
@@ -0,0 +1,102 @@
+{
+  "version": "0.2.0",
+  "name": "tket.measurement",
+  "types": {
+    "Measurement": {
+      "extension": "tket.measurement",
+      "name": "Measurement",
+      "params": [],
+      "description": "A type representing the result of a measurement operation",
```
acl-cqc: Might be worth adding here, can be copied and discarded via appropriate ops. Otherwise it'd be no better than passing the unmeasured qubit ;)

tket-exts/src/tket_exts/data/tket/measurement.json
```diff
@@ -0,0 +1,102 @@
...
+          ],
+          "output": []
+        }
+      },
+      "binary": false
+    },
+    "Read": {
+      "extension": "tket.measurement",
+      "name": "Read",
+      "description": "Read a measurement, consuming it and returning a Hugr bool.",
```
acl-cqc:
```suggestion
      "description": "Consumes a measurement, converting it to a bool (Sum type).",
```

(defining "read" as being "Read a measurement" is just recursion)

tket-exts/src/tket_exts/tket/measurement.py
```diff
@@ -0,0 +1,51 @@
...
+
+from hugr.ext import Extension, OpDef, TypeDef
+from hugr.ops import ExtOp
+from hugr.tys import ExtType
+
+from ._util import TketExtension, load_extension
+
+
+class MeasurementExtension(TketExtension):
+    """Measurement result operations."""
```
acl-cqc:
```suggestion
    """Operations on Measurement Results."""
```
i.e. to avoid ambiguity vs "measurement (result-operations)"

acl-cqc:
```suggestion
    """The Measurement result type and operations on it."""
```

acl-cqc: Maybe we should go for MeasurementResult or just MeasureResult? Or just Measurement??? (I bet this has been debated before)

tket-exts/src/tket_exts/tket/measurement.py
```diff
@@ -0,0 +1,51 @@
...
+    def OPS(self) -> List[OpDef]:
+        """Return the operations defined by this extension"""
+        return [
+            self.dup.op_def(),
+            self.free.op_def(),
+            self.read.op_def(),
+        ]
+
+    @functools.cached_property
+    def measurement_t(self) -> ExtType:
```
acl-cqc, line 32: See, measurement result is too long that you don't want to use it ;-) ;-)

`measure_result_t`? Or just call it Measurement? That's what your implicitly doing in code if this is called `measurement_t`?

tket-exts/src/tket_exts/tket/measurement.py
```diff
@@ -0,0 +1,51 @@
...
+        return self().get_type("Measurement").instantiate([])
+
+    @functools.cached_property
+    def dup(self) -> ExtOp:
+        """Duplicate a measurement, consuming the original value."""
+        return self().get_op("Dup").instantiate()
+
+    @functools.cached_property
+    def free(self) -> ExtOp:
+        """Consume a measurement without reading it."""
```
acl-cqc: Ah as per json. I'm not sure if either is generated from the other here.

tket-exts/src/tket_exts/__init__.py
```diff
@@ -40,9 +41,15 @@
     "wasm",
     "modifier",
     "global_phase",
+    "measurement",
 ]
 
-bool: BoolExtension = tket.bool.BoolExtension()
+
+@deprecated("Use the built-in `Bool` type instead")
```
acl-cqc:
```suggestion
@deprecated("Use the `Bool` <Hugr/prelude/Sum> type instead")
```
pick whatever's most appropriate, but otherwise I'm not sure what Bool is built-into

acl-cqc: Moreover, this is a breaking change - you now need to add `()` - as well as being deprecated...can we avoid the need for extra brackets?

tket-py/tket/extensions/__init__.py
acl-cqc, line 16: I mean, how long do we go on supporting tket-exts 0.10 for? :)

tket-py/tket/extensions/__init__.py
```diff
@@ -19,15 +18,13 @@
 
     new_exts = ["gpu"]
 except ImportError:
-    bool = opaque_bool  # type: ignore[assignment]
     new_exts = []
 
 
-# TODO: Remove the deprecated `opaque_bool` export in a breaking change.
+# TODO: Remove the deprecated `bool` export in a breaking change.
```
acl-cqc: Yah, as [previously](https://github.com/Quantinuum/tket2/pull/1558/changes#r3225360556), I mean this is a breaking change already

tket-exts/src/tket_exts/__init__.py
acl-cqc: And....the rust bool extension is gone, but the python one lives on (deprecated)? Was the rust `BOOL_EXTENSION` the python `opaque_bool` ??

tket-qsystem/src/extension/qsystem.rs
```diff
@@ -636,11 +640,11 @@ mod test {
                 .unwrap();
 
             let q0 = func_builder.add_rz(q0, angle).unwrap();
-            let [q0, _b] = func_builder.add_measure_reset(q0).unwrap();
-            let b = func_builder.add_measure(q0).unwrap();
+            let [q0, b1] = func_builder.add_measure_reset(q0).unwrap();
```
acl-cqc: suggest 'm1' and 'm2' rather than 'b1' and 'b2' ;)

tket-qsystem/src/replace_measurement.rs
```diff
@@ -0,0 +1,255 @@
...
+    /// The HUGR was found to contain non-local edges.
+    NonLocalEdgesError(FindNonLocalEdgesError<N>),
+    /// There was an error while replacing the type/ops.
+    ReplacementError(ReplaceTypesError),
+}
+
+/// A HUGR -> HUGR pass replacing `tket.measurement` with `future(bool_t)`.
+///
+/// [TketOp::MeasureFree] and [QSystemOp::Measure] ops are replaced by
+/// [QSystemOp::LazyMeasure], while [QSystemOp::MeasureReset] ops are
```
acl-cqc: Hmmm. This seems odd. See overall PR comment.

tket-qsystem/src/replace_measurement.rs
```diff
@@ -0,0 +1,255 @@
...
+    use hugr::extension::simple_op::MakeOpDef;
+    use hugr::ops::OpType;
+    use hugr::types::Type;
+    use hugr::types::TypeRow;
+    use rstest::rstest;
+
+    use crate::extension::qsystem::QSystemOpBuilder;
+
+    #[test]
+    fn test_replace_measurement_type_and_read() {
```
acl-cqc: Tests are fine, but they do just kinda repeat the single-op-to-single-op definition of the lowering. Lowering a complete circuit with multiple TketOp's wired together that validates before and after would give a stronger guarantee i.e. that the replacement ops were appropriate replacements (with consistent types) that can still be wired together in the same configuration

tket-qsystem/src/replace_measurement.rs
```diff
@@ -0,0 +1,255 @@
...
+        assert_eq!(sig.input(), &TypeRow::from(vec![future_type(bool_t())]));
+        assert_eq!(sig.output(), &TypeRow::from(vec![bool_t()]));
+
+        // Check that the op was replaced.
+        assert!(
+            h.nodes()
+                .any(|n| FutureOpDef::try_from(h.get_optype(n)) == Ok(FutureOpDef::Read))
+        );
+        assert!(
+            !h.nodes()
```
acl-cqc: ...and perhaps make this a broader check that no op is from the measurement extension (or tket extension!)

tket-qsystem/src/replace_measurement.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    #[test]
+    fn test_replace_measure_reset_op() {
+        let mut dfb = DFGBuilder::new(inout_sig(
+            vec![hugr::extension::prelude::qb_t()],
+            vec![hugr::extension::prelude::qb_t(), measurement_type()],
+        ))
+        .unwrap();
+        let [q] = dfb.input_wires_arr();
+        let out = dfb.add_measure_reset(q).unwrap();
```
acl-cqc: nit: good to test `fn add_measure_reset` I guess, but you could combine this with previous if you used `add_op` and just finished with its outputs

tket/src/extension/measurement.rs
```diff
@@ -0,0 +1,208 @@
...
+        MEASUREMENT_TYPE_ID.to_owned(),
+        vec![],
+        "A type representing the result of a measurement operation".into(),
+        TypeBound::Linear.into(),
+        &extension_ref,
+    )
+}
+
+/// Returns a `Measurement` [CustomType].
+pub fn measurement_custom_type(extension_ref: &Weak<Extension>) -> CustomType {
```
acl-cqc: I guess this has to be like this, but does it really need to be `pub` (and re-exported!)? Surely it should only be used during initialization of the extension itself (when the `Weak` is all you've got)? (Perhaps even inline it into `init_signature` as that'll be the only use left)

tket/src/extension/measurement.rs
```diff
@@ -0,0 +1,208 @@
...
+        vec![],
+        MEASUREMENT_EXTENSION_ID,
+        TypeBound::Linear,
+        extension_ref,
+    )
+}
+
+/// Returns a `Measurement` [Type].
+pub fn measurement_type() -> Type {
+    measurement_custom_type(&Arc::downgrade(&MEASUREMENT_EXTENSION)).into()
```
acl-cqc:
```suggestion
    MEASUREMENT_EXTENSION.get_type(MEASUREMENT_TYPE_ID).instantiate(&[]).unwrap()
```

acl-cqc: Possibly returning CustomType, as you can easily `.into()` to get a Type, or possibly even two funcs, if we really want

tket/src/extension/measurement.rs
```diff
@@ -0,0 +1,208 @@
...
+        try_from_name(op_def.name(), op_def.extension_id())
+    }
+
+    fn extension(&self) -> ExtensionId {
+        MEASUREMENT_EXTENSION_ID
+    }
+
+    fn description(&self) -> String {
+        match self {
+			MeasurementOp::Read => "Read a measurement, consuming it and returning a Hugr bool.".into(),
```
acl-cqc: nit: (again) recursive definition of "read"

tket/src/serialize/pytket/config/type_translators.rs
```diff
@@ -158,16 +157,6 @@ impl TypeTranslatorSet {
         if typ1 == typ2 {
```
acl-cqc: So for now this method can just be `{ typ1 == typ2 }`

tket/src/serialize/pytket/config/type_translators.rs
```diff
@@ -180,37 +169,19 @@ impl TypeTranslatorSet {
...
         target_type: &Type,
-        builder: &mut DFGBuilder<&mut Hugr>,
+        _builder: &mut DFGBuilder<&mut Hugr>,
     ) -> Result<Wire, PytketDecodeError> {
         if initial_type == target_type {
             return Ok(wire);
         }
 
-        let map_build_error = |e: BuildError| PytketDecodeErrorInner::CannotTranslateWire {
+        let _map_build_error = |e: BuildError| PytketDecodeErrorInner::CannotTranslateWire {
```
acl-cqc: Can certainly remove this variable. It's tempting to remove a whole lot more (like the whole `transform_typed_value`) but I guess it can stay for now. (Really it'd only be needed if we enact the "TODO: We should allow custom TypeTranslators to expand this checks, and implement their own translations." above)

tket/src/serialize/pytket/extension/tket.rs
```diff
@@ -72,11 +72,17 @@ impl TketOpEmitter {
             TketOp::Ry => PytketOptype::Ry,
             TketOp::Toffoli => PytketOptype::CCX,
             TketOp::Reset => PytketOptype::Reset,
-            TketOp::Measure => PytketOptype::Measure,
+            TketOp::Measure => {
+                // As the measurement type is not supported, we don't translate
```
acl-cqc:
```suggestion
                // As the measurement type is not supported, for now we don't translate measure ops either.
```

tket/src/serialize/pytket/extension/tket.rs
```diff
@@ -72,11 +72,17 @@ impl TketOpEmitter {
...
+                // measure ops as well for now.
+                return Ok(EncodeStatus::Unsupported);
+            }
             // We translate `MeasureFree` the same way as a `Measure` operation.
-            // Since the node does not have outputs the qubit/bit will simply be ignored,
-            // but will appear when collecting the final pytket registers.
-            TketOp::MeasureFree => PytketOptype::Measure,
+            TketOp::MeasureFree => {
+                // As the measurement type is not supported, we don't translate
```
acl-cqc: as previous

tket/src/serialize/pytket/tests.rs
```diff
@@ -291,14 +291,14 @@ fn circ_preset_qubits() -> Hugr {
 /// including multiple outputs of the same register.
 #[fixture]
 fn circ_preset_bits() -> Hugr {
-    let input_t = vec![bool_type()];
-    let output_t = vec![bool_type(), bool_type(), bool_type()];
+    let input_t = vec![bool_t()];
+    let output_t = vec![bool_t(), bool_t(), bool_t()];
```
```suggestion
    let output_t = vec![bool_t(); 3];
```

tket/src/serialize/pytket/tests.rs
```diff
@@ -918,12 +901,12 @@ fn check_no_tk1_ops(hugr: &Hugr) {
...
 #[rstest]
 #[case::simple(SIMPLE_JSON, 2, 2, false)]
-#[case::simple_measure(SIMPLE_MEASURE, 4, 2, false)]
+#[case::simple_measure(SIMPLE_MEASURE, 4, 2, true)]
 #[case::multi_register(MULTI_REGISTER, 2, 3, false)]
 #[case::unknown_op(UNKNOWN_OP, 2, 3, true)]
 #[case::small_parametrized(SMALL_PARAMETERIZED, 1, 1, false)]
 #[case::parametrized(PARAMETERIZED, 4, 2, true)] // TK1 op is not supported
-#[case::barrier(BARRIER, 3, 3, false)]
+#[case::barrier(BARRIER, 3, 3, true)]
```
acl-cqc, line 1006: Why does this now use tket1_ops but didn't before?

tket/src/ops.rs
```diff
@@ -180,7 +180,7 @@ pub enum TketOp {
     /// - A qubit
     ///
     /// Outputs:
-    /// - A boolean indicating whether the qubit was measured as 1
+    /// - A measurement indicating whether the qubit was measured as 1
```
acl-cqc:
```suggestion
    /// - A [Measurement](measurement_type) indicating whether the qubit was measured as 1
```

tket/src/extension.rs
```diff
@@ -18,18 +18,21 @@ use hugr::types::{PolyFuncType, PolyFuncTypeRV};
...
 /// Definition for ops used by Guppy.
 pub mod guppy;
+/// Definition for measurement types.
+pub mod measurement;
 pub mod modifier;
 /// Definition for Angle ops and types.
 pub mod rotation;
 pub mod sympy;
 
+pub use measurement::{
```
acl-cqc: Are we sure about this? `tket.extension` doesn't re-export anything else from any submodule; elsewhere, clients import from *the particular* extension defining the thing.

acl-cqc (commented):
Thanks Tatiana! Generally looks good, mostly just ideas for wordings. However there are a couple of things I'm not sure about...
* What's going on with the Rust BOOL_EXTENSION being removed, but the python bool extension being merely deprecated?
* Why does the QSystem extension include ops at *both* levels of abstraction (Measure(Reset) returning MeasurementResult and LazyMeasure(Reset) returning `Future<bool_t>`)?

On the latter...well, it's not because LazyMeasure is any lazier than Measure; they do the same thing because one is lowered into the other.

It might be so that people can write at both "levels" whilst still using QSystem gates, but they could achieve that by using  TketOp::Measure alongside QSystem gates.

Is there some advantage in using the "non-Lazy" QSystem Measure ops rather than either TketOp measure-ops or just the "Lazy" ones? Yes the non-lazy ops give them back MeasurementResults, but the only ops available to manipulate MeasurementResults have equivalents available on Futures too, and if you want MRs so you can (say) put them into collections alongside MRs from TketOps, why not use the TketOps themselves? (Or if you don't want to use the Tket extension then just use Lazy measure and Futures directly)...

So I am wondering if we should remove LazyMeasure(Reset) and make Measure(Reset) return a future.

aborgna-q:
Pelase update the extension versions! (either breaking or non-breaking bumps as appropriate)

tket-exts/src/tket_exts/tket/measurement.py
tatiana-s: I am not sure what you mean because it already is called Measurement everywhere? (besides some comments which I should adjust to avoid confusion it seems)

tket-exts/src/tket_exts/__init__.py
tatiana-s: I will just remove the python extension completely too 👍

tket/src/serialize/pytket/tests.rs
tatiana-s: Until https://github.com/Quantinuum/tket2/issues/1570 is done (needs to be done before this can be released, I guess could have been done in this PR but seemed better to separate since this already is a large PR) measurements can't get encoded/decoded so they become opaque tket1 ops

tket/src/extension/measurement.rs
tatiana-s: It's done like this in a couple of other extensions, not sure if there is a particular reason

aborgna-q:
fyi, the CI is currently failing at `serialize::pytket::tests::encoded_circuit_attributes` because the test HUGR does not contain any ops serializable in pytket.

This is expected, as the circuit is only  measurements and qfrees. Looking at the test we're not doing anything that's not in other tests already so I'd just remove it altogether.

tatiana-s:
@acl-cqc I've now had to update quite a lot of QIS snapshot tests, probably worth looking at least a few to check it the changes to the LLVM look reasonable?

tket-exts/src/tket_exts/__init__.py
```diff
@@ -86,6 +79,7 @@ def tket_registry() -> ExtensionRegistry:
         tket.wasm.WasmExtension(),
         tket.modifier.ModifierExtension(),
         tket.global_phase.GlobalPhaseExtension(),
+        tket.measurement.MeasurementExtension(),
```
acl-cqc, line 90: why do we not use the toplevel variables `measurement`, `global_phase`, etc.?

tket-exts/tests/test_validate_exts.py
```diff
@@ -51,21 +34,30 @@ def ext_gpu() -> Tuple[TketExtension, List[ExtType], List[ExtOp]]:
 
 def ext_guppy() -> Tuple[TketExtension, List[ExtType], List[ExtOp]]:
     ext = tket_exts.guppy
-    bool_t = tket_exts.bool.bool_t
+    rot_t = tket_exts.rotation.rotation  # Arbitrary non-linear type for testing.
```
acl-cqc, line 37: Shouldn't this be an arbitrary *affine* type? Or do we not have any....yeah we do, arrays?

tket-qsystem/src/replace_measurement.rs
```diff
@@ -0,0 +1,187 @@
...
+            .unwrap()
+            .outputs_arr();
+
+        // Read one
+        let b1 = circuit
+            .add_dataflow_op(MeasurementOp::Read, [m1_1])
+            .unwrap()
+            .out_wire(0);
+
+        // Free the rest
```
acl-cqc: Looks good - could I, say, put m1_2 and m2 into an array and free that?

I note this pass is deployed *before* LowerDropsPass, so we might want to check dropping an array of measurements lowers correctly through the whole sequence of passes. (EDIT ok I see LowerDropsPass already tells the linearizer about Future, and LowerDropsPass should never see Measurements, so that should be fine, but it would be a good test if we don't have that already?)

tket-py/tket/extensions/__init__.py
```diff
@@ -19,15 +18,13 @@
...
 __all__ = [
     "debug",
     "bool",
```
acl-cqc: I'm surprised this works, there is no local variable called `bool` in scope right? ooh - this might be re-exporting the python type bool???

tket-exts/tests/test_validate_exts.py
```diff
@@ -51,21 +34,30 @@ def ext_gpu() -> Tuple[TketExtension, List[ExtType], List[ExtOp]]:
...
         [],
-        [ext.drop(bool_t)],
+        [ext.drop(rot_t)],
     )
 
 
 def ext_futures() -> Tuple[TketExtension, List[ExtType], List[ExtOp]]:
     ext = tket_exts.futures
-    bool_t = tket_exts.bool.bool_t
+    rot_t = tket_exts.rotation.rotation  # Arbitrary non-linear type for testing.
```
acl-cqc, line 47: should this also be affine? e.g. an array?

tket-exts/src/tket_exts/tket/measurement.py
acl-cqc: Yeah this is fine (I didn't mean to change this, and you didn't) - just to change "measurement result" everywhere else to just "measurement", which you pretty much have 👍  :-)

acl-cqc (commented):
Looks good thanks @tatiana-s. Think structure of extensions/measure ops is better now and this is pretty much there. The mass of changes to LLVM are hard to grok (much of the noise is blocks being reordered), I will find some way to dig into at least a couple of these before approval but thought I'd give a few comments now - these are pretty minor although I propose another test.

To confirm, all the `tool.uv.sources` updates in the `test_files`...IIUC the idea that we have to merge this + release tket2 first, only then can we merge the [guppy PR](https://github.com/Quantinuum/guppylang/pull/1707), and then release guppy, whereupon we can make another PR to remove these directives.

Test looks better now, TYVM, but I wonder if we should have one where we drop an array of Measurements (maybe after copying the array and `collect_measurements` only on one copy) and run that through the whole QSystemPass.

acl-cqc (commented):
Had a look at a couple of snapshots - `discard_qb_array`, which really shouldn't have changed, and `measure_qb_array`, which one would expect to.

Diffing both against branch `acl/update-guppylang` (which just updates the version of `guppylang` used in `generate_hugrs.py`, to a more recent guppy main but not including any measure-future changes) reduces the diff significantly. `discard_qb_array` is indeed unchanged modulo renaming(/numbering, but not ordering) of many basic blocks. `measure_qb_array` gets about 50 lines longer, including 3 new BBs, but I think that's consistent with this PR adding a separate call to `collect_measurements` with its own loop (right?).

So snapshots look plausible.

tket-exts/src/tket_exts/__init__.py
tatiana-s: Not sure, I just added things in a way that matched what was already there

tket-exts/tests/test_validate_exts.py
tatiana-s: Why would it need to be affine? This just needs to be a random type that can go into a future, the opaque bool it used beforehand also wasn't affine

tket-py/tket/extensions/__init__.py
tatiana-s: Forgot to remove this, good spot

tket-qsystem/src/replace_measurement.rs
tatiana-s: `LowerDropsPass` doesn't have anything to do with the free op on measurements, both before and after replacement through futures - drops are only inserted in guppy on affine types that are linear in hugr, measurements and therefore arrays of measurements are linear in both guppy and hugr, you can't drop an array of measurements at all. So I am not sure what the test would show?

tket-exts/tests/test_validate_exts.py
acl-cqc: Well....I can't *remember* why I wrote this now, but looking around for clues...does the `ext.drop(rot_t)` do anything on a copyable type? (It would on an array)

tket-exts/tests/test_validate_exts.py
acl-cqc: (This one is fine, as it's the future of `rot_t` that's being tested, not the `rot_t` itself)

tket-py/src/passes.rs
```diff
@@ -216,13 +216,12 @@ fn resolve_modifiers(circ: &mut CompilationState, scope: Option<PyPassScope>) ->
 }
 
 #[pyfunction]
-#[pyo3(signature=(circ, *, constant_fold = true, monomorphize = true, force_order = true, lazify = true, hide_funcs = true, scope = None))]
+#[pyo3(signature=(circ, *, constant_fold = true, monomorphize = true, force_order = true, hide_funcs = true, scope = None))]
```
acl-cqc, line 219: Hmmm so there is no way to turn off the linearization of Measurement into Future? I admit you wouldn't want to run the whole pass without that,

tket-qsystem/src/extension/qsystem/common.rs
```diff
@@ -100,7 +100,9 @@ impl SharedOp {
                 vec![Type::from(option_type(one_qb_row.clone()))],
             ),
             SharedOp::QFree => Signature::new(one_qb_row.clone(), type_row![]),
-            SharedOp::MeasureReset => Signature::new(one_qb_row, vec![qb_t(), bool_type()]),
+            SharedOp::FutureToMeasurement => {
+                Signature::new(vec![future_type(bool_t())], vec![measurement_type()])
```
acl-cqc:
```suggestion
                Signature::new([future_type(bool_t())], [measurement_type()])
```

tket-qsystem/src/extension/qsystem/common.rs
```diff
@@ -218,7 +208,7 @@ pub(crate) trait PhasedXRzSynth: CommonOpBuilder<Self::Op> {
     /// Build a TryQAlloc gate using the platform's native operation.
     fn synth_try_alloc(&mut self) -> Result<Wire, BuildError>;
     /// Build a MeasureReset gate using the platform's native operation.
-    fn synth_measure_reset(&mut self, qb: Wire) -> Result<[Wire; 2], BuildError>;
+    fn synth_lazy_measure_reset(&mut self, qb: Wire) -> Result<[Wire; 2], BuildError>;
```
acl-cqc, line 211: update doc on prev line

tket-qsystem/src/extension/qsystem/helios.rs
```diff
@@ -546,7 +530,7 @@ mod test {
         let hugr = {
             let mut func_builder = FunctionBuilder::new(
                 "all_ops",
-                Signature::new(vec![qb_t(), float64_type()], vec![bool_type()]),
+                Signature::new(vec![qb_t(), float64_type()], vec![bool_t()]),
```
acl-cqc:
```suggestion
                Signature::new([qb_t(), float64_type()], [bool_t()]),
```

tket/src/extension/measurement.rs
acl-cqc: Then I think we can do significantly better than elsewhere and I won't make you update other places to match this ;-)

rename this function and hide it. It'll only be called directly in one place, inside the impl for MakeRegisteredOp

Then add a new
`pub fn measurement_custom_type() -> CustomType ` - this could either
* call the hidden one with `&Arc::downgrade(&MEASUREMENT_EXTENSIONSION)`, as you are currently doing in `pub fn measurement_type()`
* or, instantiate via `MEASUREMENT_EXTENSION.get_type(MEASUREMENT_TYPE_ID).unwrap().instantiate([]).unwrap()` - I kinda feel that's better TBH and that would leave the hidden method with only one caller whereupon you could just inline + remove it

Finally change `pub fn measurement_type() -> Type` just to call `measurement_custom_type().into()`

tket-exts/src/tket_exts/data/tket/measurement.json
acl-cqc: Perhaps just add the word "copyable" in the description in order to highlight this

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -127,12 +144,229 @@ fn register_legacy_qsystem_replacements(lowerer: &mut ReplaceTypes, platform: QS
     }
 }
 
+/// Register any replacements related to the `Measurement` type.
+fn register_measurement_replacements(lowerer: &mut ReplaceTypes) {
+    // As the measurement type acts like an alias for `Future<Bool>`, most replacements
+    // are straightforward.
+    lowerer.set_replace_type(
+        measurement_type().as_extension().unwrap().clone(),
```
acl-cqc: See https://github.com/Quantinuum/tket2/pull/1558/changes#r3347834631 then this here will just be `measurement_custom_type()` which is much easier

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -127,12 +144,229 @@ fn register_legacy_qsystem_replacements(lowerer: &mut ReplaceTypes, platform: QS
...
+    }
+    .to_extension_op()
+    .unwrap();
+    lowerer.set_replace_op(
+        &MeasurementOp::Read.to_extension_op().unwrap(),
+        NodeTemplate::SingleOp(future_bool_op.into()),
+    );
+
+    // The following replacements and lineariser registrations are needed in order to
+    // allow the copyable `Measurement` type to be replaced with the non-copyable
```
acl-cqc:
```suggestion
    // allow the copyable `Measurement` type to be replaced with the linear
```

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -127,12 +144,229 @@ fn register_legacy_qsystem_replacements(lowerer: &mut ReplaceTypes, platform: QS
...
+    let free_op = FutureOp {
+        op: FutureOpDef::Free,
+        typ: bool_t(),
+    }
+    .to_extension_op()
+    .unwrap();
+    lowerer
+        .linearizer_mut()
+        .register_simple(
+            future_type(bool_t()).as_extension().unwrap().clone(),
```
acl-cqc: I mean it'd be great to have a `future_custom_t` too but I guess that's not for now

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -127,12 +144,229 @@ fn register_legacy_qsystem_replacements(lowerer: &mut ReplaceTypes, platform: QS
...
+        (
+            array::EXTENSION.to_owned(),
+            array_type as fn(u64, Type) -> Type,
+        ),
+        (
+            borrow_array::EXTENSION.to_owned(),
+            borrow_array_type as fn(u64, Type) -> Type,
+        ),
+    ] {
+        // `Clone` ops get replaced with a DFG that the lineariser can act on.
```
acl-cqc: I kinda feel this one probably belongs in ReplaceTypes::default() TBH....much easier to do that change now ReplaceTypes is in this repo

tket-qsystem/src/lower_drops.rs
```diff
@@ -37,7 +37,7 @@ impl<H: HugrMut<Node = Node>> ComposablePass<H> for LowerDropsPass {
 
         // future(bool) is not in the default linearizer handler so we add it here.
         // TODO: Create ReplaceTypes with future(bool) linearized by default to avoid
-        // code duplication with ReplaceBools pass.
+        // code duplication with ReplaceMeasurements pass.
```
acl-cqc: This would be a good time to do this, I think. You've just moved the other copy of this code inside `register_measurement_replacements` so it'd be good to just put it in its own `register_future_linearizers`. Or have a `tket-qsystem` function that gives back `ReplaceTypes::default()` with these.

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -921,4 +1168,155 @@ mod test {
             .collect();
         assert_eq!(sol_ops, vec![SolOp::TryQAlloc, SolOp::Reset, SolOp::QFree],);
     }
+
+    /// Build a HUGR containing measurement ops (both from `tket.quantum` and
+    /// `tket.qsystem`) and verify that it no longer contains any measurement types
+    /// after the pass.
```
acl-cqc, line 988: Sweet :)

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -921,4 +1168,155 @@ mod test {
...
+        dfb.add_generic_array_discard::<AK>(elem_ty, size, arr2)
+            .unwrap();
+        let mut h = dfb.finish_hugr_with_outputs([arr1]).unwrap();
+
+        h.validate().unwrap();
+        lower_tk2_ops(&mut h, PassScope::Global(Preserve::Public), platform).unwrap();
+        h.validate().unwrap();
+
+        let sig = h.signature(h.entrypoint()).unwrap();
+        let future_arr_ty = AK::ty(size, future_type(bool_t()));
```
acl-cqc: easier just to put this in a TypeRow here, then no need to clone

tket-qsystem/src/extension/qsystem/sol.rs
```diff
@@ -80,12 +78,13 @@ pub enum SolOp {
     QFree,
     /// Reset a qubit to the Z |0> eigenstate.
     Reset,
-    /// Measure a qubit and reset it to the Z |0> eigenstate.
-    MeasureReset,
     /// Measure a qubit (return 0 or 1) or detect leakage (return 2).
     LazyMeasureLeaked,
     /// PhasedXX gate (alias 'rpp')
     PhasedXX,
+    /// Convert a `Future<Bool>` to a `Measurement`` (for compatibility with the TKET
```
acl-cqc:
```suggestion
    /// Convert a `Future<Bool>` to a `Measurement` (for compatibility with the TKET
```

tket-qsystem/src/pytket/tests.rs
```diff
@@ -275,73 +273,72 @@ fn json_roundtrip(
...
-
-    let reser = SerialCircuit::encode(
-        &deser,
-        EncodeOptions::new().with_config(qsystem_encoder_config(QSystemPlatform::Helios)),
-    )
-    .unwrap();
-    validate_serial_circ(&reser);
-    compare_serial_circs(ser, &reser);
-}
+// TODO: Revisit as part of https://github.com/Quantinuum/tket2/issues/1570.
```
acl-cqc: Would it be better to paste the old code into that issue and remove it here? It's basically going to rot either way until we solve #1570

tket-qsystem/src/lib.rs
```diff
@@ -461,15 +441,14 @@ mod test {
                 .count()
         };
 
-        // Check there are no public funcs (after hiding)
+        // Check there are no public funcs regardless of hiding or not.
```
acl-cqc: I'm a little surprised that this doesn't start having something to hide again given we've kinda reinstated lazification

acl-cqc: Ah - change the array above to be an array of Measurements not booleans

tket/src/serialize/pytket/config/type_translators.rs
```diff
@@ -180,37 +166,11 @@ impl TypeTranslatorSet {
         wire: Wire,
         initial_type: &Type,
         target_type: &Type,
-        builder: &mut DFGBuilder<&mut Hugr>,
+        _builder: &mut DFGBuilder<&mut Hugr>,
```
acl-cqc, line 169: This suggests we may want some support for Measurement here instead, but that can be a future PR. Raise an issue perhaps?

tket/src/serialize/pytket/extension/tket.rs
acl-cqc: I repeat this rewording suggestion, this is not good English (sorry!)

tket/src/serialize/pytket/extension/tket.rs
acl-cqc: and here

tket/src/serialize/pytket/tests.rs
```diff
@@ -992,7 +968,10 @@ fn circ_discard_first_qubit() -> Hugr {
 
     let [q1, q2] = h.input_wires_arr();
 
-    h.add_dataflow_op(TketOp::MeasureFree, [q1]).unwrap();
+    let [_msmt] = h
```
acl-cqc: Why the change here?

tket/src/serialize/pytket/tests.rs
```diff
@@ -1501,43 +1455,44 @@ fn serial_decode_missing_output_bit_returns_decode_error() {
...
-        .ok()
-        .expect("single reassembled function");
-    let reassembled_signature = reassembled
-        .get_optype(reassembled_function)
-        .inner_function_type()
-        .expect("Function definition")
-        .into_owned();
-    assert_eq!(&circ_signature.output, &reassembled_signature.output);
-}
+// TODO: Revisit as part of https://github.com/Quantinuum/tket2/issues/1570.
```
acl-cqc: Similarly, consider adding something to the issue - either the removed code, or a link to the PR removing it or the last commit with it

acl-cqc (commented):
Blimey. Kudos @tatiana-s for your immense work on this epic. I feel exhausted just having read it (and I did not look at all the snapshots this time!), never mind actually writing it...

Mostly this is great, I like the new Copyable Measurement (even if not the extra code required but we did have that before!).

Still a couple issues/concerns - in comments (along with a bunch of one-liners) but I'll list here too
* Test-hide-funcs could/should be able to test something now I think
* Refactor configuring of ReplaceTypes. You are moving this code so it'd be good to do it now. Some configury could probably go into `ReplaceTypes::default` i.e. bits relating only to types defined in hugr-core (which includes arrays, but not guppy drop, future or measurement). Other bits e.g. future-linearization we should have a common helper, perhaps a tket-qsystem factory method that adds this to `ReplaceTypes::default`.
* Could improve `measurement_type` and `measurement_custom_type`, this is not a big change but would be significantly neater where it's used

Appreciate yet another iteration but the first two do result from the significant change of linearizing Measurement and so I think this will be short as long as there are not more massive changes on tket2 main to merge in first...

I also ponder
* <strike>The many `tool.uv.sources` that we are adding to `guppy_examples`, `guppy_optimization` etc....given we are going to change these all when guppy is released, do the tests fail if we leave the guppy-sources unchanged and hence the hugrs? That would significantly reduce the number of files changed by the PR, admittedly at the cost of reducing the checks that this PR "works" with...an unreleased version of guppy, so maybe that's not so bad. Can you revert those bits?</strike>
EDIT: discussed with Agustin, we are not too bothered about maintaining support for old hugrs now, so ok let's go with temporary pinning

tket-qsystem/src/pytket/tests.rs
tatiana-s: As I already have a draft PR up for that issue can just delete it here for now

tket/src/serialize/pytket/config/type_translators.rs
tatiana-s: https://github.com/Quantinuum/tket2/issues/1570 kind of deals with that

tket-qsystem/src/helpers.rs
```diff
@@ -0,0 +1,30 @@
+use hugr::extension::prelude::bool_t;
+use hugr::extension::simple_op::MakeRegisteredOp;
+use tket::passes::ReplaceTypes;
+use tket::passes::replace_types::NodeTemplate;
+
+use crate::extension::futures::{FutureOp, FutureOpDef, future_type};
+
+pub(crate) fn replace_types_with_qsystem_defaults() -> ReplaceTypes {
```
acl-cqc: optional nit: maybe just `qsystem_replace_types` or `qsystem_(default_)lowerer` would be a bit shorter, but this is fine. Or even `lowerer_with_future_linearization`

tket/src/passes/replace_types/helpers.rs
```diff
@@ -0,0 +1,178 @@
+//! Common configurations and helper functions for the [ReplaceTypes] pass.
```
acl-cqc: Did you consider putting this in `handlers.rs`? (I'm fine either way)

tket/src/passes/replace_types/helpers.rs
```diff
@@ -0,0 +1,178 @@
...
+                if elem_ty.copyable() {
+                    return Ok(None);
+                }
+
+                let array_ty = type_fn(size, elem_ty);
+                Ok(Some(rt.get_linearizer().copy_discard_op(&array_ty, 2)?))
+            },
+        );
+
+        // `Discard` ops get replaced with a `Drop` op which can be lowered in the
```
acl-cqc: Argh. Sorry, my bad, didn't think this through far enough. I don't think we should have `ReplaceTypes::default()` requiring use of *QSystem* `LowerDropsPass`...can we either

* lower directly to a `rt.get_linearizer().copy_discard_op(...,0)`, as LowerDropsPass does. It is code duplication, but a minimal amount, and parallels what we're already doing for clone just above (with `copy_discard_op(...,2)`)
* Not put this in `ReplaceTypes::default()` after all...I guess that means moving `replace_array_ops_requiring_copyable_bounds` into tket-qsystem `lower.rs` instead

My weak preference is for the former if it works, but I'm not sure whether it does (it would require that any other handlers for types in the array getting dropped are also registered everywhere this is used, as opposed to currently said handlers only being needed in LowerDropsPass - however, perhaps `replace_types_with_qsystem_defaults` does that for us). I'm ok either way.

acl-cqc (approved):
Thanks for changes Tatiana! Afraid I have introduced a problem with one of my previous suggestions, so have written a couple of solutions, but apart from that this looks good to me :).

PR #1561, opened 2026-05-01
fix: Incorrect matching of pytket parameters to unsupported graph ports

### LINKED ISSUE #1516: [Bug]: Invalid hugr produced from `PytketHugrPass`
### Description

A simple `PytketHugrPass` that just does `RemoveRedundancies` on a small hugr succeeds but generates an invalid hugr.

### Steps to Reproduce

The following script constructs the hugr using guppy and then compiles it and attempts to validate the result:

```python
from guppylang import guppy
from guppylang.std.builtins import owned
from guppylang.std.quantum import qubit
from guppylang.std.quantum.functional import ch
from hugr.package import Package
from hugr.passes.scope import GlobalScope
from pytket.passes import RemoveRedundancies
from selene_hugr_qis_compiler import check_hugr
from tket.passes import PytketHugrPass

@guppy
def test(q1: qubit @ owned, q2: qubit @ owned) -> tuple[qubit, qubit]:
    q1, q2 = ch(q1, q2)
    return (q1, q2)

pkg = test.compile_function()
check_hugr(pkg.to_bytes())
h = pkg.modules[0]

p = PytketHugrPass(RemoveRedundancies()).with_scope(GlobalScope.PRESERVE_ENTRYPOINT)
h1 = p(h)
pkg1 = Package([h1])
check_hugr(pkg1.to_bytes())
```

### Expected Behaviour

A valid hugr should be produced.

### Actual Behaviour

The script generates the following error:
```
Traceback (most recent call last):
  File "/home/alec/play/guppylang/comp.py", line 23, in <module>
    check_hugr(pkg1.to_bytes())
    ~~~~~~~~~~^^^^^^^^^^^^^^^^^
selene_hugr_qis_compiler.HugrReadError: in package with generator guppylang (guppylang-internals-v0.32.0)-v0.21.11

Caused by:
    0: Package validation error.
    1: An operation CFG contains invalid children: The BasicBlock Input node of a CFG has a signature of [qubit, qubit, rotation], which differs from the expected type row [qubit, qubit]. In parent Node(25), child Node(26)
    2: The BasicBlock Input node of a CFG has a signature of [qubit, qubit, rotation], which differs from the expected type row [qubit, qubit]
```

### Component

_No response_

### Environment

```markdown
gupplang @ 6af48fd4c612bfab7bf4745900af742c6ee476ec
```

### Additional Context

_No response_

aborgna-q:
Fixes #1516

There was a problem in how we match pytket command parameters that reference a unsupported subgraph output port.
The encoder generated the name using a sequential id for each port, but the decoder used the `PortIndex` instead.

Fixed the bug, added the issue repro as a guppy example, and made an LLM create a reduced rust test case.

Note that while this is a needed bug fix, there's a bigger problem that I'm not fixing in this PR:
Due to how pytket circuits use parameters, each time we find a new unknown variable name the decoder adds it as a new angle input to the region being decoded.
This is useful when decoding user-generated pytket circuits, but in general if the pytket circuit was encoded from a Hugr region there shouldn't be any new inputs added. These new inputs end up making the Hugr invalid instead.
There's different ways we could avoid this, but I'll do that in a follow up PR.


cqc-alec (approved):
Thanks!

PR #1562, opened 2026-05-01
feat!: Ignore empty circuits when encoding Hugr regions into pytket

aborgna-q:
When encoding the regions of a hugr as pytket circuits, `EncodedCircuit::iter` returned a bunch of empty circuits for regions with only unsupported operations.

This was mostly noise, and could reduce performance unless we took care to check if the circuit was empty before doing any expensive optimization call.

BREAKING CHANGE: Added a `EncodeOptions.keep_empty_circuits` option that defaults to filtering-out empty circuits from `EncodedCircuit`. The entrypoint of a Hugr may no longer be in the encoded circuit, even if it was a dataflow region.


tket/src/serialize/pytket/circuit.rs
```diff
@@ -306,7 +306,8 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
             circuits: HashMap::new(),
             opaque_subgraphs: OpaqueSubgraphs::new(0),
         };
-        enc.encode_circuits(hugr, entrypoint, options)?;
+        // Encode the entrypoint circuit, ensuring it's not skipped even if it's empty.
```
acl-cqc: Ok, so this interacts with the options which are provided, potentially with encode_subcircuits: true in which case your EncoderCircuitFilter will affect those too right?

I was going to suggest putting the circuit filter in the `EncodeOptions`, indeed as an arbitrary callback (which here you could augment/override any provided callback to ensure the entrypoint node was always included). However, I note you keep the filter here private, so maybe not. In which case see

tket/src/serialize/pytket/circuit.rs
```diff
@@ -367,6 +369,11 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
                 }
             }
 
+            // Ignore empty circuits, for regions with no supported operation.
+            if !filter_results.filter(&encoded.serial_circuit) {
```
acl-cqc: Hmmm, ok. So the "interesting" bit here is that you don't avoid encoding the circuit, rather you consider discarding it after you've encoded.

This raises the question whether we need to filter the circuits *stored* at all, or whether you might just leave the caller to remove unnecessary elements from `self.circuits` afterwards. I think the answer to that depends on whether `new_standalone_with_entrypoint` really wants to preserve all empty subcircuits not just the entrypoint...

tket/src/serialize/pytket/circuit.rs
```diff
@@ -532,3 +539,22 @@ impl<Node: HugrNode> IndexMut<Node> for EncodedCircuit<Node> {
...
+    All,
+}
+
+impl EncoderCircuitFilter {
+    fn filter(&self, circuit: &SerialCircuit) -> bool {
+        match self {
+            EncoderCircuitFilter::NonEmpty => circuit
+                .commands
+                .iter()
+                .any(|cmd| cmd.op.op_type != tket_json_rs::OpType::Barrier),
```
acl-cqc: Are we not counting barriers because they represent opaque subgraphs? (Sounds like a good reason, but worth a comment.)

But do *all* barriers represent opaque subgraphs? Is there no other reason they could be produced?

tket/src/serialize/pytket/tests.rs
```diff
@@ -445,12 +451,20 @@ fn circ_non_local() -> Hugr {
...
+    let [q] = h.add_dataflow_op(TketOp::H, [q]).unwrap().outputs_arr();
+
     let [q] = {
         let mut dfg = h
             .dfg_builder(Signature::new(inner_input_t, output_t), [q])
             .unwrap();
-        // Rx with non-local input
         let [q] = dfg.input_wires_arr();
+
+        // Dummy quantum function to ensure this circuit gets encoded.
```
acl-cqc: Surely this one here is unnecessary, there is a TketOp::Rx beneath?

tket/src/serialize/pytket/tests.rs
```diff
@@ -607,6 +621,9 @@ fn circ_nested_dfgs() -> Hugr {
     let [qb] = h.input_wires_arr();
     let rot = h.add_load_value(ConstRotation::new(0.5).unwrap());
 
+    // Dummy quantum function to ensure this circuit gets encoded.
```
```suggestion
    // Include quantum op to ensure this circuit gets encoded.
```
It's not a dummy, it does something!!

These are a bit of a PITA, presumably they are because you want to assert the number of circuits is the same in both cases, or something like that?

acl-cqc (commented):
Ok, so there is a bit of a question here...

tket/src/serialize/pytket/tests.rs
aborgna-q: The `Rx` has a non-local angle parameter.

We cannot encode that as a pytket command, so the rx itself is unsupported.

tket/src/serialize/pytket/tests.rs
aborgna-q: The reason is we want to test small hugr that present some edge cases, but if we make them only contain the edge case they end up as empty pytket circuits in most cases.

I guess "dummy" is not the right word for that.

tket/src/serialize/pytket/circuit.rs
aborgna-q: Yeah, the enum was simply to abstract avoid passing a `bool` parameter to the internal `encode_circuit` method.

I think that adding it as an option may be valid though. Will try doing so.

> In which case see

I think you missed a link there?

aborgna-q: I replaced the enum with a boolean flag in `EncodeOptions`.

tket/src/serialize/pytket/circuit.rs
aborgna-q: Not all barriers are opaque ops, sometimes we can translate hugr barriers to bare pytket ones.

But if all commands are barriers, the pytket circuit is still not doing anything.

tket/src/serialize/pytket/tests.rs
```diff
@@ -445,12 +451,20 @@ fn circ_non_local() -> Hugr {
...
         let mut dfg = h
             .dfg_builder(Signature::new(inner_input_t, output_t), [q])
             .unwrap();
-        // Rx with non-local input
         let [q] = dfg.input_wires_arr();
+
+        // Extra quantum op to ensure this circuit gets encoded.
+        let [q] = dfg.add_dataflow_op(TketOp::H, [q]).unwrap().outputs_arr();
+
+        // Rx with non-local input
```
acl-cqc, line 467: Let's just comment that this is unencodable/unsupported/whatever (right?)

tket/src/serialize/pytket/tests.rs
acl-cqc: driveby: should we either
(a) not class this as temporary disablement, but rather, preferred behaviour?
(b) if we really want "encoding of DFG and function calls as CircBoxes" then use a different config for this case so all the other cases can do it?

tket/src/serialize/pytket/tests.rs
```diff
@@ -1142,10 +1183,10 @@ fn fail_on_modified_hugr(circ_tk1_ops: Hugr) {
 #[case::preset_parameterized(circ_parameterized(), 1, CircuitRoundtripTestConfig::Default)]
 #[case::nested_dfgs(circ_nested_dfgs(), 2, CircuitRoundtripTestConfig::Default)]
 #[case::flat_opaque(circ_tk1_ops(), 1, CircuitRoundtripTestConfig::Default)]
-#[case::unsupported_subtree(circ_unsupported_subtree(), 3, CircuitRoundtripTestConfig::Default)]
+#[case::unsupported_subtree(circ_unsupported_subtree(), 0, CircuitRoundtripTestConfig::Default)]
```
acl-cqc, line 1264: driveby nit: I think in general the rust test name includes the names of the parameters, so do you really need to repeat it?

tket/src/serialize/pytket.rs
```diff
@@ -151,6 +151,9 @@ impl TKETDecode for SerialCircuit {
             });
         }
 
+        // Make ruse we don't drop the entrypoint circuit if it's empty.
```
acl-cqc:
```suggestion
        // Make sure we don't drop the entrypoint circuit if it's empty (or any other).
```

tket/src/serialize/pytket/circuit.rs
acl-cqc: Ooops, I meant see https://github.com/Quantinuum/tket2/pull/1562#discussion_r3196723565.

But ok, so IIUC, here we'll respect the preference in the supplied EncodeOptions, so you can delete the comment here.

tket/src/serialize/pytket/circuit.rs
```diff
@@ -367,6 +368,20 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
                 }
             }
 
+            // Ignore empty circuits, for regions with no supported operation.
+            //
+            // A circuit is empty if it only contains barriers representing unsupported HUGR subgraphs, or plain
+            // pytket barriers (as they are effectively no-ops).
```
acl-cqc, line 378: 👍

tket/src/serialize/pytket.rs
acl-cqc: This is the potentially-surprising behaviour, in that we don't respect the supplied `EncodeOptions` here, but it's also a trait impl so there's not really a great place to document it. (We could mention this in the method doc-comment nonetheless, maybe that's the best option available?)

acl-cqc (approved):
Yup thanks Agustin. I think the extra EncodeOptions is fine, and it sounds like you don't mind that "always encoding the entrypoint circuit" means we'll always-encode/not-skip any subcircuits too, so ok - a couple of comments could do with updating but LGTM :)

tket/src/serialize/pytket.rs
aborgna-q: I'll add it to the trait method, as a possible behaviour.

tket/src/serialize/pytket/tests.rs
aborgna-q: Yeah, the hugr builders are function calls so rstest does not register their name.

If we don't name the cases, we end up with
```
case::0 ...
case::1 ...
...
```

tket/src/serialize/pytket/tests.rs
aborgna-q: Yeah, we won't implement that.
I removed the TODO.

PR #1565, opened 2026-05-05
fix: address Miri unsoundness by installing nextest and skipping FFI/long running tests

### LINKED ISSUE #1483: 💥 Unsoundness check fail on main

[Please investigate](https://github.com/quantinuum/tket2/actions/runs/23728773158).

CalMacCQ:
fixes #1483 

Seems we didn't install nexttest in `unsoundness.yml`. Not sure why this worked before?

 Testing whether this resolves the unsoundness check with workflow dispatch here -> https://github.com/Quantinuum/tket2/actions/runs/25375551456/job/74409590210

Turns out we need to skip some tests that use CFFI as MIRI cannot handle those.


.github/workflows/unsoundness.yml
```diff
@@ -44,9 +44,12 @@ jobs:
           cargo miri setup
       - uses: quantinuum/hugrverse-env/install-hugrenv-action@main
 
+      - name: Install nextest
+        uses: taiki-e/install-action@nextest
+
       # Run miri unsoundness checks.
       #
-      # Skip tests that take over 5mins in CI.
+      # Skip tests that take over 5mins in CI or tests which use CFFI/ztd.
```
aborgna-q: ztd-using test need to be ignored rather than skipped, since they require compile-time linking to ztd stuff.
```
    #[cfg_attr(miri, ignore)]
```

aborgna-q (approved):

PR #1580, opened 2026-05-11
feat!: Upgrade hugr dependency to 0.28.0

maximilianruesch:
Prepares an update to the latest (unreleased) version of `hugr`, most importantly with the type-term unification.

BREAKING CHANGE: Updated public dependency `hugr` to `0.28.0`

Not a tket-eccs breaking change:
BEGIN_COMMIT_OVERRIDE
feat: Recompile hugrs with `hugr 0.17.0` (#1580)
END_COMMIT_OVERRIDE


cgh-qtnm:
I believe I will need to merge #1521 into this PR to cleanly upgrade everything to the next HUGR version (due to breaking changes in https://github.com/Quantinuum/hugr/pull/3026). Let me know when this PR is updated to [6742c53](https://github.com/Quantinuum/hugr/commit/6742c53505b2e13c3234dd847a6201759e73b757) and then I will retarget the debug info PR and get it ready to merge.

cgh-qtnm:
#1521 is ready to merge into this branch

maximilianruesch:
@cgh-qtnm We have decided to keep the two PRs separate, and do the minimal amount of work needed to merge the upgrade on its own, to keep two reviewed units (and keep the upgrade and what is essentially a feature addition separate).

Nonetheless, basing your work off this PR is still useful, as it means you can merge right after this PR goes in. Please do not merge your PR into this one.

tket-py/test/test_errors.py
```diff
@@ -26,5 +26,5 @@ def test_unresolved_op() -> None:
     err = str(excinfo.value)
     assert "Could not read CompilationState from bytes" in err
     assert "Error reading package payload in envelope." in err
-    assert "OpaqueOp:unknown.unresolved" in err
+    assert "unknown.unresolved" in err
```
acl-cqc, line 29: Curious. Doesn't just need to be `OpaqueOp::unknown.resolved` or similarly trivial, does it?

tket/src/passes/monomorphize.rs
```diff
@@ -233,7 +233,13 @@ fn escape_dollar(str: impl AsRef<str>) -> String {
 
 fn write_type_arg_str(arg: &TypeArg, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
     match arg {
-        TypeArg::Runtime(ty) => f.write_fmt(format_args!("t({})", escape_dollar(ty.to_string()))),
+        TypeArg::ExtensionType(cty) => {
```
acl-cqc, line 236: Hmmm, should we note this breaking change to name mangling in the notes?

tket/src/passes/replace_types.rs
```diff
@@ -719,16 +718,16 @@ impl ReplaceTypes {
...
-                    TypeEnum::Extension(exty) => match self.consts.get(exty) {
-                        Some(const_fn) => Some(const_fn(e, self)),
-                        None => self
-                            .param_consts
-                            .get(&exty.into())
-                            .and_then(|const_fn| const_fn(e, self).transpose()),
-                    },
-                    _ => None,
-                };
+                let new_const =
```
acl-cqc, line 721: Could get closer to the original by matching on `&*e.get_type()` with cases for `Term::ExtensionType` and `_`, but it's much the same length, so I'm not bothered

tket/src/passes/replace_types.rs
```diff
@@ -1236,10 +1229,10 @@ mod test {
                 .unwrap();
             },
         );
-        fn option_contents(ty: &Type) -> Option<Type> {
+        fn option_contents(ty: Type) -> Option<Type> {
```
acl-cqc: This would probably make more sense taking `&[Term]` (it's only called on the result of `just_elem_type`), and starting:
```
let st = match args {
   [Term::SumType(st)] => st,
   _ => return None
}
```

We could then avoid cloning the first variant by matching against `Term::List` rather than `TypeRow::try_from`, but given this is test code...the problem is clarity rather than amount of cloning

acl-cqc (approved):
Thanks @maximilianruesch and @aborgna-q ! The bulk of the change here is indeed type-term unification, I've removed a good few `clone`s and various carbuncles, there are a couple more in the comments that I could...

tket/src/passes/monomorphize.rs
maximilianruesch: Oh oof. Wait does this mean that HUGRs monomorphise differently, and so old HUGRs will not be linking-compatible with new ones?

tket-py/test/test_errors.py
aborgna-q: This is an artifact of https://github.com/Quantinuum/hugr/pull/3090/changes#diff-b7c397c923b713acd7c5781ca7fbf35cc2da6c84db69abe1e515ed44a69fcf87R87-R97.

We unified the handling of `ExtensionOp` and `OpaqueOp` errors using `OpaqueOp::qualified_name` instead of `OpType::to_string`. Keeping the `OpaqueOp:` suffix seemed unnecessary.

Note that we do not promise stability of error string messages.

tket/src/passes/replace_types.rs
aborgna-q: Updated.

PR #1582, opened 2026-05-12
fix!: Replace non-deterministic iterations on hash maps

aborgna-q:
Enables the `iter_over_hash_type` clippy lint, and fixes all cases of iteration over hashmaps and hashsets that could cause non-deterministic outputs.

BREAKING CHANGE: Changed public field `PartialSum::0` from a `HashMap` to a `BTreeMap`.

### DIFF
--- a/tket/src/passes/dataflow/partial_value.rs
+++ b/tket/src/passes/dataflow/partial_value.rs
@@ -4,8 +4,8 @@ use hugr_core::Node;
 use hugr_core::types::{SumType, Type, TypeArg, TypeEnum, TypeRow};
 use itertools::{Itertools, zip_eq};
 use std::cmp::Ordering;
-use std::collections::HashMap;
-use std::hash::{Hash, Hasher};
+use std::collections::BTreeMap;
+use std::hash::Hash;
 use thiserror::Error;
 
 use super::row_contains_bottom;
@@ -63,14 +63,14 @@ pub struct LoadedFunction<N> {
 
 /// A representation of a value of [`SumType`], that may have one or more possible tags,
 /// with a [`PartialValue`] representation of each element-value of each possible tag.
-#[derive(PartialEq, Clone, Eq)]
-pub struct PartialSum<V, N = Node>(pub HashMap<usize, Vec<PartialValue<V, N>>>);
+#[derive(PartialEq, Clone, Eq, Hash)]
+pub struct PartialSum<V, N = Node>(pub BTreeMap<usize, Vec<PartialValue<V, N>>>);
 
 impl<V, N> PartialSum<V, N> {
     /// New instance for a single known tag.
     /// (Multi-tag instances can be created via [`Self::try_join_mut`].)
     pub fn new_variant(tag: usize, values: impl IntoIterator<Item = PartialValue<V, N>>) -> Self {
-        Self(HashMap::from([(tag, Vec::from_iter(values))]))
+        Self(BTreeMap::from([(tag, Vec::from_iter(values))]))
     }
 
     /// The number of possible variants we know about. (NOT the number
@@ -107,10 +107,6 @@ impl<V: AbstractValue, N: PartialEq + PartialOrd> PartialSum<V, N> {
     /// whether `self` has changed.
     ///
     /// Fails (without mutation) with the conflicting tag if any common rows have different lengths.
-    #[expect(
-        clippy::iter_over_hash_type,
-        reason = "join is pointwise over independent variant tags, so iteration order cannot affect the lattice result"
-    )]
     pub fn try_join_mut(&mut self, other: Self) -> Result<bool, usize> {
         for (k, v) in &other.0 {
             if self.0.get(k).is_some_and(|row| row.len() != v.len()) {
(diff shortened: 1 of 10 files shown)


tket/src/passes/dataflow/partial_value.rs
```diff
@@ -265,6 +273,10 @@ impl<V: Clone, N: Clone> PartialSum<V, N> {
 }
 
 impl<V: PartialEq, N: PartialEq + PartialOrd> PartialOrd for PartialSum<V, N> {
+    #[expect(
+        clippy::iter_over_hash_type,
+        reason = "key iteration only fills presence bitmaps; value comparison below uses numeric tag order"
```
acl-cqc: I wonder if there is some BitVec-like structure that would allow creation via `::from_iter(self.0.keys().map(|k| (k,1))` or something like that but ok

tket/src/passes/dataflow/partial_value.rs
```diff
@@ -302,7 +317,16 @@ impl<V: std::fmt::Debug, N: std::fmt::Debug> std::fmt::Debug for PartialSum<V, N
 
 impl<V: Hash, N: Hash> Hash for PartialSum<V, N> {
     fn hash<H: Hasher>(&self, state: &mut H) {
-        for (k, v) in &self.0 {
+        // Highest tag present in the sum.
+        //
+        // Hashing must be independent of `HashMap` iteration order so that equal
+        // `PartialSum`s always feed the same sequence into the caller's hasher.
+        let max_variant_key = self.0.keys().copied().max().unwrap_or(0);
```
acl-cqc: I have to wonder whether `sorted` (O(n.log(n) in the number of keys present) would be better or worse than what you're doing (O(n) in the max possible *index* of key), but I'm pretty sure it would be a lot lot shorter. Not a strong preference though, fine with this!

tket/src/passes/dataflow/partial_value.rs
```diff
@@ -567,6 +591,10 @@ mod test {
     }
 
     impl TestSumType {
+        #[expect(
+            clippy::iter_over_hash_type,
+            reason = "test validation checks every possible tag independently, so iteration order cannot affect the result"
```
acl-cqc: This is actually pretty easy to rewrite as the `Self::Branch` is a Vec. You would need to check the PartialValue's max key tho, so fine either way.

tket/src/serialize/pytket/encoder/unsupported_tracker.rs
```diff
@@ -91,6 +91,11 @@ impl<N: HugrNode> UnsupportedTracker<N> {
         // and use it here. For now we just traverse all unextracted nodes.
         let mut nodes = BTreeSet::new();
         nodes.insert(node);
+
+        #[expect(
+            clippy::iter_over_hash_type,
+            reason = "iteration only filters membership; nodes are collected into a BTreeSet before any output-affecting use"
```
acl-cqc: Keen to learn more about this exciting new lint I tried:
```
nodes.extend(
            self.nodes
                .iter()
                .filter(|(_, data)| self.components.find_mut(data.component) == representative)
                .map(|(&n, _)| n),
        );
```
and not only was that accepted, but I was surprised to find that the removal loop below (`for n in &nodes { self.nodes.remove(n); }`) didn't set off the lint either, although the original loop adding to `nodes` did (as your `expect` suggests)

tket/src/serialize/pytket/tests.rs
```diff
@@ -152,6 +152,10 @@ const IMPLICIT_PERMUTATION: &str = r#"{
     }"#;
 
 /// Check some properties of the serial circuit.
+#[expect(
```
acl-cqc, line 155: Hmmmm, given the map you are iterating through is only the one that you build here on line 172, +1 for deriving `Ord` for `register::ElementId` in `tket_json_rs` and then just using BTreeMap here. But not in this PR, ok.

tket/src/serialize/pytket/tests.rs
```diff
@@ -192,6 +196,10 @@ fn validate_serial_circ(circ: &SerialCircuit) {
     );
 }
 
+#[expect(
+    clippy::iter_over_hash_type,
+    reason = "test comparison checks command multiplicities; iteration order only affects which mismatch is reported first"
```
acl-cqc, line 201: Yeah, I mean you could just check equality of the HashMap's, or break out the keys, all kinds of things, but fine

acl-cqc (approved):
Looks great, thanks Agustin. You might try [this one](https://github.com/Quantinuum/tket2/pull/1582/changes#r3260698670) but generally fine with all.

tket/src/passes/dataflow/partial_value.rs
aborgna-q: I ended up changing the `HashMap` to a `BTreeMap` instead.
It's a breaking change, since it's a public field, but it lets us simplify all this messy code by just using `Itertools::merge_join_by` (with deterministic order).

tket/src/passes/dataflow/partial_value.rs
aborgna-q: See other comment. Replaced the hashmap with a btreemap.

tket/src/serialize/pytket/encoder/unsupported_tracker.rs
aborgna-q: Weird. I guess the lint analysis has its limits.

I replaced the double iteration with `nodes.extend(self.nodes.extract_if(...))`. The lint also doesn't complain with that. (Though it should :P )

aborgna-q:
Waiting for #1605 to fix CI issues in `main`

PR #1595, opened 2026-05-15
fix: Set inputs of functions loaded as values to top during dataflow analysis

### LINKED ISSUE #1598: [Bug]: Constant folding turns finite loops into infinite loops due to unsoundness
### Description

The constant folding pass turns finite loops into infinite loops, if the loop CFG is affected by unsound values propagated by the inputs. This extends to inputs that, at least visibly in the code, should not affect the loop at all, but in the compiled HUGR they still decide between code paths in the CFG (even though the choice does not matter for actually executed operations).

The original bug occurred with a custom operation that was able to call a function loaded as a value. Constant folding did not recognise this as a call, and thus did not know the input arguments at the callsite, so the input wires of the function were left as BOTTOM (as is the default). However, this is unsound, as the function is still reachable (through the custom op). In fact, any function loaded as a value must be considered reachable, except in cases where we can explicitly prove it isn't.

In the below case, BOTTOM for the input wires propagates into the `if` statement, resulting in a BOTTOM evaluation for its boolean, and thus none of the subsequent basic blocks are reachable. However, since this also includes (transitively) the basic block where the `range` implementation increments its state, constant folding considers that implementation to never increment its state, so it starts inlining the first (and to its eyes only) value, which is the starting parameters. So the range implementation always returns the same value, and if the starting parameters are such that it returns `Some`, the loop is now infinite.

> [!IMPORTANT]
> Constant folding is not only applied when `NormalizeGuppy` is called, but also as part of the `QSystemPass`, that is run as part of the `qis-compiler`: https://github.com/Quantinuum/tket2/blob/97ac3a9cf13b2db7f188df458e1c195820677dbe/qis-compiler/rust/lib.rs#L145-L148

### Steps to Reproduce

With `guppylang==0.21.14`, `selene-sim==0.2.15`, `selene-core==0.2.9`, `hugr==0.16.0` and `tket==245cb2b8dcc6e8001e91ebac60bee7d84d7f6e47` (due to a zig bitcode failure on the latest commit), the following test case hangs infinitely when run, and when cancelling the [selene diagnostic stream](https://github.com/Quantinuum/selene/wiki/Tips-and-Tricks#print-debugging-the-diagnostic-interface) shows that the loop is indeed infinite:
```python
from guppylang import guppy
from guppylang.std.platform import result
from hugr.hugr.render import RenderConfig
from hugr.ops import FuncDefn
from hugr.package import Package
from hugr.tys import Visibility
from selene_helios_qis_plugin import HeliosInterface, LogLevel
from selene_sim import build
from selene_sim.backends.bundled_simulators import Coinflip

def set_visibility(pkg: Package, func_name: str, visibility: Visibility) -> None:
    for hugr in pkg.modules:
        for n, nd in hugr.nodes():
            if isinstance(nd.op, FuncDefn) and nd.op.f_name == func_name:
                nd.op.visibility = visibility


def test_me() -> None:
    @guppy(link_name="i.am.broken")
    def i_am_broken(a: int) -> None:
        for i in range(1):
            if a > 0:
                pass
            result("loop", i)
        result("finish", -1)

    lib = guppy.library(i_am_broken).compile()

    # Set private so it is not recognised as an entrypoint, emulating "callable from unknown"
    set_visibility(lib, "i.am.broken", "Private")

    nm = NormalizeGuppy(constant_folding=True, remove_dead_funcs=False)
    nm(lib.modules[0], inplace=True)

    set_visibility(lib, "i.am.broken", "Public")

    render_conf = RenderConfig(display_node_id=True, max_node_label_length=None)
    lib.modules[0].render_dot(render_conf).view()

    @guppy.declare(link_name="i.am.broken")
    def i_am_broken_decl(a: int) -> None:
        ...

    @guppy
    def main() -> None:
        i_am_broken_decl(5)

    pkg = main.compile()
    pkg = pkg.link(lib)

    runner = build(pkg, interface=HeliosInterface(log_level=LogLevel.DIAGNOSTIC))
    # Hangs infinitely
    results = list(runner.run(Coinflip(), n_qubits=0))

    assert results == [{'finish': [-1], 'loop': [0]}]

```

### Expected Behaviour

Constant folding should not be unsound.

### Actual Behaviour

Constant folding is unsound.

### Component

Rust library

### Environment

```markdown
See reproducer.
```

### Additional Context

_No response_

maximilianruesch:
Any private function that is currently called via an unknown mechanism is currently not assumed to be called, thus violating the open-world-assumption, and leaving the input values on their initialisation value (BOTTOM). This leads to unsound scenarios, such as when a loop body suddenly interferes with the loopiness of the loop itself.

This fix is to, for any function loaded as a value by a `LoadFunction`, set all inputs to TOP, since it may be called from anywhere. Although this is highly overapproximating, and in many cases we do know exactly how a function value will be called, the topmost priority is to be sound and it is a separate issue to follow up on this to increase precision.

Closes #1598


tket/src/passes/dataflow/datalog.rs
```diff
@@ -365,6 +364,17 @@ fn run_datalog<V: AbstractValue, H: HugrView>(
             if let Some(fields) = out_in_row.unpack_first(succ_n, df_block.sum_rows.get(succ_n).unwrap().len()),
             for (out_p, v) in fields.enumerate();
 
+        // Functions ---------------
+        // Any function that is loaded as a value is overapproximated to have possibly arbitrary
```
acl-cqc:
```suggestion
        // Any function that is loaded as a value is overapproximated as being callable with arbitrary
```

tket/src/passes/dataflow/datalog.rs
```diff
@@ -365,6 +364,17 @@ fn run_datalog<V: AbstractValue, H: HugrView>(
...
+        // inputs.
+        out_wire_value(inp, p, PV::top()) <--
+            node(n),
+            if hugr.get_optype(*n).is_load_function(),
+            out_wire_value(n, OutgoingPort::from(0), v),
```
acl-cqc, line 377: I *think*
```suggestion
            out_wire_value(n, OutgoingPort::from(0), PartialValue::LoadedFunction(LoadedFunction {func_node: func, ..})),
```
works, certainly some patterns are allowed

tket/src/passes/dataflow/datalog.rs
```diff
@@ -365,6 +364,17 @@ fn run_datalog<V: AbstractValue, H: HugrView>(
...
+            if let PartialValue::LoadedFunction(LoadedFunction { func_node: func, .. }) = v,
```
acl-cqc, line 382: I mean, it'd be good to unwrap/expect here, but of course we have to defend against the LoadFunction not having been processed yet i.e. initially we probably just have Bottom

acl-cqc (approved):
You can dismiss my approval if you'd rather have someone else's ;)

tket/src/passes/dataflow/datalog.rs
acl-cqc: Not sure if it's worth starting an issue to discuss the problems here (in case anyone tries an "obvious" fix that is actually unsound), or noting more in the comment.

maximilianruesch: I extended a comment somewhat

tket/src/passes/dataflow/datalog.rs
maximilianruesch: does not work on my machine

tket/src/passes/dataflow/datalog.rs
maximilianruesch: I can add an assertion that expects this

aborgna-q (approved):
👍 

There was a CI issue on `main`.  It's fixed now, so you'll need to merge or rebase.

PR #1602, opened 2026-05-18
fix(qis-compiler): trim trailing NUL from public bitcode

qartik:
`compile_to_bitcode()` was returning the raw bytes from `write_bitcode_to_memory().as_slice()`, which includes LLVM's implicit trailing NUL in `inkwell 0.9`. That makes the public `.bc` payload file-incompatible and matches the `Invalid bitcode signature` failure discussed in Slack.

Regression occurred as follows:
- The `inkwell 0.9` bump happened in `hugr` PR https://github.com/Quantinuum/hugr/pull/3039
- That change was then included in `hugr` release `0.27.1`, via release PR https://github.com/Quantinuum/hugr/pull/3014
- `tket2` picked it up later in #1568

This PR keeps the internal LLVM buffer behavior unchanged but trims the extra terminator from the public byte payload returned by `compile_to_bitcode()`.

## Changes
- add a small helper in `qis-compiler/rust/lib.rs` to convert an LLVM memory buffer into public bitcode bytes
- use that helper in `compile_to_bitcode()` instead of returning the raw buffer slice directly
- add a regression test that compiles a fixture HUGR, writes the returned bytes to a real `.bc` file, reparses it through LLVM, and checks that the public bytes match the raw LLVM buffer without the implicit terminator

## Validation
- `cargo test -p selene-hugr-qis-compiler`

### DIFF
--- a/qis-compiler/rust/lib.rs
+++ b/qis-compiler/rust/lib.rs
@@ -210,6 +210,19 @@ fn optimize_module(module: &Module, args: &CompileArgs) -> Result<()> {
     Ok(())
 }
 
+/// Copy LLVM bitcode into a public byte buffer.
+///
+/// LLVM's in-memory bitcode writer appends an implicit trailing NUL byte. That
+/// terminator is required for some in-process LLVM APIs but must not be exposed
+/// in the public bitcode payload.
+fn public_bitcode_bytes(memory_buffer: &inkwell::memory_buffer::MemoryBuffer<'_>) -> Vec<u8> {
+    let bytes = memory_buffer.as_slice();
+    match bytes.last() {
+        Some(0) => bytes[..bytes.len() - 1].to_vec(),
+        _ => bytes.to_vec(),
+    }
+}
+
 fn get_entry_point_name(namer: &Namer, hugr: &impl HugrView<Node = Node>) -> Result<String> {
     const HUGR_MAIN: &str = "main";
     let (name, entry_point_node) = if hugr.entrypoint_optype().is_module() {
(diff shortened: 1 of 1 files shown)


qis-compiler/rust/lib.rs
```diff
@@ -479,6 +492,56 @@ mod selene_hugr_qis_compiler {
...
+    use std::{
+        fs,
+        time::{SystemTime, UNIX_EPOCH},
+    };
+    use tket::hugr::llvm::inkwell::{
+        context::Context, memory_buffer::MemoryBuffer, module::Module,
+    };
+
+    fn parse_bitcode_as_file(bitcode: &[u8]) -> Result<Module<'static>, String> {
+        let file_name = format!(
```
cqc-alec, line 511: Could alternatively use https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html .

cqc-alec (approved):

qartik:
Will wait for @hsemenenko to confirm this fixes the issue he reported before merging.

qis-compiler/rust/lib.rs
qartik: @cqc-alec tried but the CI didn't like it. Might have to revert that commit.

cqc-alec: OK. Weird.

hsemenenko:
@qartik I've tested this branch with `guppylang` main and it has resolved the issue.

aborgna-q:
Looks like this PR broke CI -.-
https://github.com/Quantinuum/tket2/actions/runs/26047247824/job/76574580537#step:8:633

cqc-alec:
> Looks like this PR broke CI -.- https://github.com/Quantinuum/tket2/actions/runs/26047247824/job/76574580537#step:8:633

Weird, this seems to be an intermittent failure.

cqc-alec:
(It occurred previously on this branch, and we thought it was because of the commit that used `NamedTempFile` so @qartik reverted that. But maybe it was unrelated...)

PR #1603, opened 2026-05-18
feat: Allowing control flow in `control`

### LINKED ISSUE #1552: [Feature]: Control with control flow
### Problem Statement

Now, https://github.com/Quantinuum/tket2/blob/e61d6e7bba2794f41621e1903b7b7ae53e4982fc/tket/src/modifier/modifier_resolver.rs#L1049-L1050 

prevents us from solving modifiers on control flow, we should allow with control

nicolaassolini-qntm:
close #1552 

- Enabled control modifiers on multi-block CFGs. When `dagger` is present, only control flow-free CFGs are allowed.
- Unified CFG rewriting through `modify_signature`-based control type handling, so single-block and multi-block CFGs follow the same rebuild path.
- Adjusted TailLoop control placement to match the data/control layout used by modified bodies.
- Added modifier examples to the resolver regression coverage, including a control-flow-inside-control-modifier case and control-flow-inside-nested-control-dagger example.
- Added short comments around CFG/control layout, `LoadFunction` copying, and call-map remapping.


tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -253,16 +253,32 @@ impl<N: HugrNode> ModifierResolver<N> {
...
     ) -> Result<(), ModifierResolverErrors<N>> {
-        // TODO:
-        // Indirect calles would be handled by its caller.
-        // However, when a loaded function is used in the other ways
-        // (e.g., passed to higher-order functions as `map` or `fold`),
-        // we need to modify it here.
+        let consumers = h.linked_inputs(n, 0).collect::<Vec<_>>();
+
+        // Check if all consumers are modifiers. If so, we can just forget the LoadFunction node and let the modifiers rebuild it.
+        if !consumers.is_empty()
```
cqc-alec, line 264: Do we actually need the non-emptiness check? If there are no callers we can still forget it...

cqc-alec (approved):

PR #1605, opened 2026-05-19
ci: Fix CI build error for qis-compiler tests

aborgna-q:
#1606 added rust unit tests to `qis-compiler`. That is a crate that uses `pyo3` and links against `libpython`, so CI started failing due to not being able to link to python.
This is a quick patch to that by whitelisting which packages get tested where. We could do something more thorough by fully building the python project before running `cargo test`, but we need to evaluate how does that impact CI runtimes.

- `pyo3`'s extension-module feature is deprecated, we can remove its uses. https://pyo3.rs/main/features#extension-module
- Updated `uv` versions used in CI
- Whitelisted cargo test runs to avoid compiling `qis-compiler`, since it requires running maturin with a fully setup uv package environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ env:
   SCCACHE_GHA_ENABLED: "true"
   RUSTC_WRAPPER: "sccache"
   # Pinned version for the uv package manager
-  UV_VERSION: "0.9.7"
+  UV_VERSION: "0.11.15"
   UV_FROZEN: 1
   # The highest and lowest supported Python versions, used for testing
   PYTHON_HIGHEST: "3.14"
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -162,10 +162,12 @@ jobs:
         run: rustup override set ${{steps.toolchain.outputs.name}}
       - uses: quantinuum/hugrverse-env/install-hugrenv-action@main
       - uses: taiki-e/install-action@nextest
+      # qis-compiler requires libpython to link test binaries;
+      # it is tested separately in the tests-qis-compiler job.
       - name: Build with all features
-        run: cargo nextest r --verbose --workspace --all-features --no-run
+        run: cargo nextest r --verbose -p badger-optimiser -p compile-rewriter -p tket -p tket1-passes -p tket-py -p tket-qsystem -p tket-qec --all-features --no-run
```
cqc-alec: Can we use `--exclude` instead of listing all except one (here and elsewhere)?

aborgna-q: Changed the whitelist for a blacklist instead

cqc-alec (approved):

PR #1607, opened 2026-05-19
ci: Create py releases as draft until wheels are built

aborgna-q:
Sets the `draft: true` option of release-please so python gh releases are first created as `draft` while the wheels are being built. We mark it as ready only after submitting to pypi.

I tried this once before but had problems with github not triggering workflows on draft releases. The solution is to set `force-tag-creation: true` in release-please, and use the tag as a trigger instead.

Additionally:
- Added provenance attestations to the built wheels.
- We now upload the wheels to the release artifacts in addition to Pypi.
- Splits the `python-pure-wheels` workflow so only the actual publishing is run inside the `release` environment. We do this already in the other two workflows.

This is a better solution that what we do now in `hugr`; creating a full release and marking it as draft at the start of the wheels building workflow. I'll migrate this new config once we verify it works correctly.

### DIFF
--- a/.github/workflows/python-pure-wheels.yml
+++ b/.github/workflows/python-pure-wheels.yml
@@ -40,18 +40,10 @@ jobs:
           RUN_EXTS: ${{ github.ref_type != 'tag' || startsWith(github.ref, 'refs/tags/tket-exts-v') }}
           RUN_ECCS: ${{ github.ref_type != 'tag' || startsWith(github.ref, 'refs/tags/tket-eccs-v') }}
 
-  build-publish:
-    name: Package and publish wheels
+  build:
+    name: Build and test wheels
     needs: check-tag
     runs-on: ubuntu-latest
-    environment: release
-    permissions:
-      # IMPORTANT: this permission is mandatory for trusted publishing
-      id-token: write
-      # Used to upload release artifacts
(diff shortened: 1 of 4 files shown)


.github/workflows/python-pure-wheels.yml
```diff
@@ -98,18 +93,55 @@ jobs:
           uv run -f dist --with ${{ matrix.target.name }} --refresh-package ${{ matrix.target.name }} --no-project -- python -c "import ${{ matrix.target.name }}"
           uvx twine check --strict dist/*
 
+  release:
+    name: Release
+    needs: [check-tag, build]
+    runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'tag' && startsWith(github.ref, 'refs/tags/tket-eccs-v') || github.ref_type == 'tag' && startsWith(github.ref, 'refs/tags/tket-exts-v') }}
```
cqc-alec: Do we need more parentheses? Or refactor as: `${{ github.ref_type == 'tag' && (startsWith(github.ref, 'refs/tags/tket-eccs-v') || startsWith(github.ref, 'refs/tags/tket-exts-v')) }}`?

cqc-alec (approved):

Issue #1060, opened 2025-08-21
Remove Circuit struct


lmondada:
@aborgna-q first floated the idea.

I first thought it was extreme, but I'm a big proponent of it now. In the tket2 APIs I never know if I should take in a Hugr or a Circuit -- they really are the same thing.

PR #1611, opened 2026-05-20
refactor!: Deprecate commands iterator

aborgna-q:
The `CommandsIterator` was one of the first utilities defined in this repo, and is quite focused on pytket-style flat circuits with basic quantum operations, qubit wires, and bit wires.
This definition grew stale quickly, but it was used throughout the repo nonetheless.

This PR removes such uses and marks the definition as `deprecated`. We should be able to remove it on `tket 0.20.0`.

I added a `toposorted_children` and a `count_ops` method to `Circuit` to replace commands usage.
These will be moved to an extension trait in the future as part of #1060.

BREAKING CHANGE: `Circuit::circuit_cost` now computes the cost of all descendant nodes of a circuit, not just `TketOp`s.

### DIFF
--- a/tket-py/src/state/cost.rs
+++ b/tket-py/src/state/cost.rs
@@ -4,6 +4,7 @@ use std::cmp::Ordering;
 use std::iter::Sum;
 use std::ops::{Add, AddAssign, Sub};
 
+use pyo3::types::PyAnyMethods;
 use pyo3::{Py, PyAny, PyResult, Python, pyclass, pymethods};
 use tket::circuit::cost::{CircuitCost, CostDelta};
 
@@ -31,11 +32,22 @@ impl Default for PyCircuitCost {
     }
 }
 
+/// Returns whether `cost` is the internal no-cost sentinel (a python `None` value).
+fn is_no_cost(cost: &Py<PyAny>, py: Python<'_>) -> bool {
+    cost.bind(py).is_none()
+}
+
 impl Add for PyCircuitCost {
     type Output = PyCircuitCost;
 
     fn add(self, rhs: PyCircuitCost) -> Self::Output {
         Python::attach(|py| {
+            if is_no_cost(&self.cost, py) {
+                return rhs;
+            }
+            if is_no_cost(&rhs.cost, py) {
+                return self;
+            }
             let cost = self
                 .cost
                 .call_method1(py, "__add__", (rhs.cost,))
@@ -48,6 +60,13 @@ impl Add for PyCircuitCost {
 impl AddAssign for PyCircuitCost {
     fn add_assign(&mut self, rhs: Self) {
         Python::attach(|py| {
+            if is_no_cost(&rhs.cost, py) {
+                return;
+            }
+            if is_no_cost(&self.cost, py) {
+                self.cost = rhs.cost;
+                return;
+            }
             let cost = self
                 .cost
                 .call_method1(py, "__add__", (rhs.cost,))
@@ -62,6 +81,9 @@ impl Sub for PyCircuitCost {
 
     fn sub(self, rhs: PyCircuitCost) -> Self::Output {
         Python::attach(|py| {
+            if is_no_cost(&rhs.cost, py) {
+                return self;
+            }
             let cost = self
                 .cost
                 .call_method1(py, "__sub__", (rhs.cost,))
@@ -76,6 +98,9 @@ impl Sum for PyCircuitCost {
         Python::attach(|py| {
             let cost = iter
                 .fold(None, |acc: Option<Py<PyAny>>, c| {
+                    if is_no_cost(&c.cost, py) {
+                        return acc;
+                    }
                     Some(match acc {
                         None => c.cost,
                         Some(cost) => cost
--- a/tket/src/circuit.rs
+++ b/tket/src/circuit.rs
@@ -318,7 +318,7 @@ impl<T: HugrView<Node = Node>> Circuit<T> {
     #[inline]
     #[deprecated(
         since = "0.19.0",
-        note = "This is a limited API that will be dropped soon. Use toposorting over `hugr.scheduling_graph()` instead."
+        note = "This is a limited API that will be dropped soon. Use toposorting over `HugrView::scheduling_graph` instead.\n<https://docs.rs/hugr/latest/hugr/trait.HugrView.html#method.scheduling_graph>"
     )]
     #[expect(deprecated)]
     pub fn commands(&self) -> CommandIterator<'_, T>
@@ -339,7 +339,7 @@ impl<T: HugrView<Node = Node>> Circuit<T> {
     #[inline]
     #[deprecated(
         since = "0.19.0",
-        note = "This is a limited API that will be dropped soon. Use toposorting over `hugr.scheduling_graph()` instead."
+        note = "This is a limited API that will be dropped soon. Use toposorting over `HugrView::scheduling_graph` instead.\n<https://docs.rs/hugr/latest/hugr/trait.HugrView.html#method.scheduling_graph>"
     )]
     #[expect(deprecated)]
     pub fn operations(&self) -> impl Iterator<Item = Command<'_, T>> + '_
--- a/tket/src/circuit.rs
+++ b/tket/src/circuit.rs
@@ -10,6 +10,7 @@ use std::collections::HashSet;
 use std::iter::Sum;
 
 use crate::passes::utils::hash::{HashError, HugrHash};
+#[expect(deprecated)]
 pub use command::{Command, CommandIterator};
 use hugr::extension::prelude::{NoopDef, TupleOpDef};
 use hugr::extension::simple_op::MakeOpDef;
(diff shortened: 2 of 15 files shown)


tket-py/src/state/cost.rs
```diff
@@ -31,11 +32,22 @@ impl Default for PyCircuitCost {
     }
 }
 
+/// Returns whether `cost` is the internal no-cost sentinel (a python `None` value).
+fn is_no_cost(cost: &Py<PyAny>, py: Python<'_>) -> bool {
+    cost.bind(py).is_none()
+}
```
maximilianruesch, line 38: How is this change related to deprecating the commands iterator?

tket/tests/badger_termination.rs
```diff
@@ -64,5 +64,6 @@ fn badger_termination(simple_circ: Circuit, nam_4_2: ECCBadgerOptimiser) {
             ..Default::default()
         },
     );
-    assert_eq!(opt_circ.commands().count(), 11);
+
+    assert_eq!(opt_circ.count_ops(|op| op.cast::<TketOp>().is_some()), 4);
```
maximilianruesch, line 67: How is `4` correct? Not saying that I know how `11` was correct either, given that the number of commands in the above array is `9`. Either way this should probably be explained in a short comment if it can.

tket/tests/badger_termination.rs
maximilianruesch: Can we do a driveby and remove this dead code?

tket/src/ops.rs
```diff
@@ -393,11 +397,6 @@ pub(crate) mod test {
         h.unwrap()
     }
 
-    #[rstest]
-    fn check_t2_bell(t2_bell_circuit: Circuit) {
-        assert_eq!(t2_bell_circuit.commands().count(), 2);
-    }
-
```
maximilianruesch, line 400: Does this mean the above fixture is unused? Why not leave the test in?

tket/src/optimiser/badger.rs
```diff
@@ -520,21 +520,35 @@ pub use badger_default::{DefaultBadgerStrategy, ECCBadgerOptimiser};
...
+    use crate::extension::rotation::RotationOp;
     use crate::serialize::load_tk1_json_str;
     use crate::serialize::pytket::DecodeOptions;
     use crate::{Circuit, TketOp};
     use crate::{extension::rotation::rotation_type, optimiser::badger::BadgerOptions};
 
     use super::{BadgerOptimiser, ECCBadgerOptimiser};
 
+    /// Returns a vector of `TketOp` nodes in the entrypoint region, in topological order.
+    fn entrypoint_tket_ops(circ: &Circuit) -> Vec<Node> {
```
maximilianruesch:
```suggestion
    /// Returns a vector of nodes in the entrypoint region that are either `TketOp` or `RotationOp`, in topological order.
    fn entrypoint_tket_or_rotation_ops(circ: &Circuit) -> Vec<Node> {
```
Seems outdated given the implementation

tket/src/optimiser/badger.rs
```diff
@@ -607,25 +621,22 @@ mod tests {
...
-        let [op1, op2]: [&OpType; 2] = opt_rz
-            .commands()
-            .map(|cmd| cmd.optype())
-            .collect::<Vec<_>>()
-            .try_into()
+        let [op1, op2]: [&OpType; 2] = entrypoint_tket_ops(&opt_rz)
+            .into_iter()
+            .map(|node| opt_rz.hugr().get_optype(node))
+            .collect_array()
             .unwrap();
```
maximilianruesch: You should either not need the type annotation, or not need `collect_array` (and instead use the standard `collect`).

tket/src/passes/utils/chunks.rs
```diff
@@ -553,8 +557,19 @@ mod test {
...
-        let h = reassembled.commands().next().unwrap().node();
+        let ops = reassembled
+            .toposorted_children(reassembled.parent())
+            .expect("circuit entrypoint should be dataflow region")
+            .filter(|&node| {
+                reassembled
+                    .hugr()
+                    .get_optype(node)
+                    .cast::<TketOp>()
+                    .is_some()
```
maximilianruesch, line 568: It seems that filtering after a specific optype cast (or multiple) is a common operation. Can we provide a trait that is implemented for all collections holding hugr nodes?

tket/src/circuit.rs
```diff
@@ -314,6 +316,11 @@ impl<T: HugrView<Node = Node>> Circuit<T> {
     ///
     /// Ignores the Input and Output nodes.
     #[inline]
+    #[deprecated(
+        since = "0.19.0",
+        note = "This is a limited API that will be dropped soon. Use toposorting over `hugr.scheduling_graph()` instead."
```
maximilianruesch:
```suggestion
        note = "This is a limited API that will be dropped soon. Use toposorting over `HugrView::scheduling_graph()` instead."
```
or whichever trait offers this function ... Ideally this would be a valid / clickable reference (not saying that it cannot be right now, I would just wonder how it would be).

tket/src/passes/commutation.rs
```diff
@@ -6,58 +6,64 @@ use derive_more::{Display, Error, From};
...
-    circuit::command::Command,
     ops::{Pauli, TketOp},
+    resource::{ResourceId, ResourceScope},
 };
 
-type Qb = crate::circuit::units::LinearUnit;
+type Qb = ResourceId;
 
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
 // remove once https://github.com/quantinuum-DEV/tket2/issues/126 is resolved
```
maximilianruesch:
```suggestion
// remove once https://github.com/quantinuum/tket2/issues/126 is resolved
```
Driveby opportunity to fix a link here. Also, please spend 5 minutes investigating if this can be easily done (I would be willing to do it as well).

tket/src/passes/commutation.rs
```diff
@@ -66,33 +72,54 @@ type SliceVec = Vec<Slice>;
...
-        slice[q.index()] = Some(com.clone());
+        slice[q.as_usize()] = Some(com.clone());
+    }
+}
+
+fn ensure_resource_slot(
+    qubit_free_slice: &mut Vec<usize>,
+    slices: &mut SliceVec,
+    resource: ResourceId,
+) {
```
maximilianruesch, line 83: I have difficulty gauging what this function does. Could you add a comment / make this (especially the name) clearer?

maximilianruesch (changes requested):

tket-py/src/state/cost.rs
aborgna-q: There's actually a breaking change in this PR: `Circuit::circuit_cost` now uses `entry_descendants` instead of `commands()`.

That means we're calculating the cost of a bunch of other node types, not only `TketOp`.
The problem here was that `PyCircuitCost` returns `None` as the cost of unknown nodes, and we weren't handling that.

I added the breaking notice to the PR description.

tket/tests/badger_termination.rs
aborgna-q: This is the circuit after optimization. It seems before we were counting every non-IO node. It makes more sense to count tket ops instead.

```mermaid
---
config:
    layout: elk
---
graph LR
    subgraph 0 ["(0) Module"]
        direction LR
        subgraph 1 ["(1) [**FuncDefn: #quot;#quot;**]"]
            direction LR
            2["(2) Input"]
            3["(3) Output"]
            5["(5) const:custom:a(π*0.1)"]
            6["(6) LoadConstant"]
            10["(10) const:custom:a(π*0.2)"]
            11["(11) LoadConstant"]
            16["(16) const:custom:a(π*-0.1)"]
            17["(17) LoadConstant"]
            22["(22) tket.quantum.H"]
            23["(23) tket.quantum.Rz"]
            24["(24) tket.quantum.H"]
            13["(13) tket.rotation.radd"]
            32["(32) tket.quantum.Rz"]
            2--"0:0<br>qubit"-->32
            2--"1:0<br>qubit"-->22
            2--"2:2<br>qubit"-->3
            5--"0:0<br>rotation"-->6
            6--"0:1<br>rotation"-->13
            10--"0:0<br>rotation"-->11
            11--"0:1<br>rotation"-->23
            16--"0:0<br>rotation"-->17
            17--"0:0<br>rotation"-->13
            22--"0:0<br>qubit"-->23
            23--"0:0<br>qubit"-->24
            24--"0:1<br>qubit"-->3
            13--"0:1<br>rotation"-->32
            32--"0:0<br>qubit"-->3
        end
    end
    ```

tket/src/ops.rs
aborgna-q: It's used by a test in `commutation`. I moved it there instead.

tket/src/passes/commutation.rs
aborgna-q: This module is quite likely to be deprecated next, I wouldn't really spend time trying to update it.

tket/src/passes/utils/chunks.rs
maximilianruesch: Out of scope for now, will revisit later perhaps.

maximilianruesch (approved):

PR #1621, opened 2026-05-23
perf: Improve modifier resolver pass by visiting nodes in a set

### LINKED ISSUE #1586: [Performance]: Investigate why ModifierResolverPass is so slow
See #1587

jpacold:
Fixes #1586 by replacing a `Vec` used to hold visited nodes with a `FxHashSet`. On my machine this reduces the wall-clock time for the slow test from ~25s to ~1s for a debug build and from ~1.4s to ~0.06s for a release build.


tket/src/modifier/modifier_resolver.rs
```diff
@@ -1269,11 +1270,11 @@ pub fn resolve_modifier_with_entrypoints_and_scope(
     // generate nodes that are not reachable from the entry points.
     // If more thorough cleanup is needed, we should run dead code elimination.
     let mut deletelist = entry_points.clone();
-    let mut visited = vec![];
+    let mut visited = FxHashSet::default();
     while let Some(node) = deletelist.pop_front() {
```
aborgna-q, line 1342: -.-

aborgna-q (approved):
Nice catch, thanks for the contribution!

PR #1623, opened 2026-05-26
fix: skip test_nested_array to resolve miri unsoundness

### LINKED ISSUE #1622: 💥 Unsoundness check fail on main

[Please investigate](https://github.com/quantinuum/tket2/actions/runs/26385260253).

CalMacCQ:
I think the previous unsoundness checks were timing out until https://github.com/Quantinuum/tket2/pull/1565 was merged. Now I think this additonal test has shown up as something which fails the unsoundness check. 

I think an extra skip is needed here. See https://github.com/Quantinuum/tket2/issues/1622#issuecomment-4542319079

closes #1622

### DIFF
--- a/tket/src/passes/borrow_squash.rs
+++ b/tket/src/passes/borrow_squash.rs
@@ -493,7 +493,7 @@ mod test {
     }
 
     #[rstest]
-    #[cfg_attr(miri, ignore)] // Opening files is not supported in (isolated) miri
+    #[cfg_attr(miri, ignore)] // Opening compressed HUGR files is not supported in (isolated) miri
     fn test_nested_array() {
         let inner_array_type = BorrowArray::ty(5, qb_t());
         let outer_array_type = BorrowArray::ty(3, inner_array_type.clone());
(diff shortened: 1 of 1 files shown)


tket/src/passes/borrow_squash.rs
```diff
@@ -493,6 +493,7 @@ mod test {
     }
 
     #[rstest]
+    #[cfg_attr(miri, ignore)]
```
cqc-alec:
```suggestion
    #[cfg_attr(miri, ignore)] // Opening files is not supported in (isolated) miri
```

cqc-alec (approved):

tket/src/passes/borrow_squash.rs
```diff
@@ -493,6 +493,7 @@ mod test {
     }
 
     #[rstest]
+    #[cfg_attr(miri, ignore)] // Opening files is not supported in (isolated) miri
```
aborgna-q: Technically we're not opening files (since we're using `include_bytes!`), but trying to call zstd to decompress the Hugr.

```suggestion
    #[cfg_attr(miri, ignore)] // Opening compressed HUGR files is not supported in (isolated) miri
```

PR #1624, opened 2026-05-26
perf: Avoid string formatting in Hugr op hashing

waamm:
Tackles a TODO by @ABorgna: replace `hashable_op` string/JSON formatting with structural hashing via `hash_op_type`

The old flow was: build a string representing the op, then hash that string. In this PR `hash_op_type` writes directly into the hasher.

Absolute hash values change but discrimination should be at least as good - this version is stricter about including the extension id.

Needed a small workaround: `FmtHasher` streams `Display` into the hasher because `OpType::name()` isn't public outside `hugr`.

`cargo bench -p tket --bench bench_main "hash a simple circuit"`  suggests a 1.5x speedup on an M4 Max.


aborgna-q (approved):
Nice. Thanks for the contribution!

PR #1628, opened 2026-05-27
refactor: Updated dependencies in modifier test files

### LINKED ISSUE #1667: [Refactor]: Update dependencies in tests after the new guppy release
Update dependencies in tests after the new guppy release by removing the references to the temporary branches. Delete the guppy branches too.

nicolaassolini-qntm:
closes #1667


tatiana-s (approved):
Does it not work with only adding the `guppylang` dependency, do you need internals as well? But approved otherwise, also good reminder that I need to do this to the other tests too

nicolaassolini-qntm:
> Does it not work with only adding the `guppylang` dependency, do you need internals as well? But approved otherwise, also good reminder that I need to do this to the other tests too

uv was failing to find the proper internal version. I can try again without the internals dependecy

aborgna-q (approved):

Issue #1234, opened 2025-11-10
Pytket encoder/decoder improvements


aborgna-q:
The main work for the decoder has been completed, but there remains additional features and bugs to work on.
This Tracking Issue is a recollection of them.

### Bugs

- [ ] https://github.com/CQCL/tket2/issues/938
- [x] #1294
- [x] #1297
- [x] #1303

### Features

- [ ] #1237
- [ ] https://github.com/CQCL/tket2/issues/1135
- [ ] https://github.com/CQCL/tket2/issues/627
- [ ] Allow `PytketTypeTranslator`s to define custom translations between types ([TODO](https://github.com/CQCL/tket2/blob/bd4bd710bf4b83fb34bdd94754e5239ed40bd88b/tket/src/serialize/pytket/config/type_translators.rs#L155-L156))

### Optional Features

- [ ] Share the decoder's `RegisterHash` ids with the general use `ResourceScope`? ([TODO](https://github.com/CQCL/tket2/blob/bd4bd710bf4b83fb34bdd94754e5239ed40bd88b/tket/src/serialize/pytket/decoder/tracked_elem.rs#L61-L62))

### Performance

- [ ] Implement iteration over a component's elements for petgraph's `UnionFind` ([TODO](https://github.com/CQCL/tket2/blob/bd4bd710bf4b83fb34bdd94754e5239ed40bd88b/tket/src/serialize/pytket/encoder/unsupported_tracker.rs#L91-L92))

PR #1631, opened 2026-05-27
feat: Encode measurements followed by reads as pytket measurements

### LINKED ISSUE #1570: [Feature]: Insert reads in the pytket decoder
### Problem Statement

After the change to measurements in https://github.com/Quantinuum/tket2/pull/1558, measure ops will no longer supported in the pytket decoder, in order to support them again we should decode them to a measure followed by a read - this would be good (necessary for backwards compatibility?) for `load_pytket` in guppy (loading pytket circuits with measurements without this would compile but not run), however since we also generally want to encourage measurements to happen in guppy itself in order to take advantage of linearity checks we might also not want to do this (would need to be documented well)

### Proposed Solution

_No response_

### Component

Rust library

### Effort Estimate

### Additional Context

_No response_

### LINKED ISSUE #1572: [Feature]: Encode measurements followed by reads as pytket measurements
### Problem Statement

#1558 disabled encoding of Hugr `Measurement`s into pytket ones, since now we always return a lazy measurement type.

Measurement ops are still relevent for some pytket optimisations (specially the GreedyPauliSimp family).
While we can no longer produce pytket measurements for all hugr measurements, we should be able to detect cases where the measurement is immediately followed with a read (without any control flow in the middle), and produce the pytket command.

This issue may be implemented together with the encoding part from #1570.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

tatiana-s:
Closes #1572 and technically #1570 (though instead of inserting read turns out you can only convert back to a measure anyway)


aborgna-q:
The `circuit_standalone_roundtrip::case7_preset_bits` errors is due to missing information on the rountrip.

We're encoding a Hugr that has multiple outputs with the same bit value. For normal roundtrip, that information is tracked in the `EncodedCircuit` context.

For standalone circuits we only have the `pytket` circuit to carry the information, and we do not encode all of the context there. There's a task in #1234 to add that capability, but it's really low priority.

I'd just remove the case from `circuit_standalone_roundtrip`. We test that circuit in `encoded_circuit_roundtrip`, which is the path that's actually being used.

tatiana-s:
Closing this in favour of https://github.com/Quantinuum/tket2/pull/1658 to avoid dealing with messy merge conflicts

PR #1634, opened 2026-05-28
fix: Modifier passes bugs with indirect call node

### LINKED ISSUE #1633: [Bug]: Higher order functions with modifiers
### Description

We have some panics when we run programs like:

```python

@guppy(dagger=True, control=True)
def apply_dagger(f: Unitary[[qubit], None], q: array[qubit, 3]) -> None:
    f(q[1])


@guppy
def main() -> None:
    q = array(qubit(), qubit(), qubit())
    h(q[0])

    with dagger:
        apply_dagger(s, q)


```

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
close #1633 

- Resolve higher-order function modifiers when the callee is loaded from a function value, including controlled and daggered cases.
- Propagate modified function types through nested containers such as `Conditional`, `CFG`, `DataflowBlock`, and `TailLoop` so carried function values keep the right signature.
- Update modifier resolution for direct calls, indirect calls, `Tag` nodes, and loop/branch boundaries to avoid type mismatches during reconstruction.
- Refresh the modifier example suite and generated Hugr artifacts for higher-order array, loop, and recursive examples.


test_files/modifier_examples/higher_order_function_w_arrays.py
```diff
@@ -0,0 +1,56 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "guppylang @ git+https://github.com/Quantinuum/guppylang.git@na/1637-allow-unitary-flag-with-higher-order-function#subdirectory=guppylang",
```
nicolaassolini-qntm: This will be changed after the corresponding guppy pr will be merged and released

test_files/run_modifier_examples/run_hugrs.py
```diff
@@ -13,9 +13,13 @@
 import numpy.typing as npt
 from hugr import Hugr
 from guppylang.emulator import EmulatorBuilder
+from hugr.hugr.render import RenderConfig
+
 
 sys.path.append(str(Path(__file__).resolve().parents[1]))
 
+SAVE_SOLVED_HUGRS = False
```
cqc-alec: I think it would be better to have this controlled from an environment variable, or some other means that doesn't require editing the code. (If using an environment varialble maybe it should point to the directory we want to use, rather than have this hardcoded.)

I'd consider removing it altogether, though, if this is just for debugging.

test_files/modifier_examples/higher_order_classical.py
```diff
@@ -0,0 +1,64 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "guppylang @ git+https://github.com/Quantinuum/guppylang.git@na/1637-allow-unitary-flag-with-higher-order-function#subdirectory=guppylang",
```
cqc-alec: Ideally we should wait for that branch to be merged. If not, please make an issue to fix this reference once that's happened.

(Same comment applies to other files in this directory.)

test_files/modifier_examples/higher_order_classical.py
```diff
@@ -0,0 +1,64 @@
...
+# ]
+# ///
+"""A simple controlled gate using modifiers"""
```
cqc-alec: Looks like this docstring has been repeated across all files in this directory.

tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -38,14 +40,108 @@ impl<N: HugrNode> ModifierResolver<N> {
...
         let new_call_fn_port = new_call.called_function_port();
         let new_call_node = new_dfg.add_child_node(new_call);
 
         self.call_map_insert(new_callee, (new_call_node, new_call_fn_port));
+        let skip_inputs = input_modifiers
+            .iter()
+            .map(|(input, _)| *input)
+            .collect::<HashSet<_>>();
+
+        // Handle one function arguments that must be modified before calling `new_call_node`.
```
cqc-alec:
```suggestion
        // Handle function arguments that must be modified before calling `new_call_node`.
```
?

tket/src/modifier/modifier_resolver.rs
```diff
@@ -877,9 +1175,33 @@ impl<N: HugrNode> ModifierResolver<N> {
             (new_node, new_node),
             (inputs, outputs),
             (input_offset, output_offset, new_offset),
+            &HashSet::new(),
         )
     }
 
+    // fn wire_node_inout_skipping_inputs<'a>(
```
cqc-alec: Remove commented code.

tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -180,6 +280,11 @@ impl<N: HugrNode> ModifierResolver<N> {
         indir_call: &CallIndirect,
         new_dfg: &mut impl Dataflow,
     ) -> Result<(), ModifierResolverErrors<N>> {
+        if !self.signature_has_quantum_data(&indir_call.signature) {
+            self.add_node_no_modification(h, n, indir_call.clone(), new_dfg)?;
```
cqc-alec: Can we add a test for this case?

tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -191,20 +296,39 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+            self.modify_fn(h, func)?
+        } else if self.control_num() == 0 {
+            *self.modifiers_mut() = modifiers;
+            let new_call = self.add_node_no_modification(h, n, indir_call.clone(), new_dfg)?;
+            let new_load = new_dfg.add_child_node(load.clone());
+            self.call_map_insert(func, (new_load, load.function_port()));
+            new_dfg.hugr_mut().connect(new_load, 0, new_call, 0);
+            return Ok(());
+        } else {
+            self.wrap_fn_with_controls(h, func, &load.type_args)?
```
cqc-alec: Can we add tests to cover these cases?

cqc-alec (commented):
Made some suggestions but I admit I don't understand the logic terribly well. A few more tests wouldn't go amiss.

tket/src/modifier/modifier_resolver/call_modify.rs
nicolaassolini-qntm: `test_files/modifier_examples/higher_order_classical.py` is the test. I also add it to https://github.com/Quantinuum/tket2/blob/7e4cbc15ad7d48e9d3c68eb528f06e665778e7f7/tket/src/modifier/modifier_resolver.rs#L2456. I can add also a specific rust test

test_files/modifier_examples/higher_order_classical.py
nicolaassolini-qntm: yes, I'm wainting for the guppy pr and release to be able to update the dependencies

tket/src/modifier/modifier_resolver/call_modify.rs
nicolaassolini-qntm: I refactored this function a bit and added a test case when the function does not need to be modified

nicolaassolini-qntm:
After the merge with the main, I add a temporary guppy branch and change the dependency of the files to point to that branch. I plan to change it to depend on `ts/future-measure` or a new guppy release before merging the PR.

cqc-alec (approved):
Thanks!

PR #1644, opened 2026-06-01
feat!: Raise an error when `power` modifier is present

### LINKED ISSUE #1615: [Feature]: Raise an Error if power modifier is present
### Problem Statement

Up to now, `power` is silently ignored. Until we've implemented the power passes, we should raise an error.

nicolaassolini-qntm:
close #1615 

Power modifier is now explicitly not supported, and attempting to use it will result in a clear error. 

* Updated the `CombinedModifier::push` method in `tket/src/modifier.rs` to return a `PowerModifierNotSupported` error when encountering the power modifier, instead of silently toggling a flag.
* Added a new error variant `PowerModifierNotSupported` to the `ModifierResolverErrors` enum in `tket/src/modifier/modifier_resolver.rs`.
* Updated `modify_indirect_call` in `tket/src/modifier/modifier_resolver/call_modify.rs` to properly wrap error messages.
* Updated `ModifierErrors` docstring
* Added `test_files/guppy_examples/use_of_power.py` as a test example of power use. 
* Added a test to ensure that when a hugr contains the power modifier, the proper error is raised.

BREAKING CHANGE: made `ModifierResolverErrors` `non_exhaustive`
BREAKING CHANGE: made `ModifierError` `non_exhaustive`
BREAKING CHANGE: changed visibility of`CombinedModifier::push` and `CombinedModifier` from public to private and added the field `node: N`


tket/src/modifier/modifier_resolver.rs
```diff
@@ -356,7 +356,7 @@ impl<N> Default for ModifierResolver<N> {
     }
 }
 
-/// Error that can occur when resolving modifiers.
+/// Errors that can occur when tracing validating chain of modifiers and their target.
 #[derive(Debug, derive_more::Error, derive_more::Display)]
 pub enum ModifierError<N = Node> {
```
aborgna-q: Since we are breaking already,

```suggestion
/// Errors that can occur when tracing validating chain of modifiers and their target.
#[derive(Debug, derive_more::Error, derive_more::Display)]
#[non_exhaustive]
pub enum ModifierError<N = Node> {
```

tket/src/modifier/modifier_resolver.rs
aborgna-q, line 397:
```suggestion
#[non_exhaustive]
pub enum ModifierResolverErrors<N = Node> {
```

tket/src/modifier/modifier_resolver.rs
```diff
@@ -420,6 +420,9 @@ pub enum ModifierResolverErrors<N = Node> {
     /// The node cannot be modified.
     #[display("Modification by {_0:?} is not defined for the node {_1}")]
     Unimplemented(Modifier, OpType),
+    /// The power modifier is not supported.
+    #[display("Power modifier is not supported")]
+    PowerModifierNotSupported,
```
aborgna-q: Other errors specify the node that failed to be converted. Add it here too.

aborgna-q (approved):
A couple small changes, but LGTM otherwise

tket/src/modifier/modifier_resolver.rs
nicolaassolini-qntm: Done it. I re-requested the review just for a final quick look

aborgna-q (approved):

PR #1658, opened 2026-06-05
feat: Encode measurements followed by reads as pytket measurements

### LINKED ISSUE #1570: [Feature]: Insert reads in the pytket decoder

### LINKED ISSUE #1572: [Feature]: Encode measurements followed by reads as pytket measurements

tatiana-s:
Copy of https://github.com/Quantinuum/tket2/pull/1631 that avoids dealing with messy merge conflicts 

Closes https://github.com/Quantinuum/tket2/issues/1572
Closes https://github.com/Quantinuum/tket2/issues/1570


tket/src/serialize/pytket/extension/tket.rs
```diff
@@ -72,16 +78,28 @@ impl TketOpEmitter {
...
+            // Since the node does not have outputs the qubit/bit will simply be
+            // ignored, but will appear when collecting the final pytket registers.
+            // However, as `MeasureFree` returns a `Measurement` type, we need to
+            // check whether it is directly followed by a `Read` turning it into a
+            // bool before we can translate it.
             TketOp::MeasureFree => {
-                // Don't translate `Measure` ops for now (the `Measurement` type they
-                // they return is currently unsupported).
-                return Ok(EncodeStatus::Unsupported);
+                let is_read = hugr.node_outputs(node).any(|out_port| {
```
aborgna-q: Why are we checking every output port?
We should only check the first output, no?

This would give invalid results if there was an order edge from the measurement to an unrelated read operation.

aborgna-q (approved):

PR #1661, opened 2026-06-08
chore: Ignore IO tests in soundness checks and improve test runtime

### LINKED ISSUE #1660: 💥 Unsoundness check fail on main

[Please investigate](https://github.com/quantinuum/tket2/actions/runs/27118373327).

aborgna-q:
Fixes https://github.com/Quantinuum/tket2/issues/1660.

drive-by: Add `--no-fail-fast` so we get all errors at once next time.
drive-by: Add more slow tests to the skip list.

Run before skips (1h 41min): https://github.com/Quantinuum/tket2/actions/runs/27147141561/job/80127920500
Run after skips (55min): https://github.com/Quantinuum/tket2/actions/runs/27153371724/job/80149847542

### DIFF
--- a/.github/workflows/unsoundness.yml
+++ b/.github/workflows/unsoundness.yml
@@ -28,8 +28,6 @@ env:
   LD_LIBRARY_PATH: "/tmp/hugrverse/lib:/tmp/hugrverse/lib64"
   LLVM_SYS_211_PREFIX: "/tmp/hugrverse"
 
-
-
 jobs:
   miri:
     name: "Miri"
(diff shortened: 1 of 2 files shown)


.github/workflows/unsoundness.yml
```diff
@@ -52,7 +49,11 @@ jobs:
       # Skip tests that take over 5mins in CI.
```
cqc-alec: Is this comment still accurate? It doesn't seem consistent with the number of new tests we're skipping (17): 17x5min = 1h25min which is more than the time shaved.
Suggest a lower (and approximate) threshold.

cqc-alec (approved):

.github/workflows/unsoundness.yml
aborgna-q: Yeah, it should be 60s. Though it's a noisy filter; when adding to the list I left most <90s.

PR #1668, opened 2026-06-11
chore: Bump hugr patch versions

aborgna-q:


maximilianruesch (commented):
Unless this is required for tket itself, this is not strictly necessary. Any downstream dependency like guppylang can steer the required hugr version itself.

Or is this because we package hugr into the tket-py wheel and thus need it to be up to date?

aborgna-q:
`tket-py` ships wheels with the specified dependency bundled in, it _needs_ the updated dependency with the bug fixes.

Also in general, requiring rust dependency non-breaking bumps is not an issue. If some downstream user really wants to use a buggy versions cargo will just compile both.

maximilianruesch (approved):

PR #1670, opened 2026-06-11
feat!: Remove deprecated definitions

aborgna-q:
Remove deprecated definitions before the new breaking release.

I'm leaving deprecations that were introduced on the previous release, only removing older code.

Also, adds `since` tags to some deprecation messages missing them.

BREAKING CHANGE: Removed deprecated definitions.

### DIFF
--- a/tket-exts/src/tket_exts/tket/qsystem/__init__.py
+++ b/tket-exts/src/tket_exts/tket/qsystem/__init__.py
@@ -23,7 +23,7 @@ __all__ = [
 
 
 class QSystemExtension(TketExtension):
-    """Deprecated (since 0.2.7): use :class:`QSystemHeliosExtension` or :class:`QSystemSolExtension` instead.
+    """Deprecated (since 0.13.0): use :class:`QSystemHeliosExtension` or :class:`QSystemSolExtension` instead.
 
     The combined ``tket.qsystem`` extension has been split into platform-specific
     extensions. Use ``tket_exts.qsystem_helios`` or ``tket_exts.qsystem_sol`` instead.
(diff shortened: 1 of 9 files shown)


tket/src/passes.rs
```diff
@@ -31,7 +31,6 @@ pub use dead_funcs::{RemoveDeadFuncsError, RemoveDeadFuncsPass};
 
 // Force a topological order on nodes.
 pub mod force_order;
-pub use force_order::{force_order, force_order_by_key};
```
aborgna-q, line 34: There is no way to deprecate re-exports, but this one here was shadowing the module definition so it needed to go.

tket-exts/src/tket_exts/tket/qsystem/__init__.py
```diff
@@ -23,7 +23,7 @@
 
 
 class QSystemExtension(TketExtension):
-    """Deprecated: use :class:`QSystemHeliosExtension` or :class:`QSystemSolExtension` instead.
+    """Deprecated (since 0.2.7): use :class:`QSystemHeliosExtension` or :class:`QSystemSolExtension` instead.
```
cqc-alec: Why this version number? Is this the upcoming tket-exts release? The current latest is 0.12.3 it seems ...

aborgna-q: Yes, it's the upcoming release: https://github.com/Quantinuum/tket2/pull/1505/changes

aborgna-q: Oh, wait. There's something iffy going on here
image

aborgna-q: Ok, changed it to `since 0.13.0`. I'll fix release-please to use that version.

cqc-alec (approved):

PR #1672, opened 2026-06-11
feat: init pg-libs and migrating pg-core

yao-cqc:
This PR migrates the first crate `pg-core`, which contains a IR for Pauli graph/quantum circuit, some core types, and a `PGPass` trait.


pg-libs/Cargo.toml
aborgna-q, line 1: Nested workspaces are not really supported ([rust-lang/cargo#5042](https://www.github.com/rust-lang/cargo/issues/5042)), we should use the root workspace instead.

That will also ease the integration between `pg-libs` and `tket` later on.

pg-libs/pg-core/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "pg-core"
+version = "0.1.0"
+edition = "2024"
+
+[dependencies]
+serde = { version = "1.0.228", features = ["derive"] }
+serde_json = "1"
```
aborgna-q: Use a single source of truth for the dependency versions at the workspace level,


```suggestion
serde = { workspace = true, features = ["derive"] }
serde_json = { workspace = true }
```

pg-libs/pg-core/Cargo.toml
```diff
@@ -0,0 +1,11 @@
+[package]
+name = "pg-core"
+version = "0.1.0"
+edition = "2024"
```
aborgna-q: Missing some labels (see `tket`'s Cargo.toml).
Some of these can be defined in the workspace,


```suggestion
edition = { workspace = true }
rust-version = { workspace = true }

license = { workspace = true }
readme = "README.md"
documentation = "https://docs.rs/pg-libs"
homepage = { workspace = true }
repository = { workspace = true }
description = "..."
keywords = ["Quantum", "Quantinuum"]
categories = ["compilers", "science::quantum-computing"]
```

pg-libs/justfile
aborgna-q, line 1: I think it's a good idea to have a specific justfile here, though we should consider which rules can be shared too.

If moving to the single root workspace, we'll need to specify the pg-libs packages explicitly here.

Note also that we have a `pre-commit` hook at the repo level.
We'll need to edit it so it runs relevant checks (and it doesn't try to test `tket` when only pg-libs gets modified)

pg-libs/pg-core/README.md
```diff
@@ -0,0 +1,8 @@
+# pg-core
+Core IR definitions for the `pg-libs` optimisation stack.
+This crate provides:
+
+- A `PauliGraph` IR that also serves as a quantum circuit representation, supporting common circuit operations.
+- The `PGPass` trait, which defines passes over a `PauliGraph`.
+
+`pg-core` provides the core types and pass trait used throughout the stack.
```
aborgna-q: It would be a good idea to link to DEVELOPMENT.md, the CHANGELOG, and docs.rs.

For now, make sure to add the license,


```suggestion
`pg-core` provides the core types and pass trait used throughout the stack.


## License

This project is licensed under Apache License, Version 2.0 ([LICENSE][] or http://www.apache.org/licenses/LICENSE-2.0).

  [LICENSE]: https://github.com/quantinuum/tket2/blob/main/LICENCE
```

pg-libs/justfile
```diff
@@ -0,0 +1,25 @@
+test-rust:
+    cargo test --workspace
+
+test-rust-all-features:
+    cargo +nightly-2025-09-14 test --workspace --all-features
```
cqc-alec: Any reason to pin to this nightly?

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
+use serde::{Deserialize, Serialize};
+
+/// The supported gate types that can appear in a Pauli graph.
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum GateType {
+    // TQE gates
+    /// A $X$ controlled $X$ gate. i.e. H(0);CX;H(0)
```
cqc-alec:
```suggestion
    /// An $X$-controlled $X$ gate, i.e. H(0);CX(0,1);H(0)
```
If we're going to include qubit indices let's do it for all the ops. (Same applies to others in this file.)

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    XX,
+    /// A $X$ controlled $Y$ gate. i.e. H(0);CY;H(0)
```
cqc-alec:
```suggestion
    /// An $X$-controlled $Y$ gate, i.e. H(0);CY(0,1);H(0)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    XY,
+    /// A $X$ controlled $Z$ gate. i.e. CX(1,0)
```
cqc-alec:
```suggestion
    /// An $X$-controlled $Z$ gate, i.e. CX(1,0)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    XZ,
+    /// A $Y$ controlled $X$ gate. i.e. V(0);CX;Vdg(0);
```
cqc-alec:
```suggestion
    /// A $Y$-controlled $X$ gate, i.e. V(0);CX(0,1);Vdg(0)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    YX,
+    /// A $Y$ controlled $Y$ gate. i.e. V(0);CY;Vdg(0);
```
cqc-alec:
```suggestion
    /// A $Y$-controlled $Y$ gate, i.e. V(0);CY(0,1);Vdg(0)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    YY,
+    /// A $Y$ controlled $Z$ gate. i.e. CY(1,0);
```
cqc-alec:
```suggestion
    /// A $Y$-controlled $Z$ gate, i.e. CY(1,0)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    YZ,
+    /// A $Z$ controlled $X$ gate. i.e. CX
```
cqc-alec:
```suggestion
    /// A $Z$-controlled $X$ gate, i.e. CX(0,1)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    ZX,
+    /// A $Z$ controlled $Y$ gate. i.e. CY
```
cqc-alec:
```suggestion
    /// A $Z$-controlled $Y$ gate, i.e. CY(0,1)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    ZY,
+    /// A $Z$ controlled $Z$ gate. i.e. CZ
```
cqc-alec:
```suggestion
    /// A $Z$-controlled $Z$ gate, i.e. CZ(0,1)
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    Z,
+    // Special gates
+    /// A computational-basis measurement gate.
+    /// args: [qubit, cbit]
+    Measure,
+    /// A qubit reset gate.
+    Reset,
+    /// A swap gate.
+    SWAP,
+    /// An opaque gate with implementation-defined behavior.
```
cqc-alec:
```suggestion
    /// An opaque gate with implementation-defined behaviour.
```

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+    RY,
+    /// A rotation around the $Z$ axis.
+    RZ,
+    /// A $ZZ$ phase gate.
+    ZZPHASE,
+    /// A phased-$X$ gate.
+    PHASEDX,
+}
+
+pub(crate) fn gate_type_n_args(gate_type: &GateType) -> Option<usize> {
```
cqc-alec, line 75: Please add docstring.

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+        | GateType::RY
+        | GateType::RZ
+        | GateType::PHASEDX => Some(1),
+        GateType::Measure => Some(2),
+        GateType::Reset => Some(1),
+        GateType::BlackBox => None,
+    }
+}
+
+pub(crate) fn gate_type_n_params(gate_type: &GateType) -> Option<usize> {
```
cqc-alec: Docstring?

pg-libs/pg-core/src/gates.rs
```diff
@@ -0,0 +1,122 @@
...
+        | GateType::V
+        | GateType::Vdg
+        | GateType::X
+        | GateType::Y
+        | GateType::Z => Some(0),
+        GateType::RX | GateType::RY | GateType::RZ | GateType::ZZPHASE => Some(1),
+        GateType::PHASEDX => Some(2),
+        GateType::Measure => Some(0),
+        GateType::Reset => Some(0),
+        GateType::BlackBox => Some(0),
```
cqc-alec: Since these all return `Some(*)` (even `BlackBox`?), does it need to be an `Option`?

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
+use serde::{Deserialize, Serialize};
+
+use crate::{ConditionalBoxData, GateType, Op, gate_type_n_args, gate_type_n_params};
+
+/// A single-qubit Pauli operator.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[repr(u8)]
+pub enum Pauli {
+    /// The $X$ Pauli operator.
+    X = 0,
```
cqc-alec, line 11: I think I'd have made I=0, but no big deal :-)

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
...
+    /// The $Y$ Pauli operator.
+    Y = 1,
+    /// The $Z$ Pauli operator.
+    Z = 2,
+    /// The identity operator.
+    I = 3,
+}
+
+/// A pg-core program consisting of a qubit count and an ordered list of operations.
+/// - All Op::Gate operations have valid parameter counts for their gate type.
```
cqc-alec: What is the context for these promises? As they don't appear to be guaranteed by the type definition, it's worth expanding on where the responsibility lies for keeping them, and how they are checked.

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    /// Inserts a single operation at the provided index.
+    pub fn insert_op(&mut self, index: usize, op: Op) {
+        validate_op(&op, self.get_n_qubits());
+        self.ops.insert(index, op);
+    }
+
+    /// Appends all operations from another graph with the same qubit count.
+    pub fn extend(&mut self, mut other: PauliGraph) {
+        debug_assert_eq!(
```
cqc-alec: Why is this only a debug assert? Shouldn't we check this in release builds? Also I don't think it should be an assert since it is a public function and the condition may be triggered by third-party code.

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
...
+
+    /// Adds an operation with classical conditions, merging with the last matching box when possible.
+    pub fn add_conditional_op(
+        &mut self,
+        op: Op,
+        conditional_bits: Vec<usize>,
+        conditional_values: Vec<bool>,
+    ) {
+        if conditional_bits.len() != conditional_values.len() {
+            panic!("Length of conditional bits and values must match");
```
cqc-alec, line 164: I think this `panic` (as well as other `panic`s in this file) would be better handled as errors, if you can find a way to do that.

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
...
+        Self {
+            n_qubits,
+            ops: vec![],
+        }
+    }
+
+    /// Replaces the operation list and returns the updated graph.
+    pub fn with_ops(mut self, ops: Vec<Op>) -> Self {
+        for op in &ops {
+            validate_op(op, self.get_n_qubits());
```
cqc-alec, line 89: All these calls to `validate_op()` panic or assert on validation failure. I think errors would be better, but at least we should document the behaviour.

pg-libs/pg-core/src/graph.rs
```diff
@@ -0,0 +1,383 @@
...
+            for inner_op in data.get_ops() {
+                validate_op(inner_op, pg_nqubits);
+            }
+        }
+        Op::SetBoundary => {}
+    }
+}
+
+fn validate_op_arg_range(maxq: usize, n_qubits: usize, op: &Op) {
+    assert!(
```
cqc-alec: Assert should be reserved for cases where failure indicates an error of logic in the current module; I don't think that's the case here.

pg-libs/pg-core/README.md
cqc-alec, line 1: Add newline at end of file.

pg-libs/pg-core/Cargo.toml
cqc-alec, line 1: Add newline at end of file.

pg-libs/justfile
cqc-alec, line 1: Add newline at end of file.

pg-libs/Cargo.toml
cqc-alec, line 1: Add newline at end of file.

cqc-alec (commented):
Thanks Yao. Left a few comments. Main general concern is error handling. Can we do better than panic when a user passes incompatible ops to the graph methods?

pg-libs/pg-core/src/graph.rs
yao-cqc: In a later crate I use base 3 encoding for XX,XY,XZ,...,ZZ gates so that's why I is the last. This enum technically doesn't have to follow that order, but I still assign it this way just to be consistent. :)

pg-libs/Cargo.toml
yao-cqc: I notice `--workspace` is used across various CI jobs, pre-commit hooks, and justfile commands. Should we exclude `pg-libs` crates from some of these? `pg-libs` already has 12 crates and will likely grow... So using `--exclude` per crate isn't ergonomic. Any ideas?

cqc-alec: I think we could start by including `pg-libs` crates, and if this becomes annoying we can rethink?

pg-libs/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[workspace]
+resolver = "3"
+members = [
+    "pg-core",
+    ]
+
```
aborgna-q, line 6: Since we are not using the root workspace, we need to bring some definitions here.


```suggestion

[workspace.package]
rust-version = "1.91"
edition = "2024"
homepage = "https://github.com/quantinuum/tket2"
repository = "https://github.com/quantinuum/tket2"
license = "Apache-2.0"

```

pg-libs/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+[profile.test]
+opt-level = 1
+
+[workspace.lints.clippy]
+type_complexity = "allow"
+too_many_arguments = "allow"
```
aborgna-q: Bring this in line with the rest of the repo.

(Have not checked it, so may need some `[expect(...)]`s.


```suggestion
[workspace.lints.rust]
unexpected_cfgs = { level = "warn", check-cfg = [
    # Set by our CI
    'cfg(ci_run)',
    # Set by codecov
    'cfg(coverage,coverage_nightly)',
] }
missing_docs = "warn"

[workspace.lints.clippy]
allow_attributes = "warn"
iter_over_hash_type = "warn"
```

.github/workflows/ci.yml
aborgna-q, line 1: Please add entries to the `/.pre-commit-config` too.

We can filter the rust files so only the relevant checks are ran on each commit.

aborgna-q (approved):
Ok, let's keep the separate workspaces now to get things rolling.

But we should re-consider the repo organization later on.

aborgna-q:
We also should add a new entry to `.github/CODEOWNERS`.
Preferably create a github team for it (as we do with `quantinuum/tket2`).

pg-libs/justfile
yao-cqc: I’ve removed this for now. It’s required by a crate I’ll add later.

yao-cqc:
Hi @cqc-alec, @aborgna-q and I have agreed to keep pg-libs in a separate workspace for now. We can integrate it more tightly once most of the migration is complete.

I’ve also addressed the error-handling issue: importing invalid serialised data now returns errors, while constructing a PauliGraph with invalid Ops will panic.

cqc-alec (approved):

PR #1678, opened 2026-06-11
fix: Add missing extensions in `tket-py`'s registry

aborgna-q:
We forgot to add these the new qsystem extensions to the registry used by `tket-py`.


maximilianruesch (approved):

maximilianruesch:
Perhaps consider opening an issue instead of a TODO

PR #1692, opened 2026-06-15
feat: Export extension registries from tket and tket-qsystem

### LINKED ISSUE #1679: [Feature]: Share common extension registry lists between `tket`, `tket-qsystem`, and `tket-py`.
### Problem Statement

Each crate has a private `REGISTRY` which lists the extensions they have access to.
This is prone to getting out of sync (see https://github.com/Quantinuum/tket2/pull/1678).

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

aborgna-q:
Centralizes the multiple extension registry definitions used around the crates:

- `tket` now exports a public `tket_extensions` list as well as a `REGISTRY` with those + prelude extensions.
- `tket-qsystem` exports a `qsystem_extensions` and a registry with `qsystem_extensions + tket_extensions + prelude`.
- The rest of the crates use those definitions, instead of redefining the lists.

Fixes an issue where `qis-compiler`'s registry was missing `tket.measurement`.
Fixes #1679.
Adds a test to check that `tket_exts` is in sync with these lists (it's currently missing `TKET1`, but I'll fix that in another PR).

### DIFF
--- a/qis-compiler/rust/lib.rs
+++ b/qis-compiler/rust/lib.rs
@@ -27,20 +27,11 @@ use std::path::PathBuf;
 use std::rc::Rc;
 use std::vec::Vec;
 use std::{fs, str, vec};
-use tket::extension::{TKET_EXTENSION, TKET1_EXTENSION, rotation};
-use tket::hugr::extension::{ExtensionRegistry, prelude};
-use tket::hugr::std_extensions::arithmetic::{
-    conversions, float_ops, float_types, int_ops, int_types,
-};
-use tket::hugr::std_extensions::{collections, logic, ptr};
 use tket::hugr::{self, llvm::inkwell};
 use tket::hugr::{Hugr, HugrView, Node};
 use tket::llvm::rotation::RotationCodegenExtension;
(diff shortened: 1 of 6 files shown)


doug-q (commented):
It's not entirely obvious to me that we should do this.

`qis-compiler` doesn't support all extensions, so shouldn't it use a registry with only those extensions it supports?

There is another layer of "extension checking" where we register llvm lowerings for all the supported ops. Perhaps that is enough.

cqc-alec (approved):
LGTM modulo Doug's comment.

aborgna-q:
> qis-compiler doesn't support all extensions, so shouldn't it use a registry with only those extensions it supports?

Even if the extensions are not supported, failing at load time due to missing extensions is not a nice UX.

We use the same set of extensions around the tket-verse, and `CompilationState` in python explicitly avoids bundling them into packages to speed up communication. That causes loading errors here if extensions 

We should rather ensure that we have checks and good errors for unsupported ops in the passes themselves.

PR #1696, opened 2026-06-15
feat: Option to include/omit tket registers in CompilerState.to_bytes

aborgna-q:
`CompilerState.to_bytes` / `.to_str` encodes a package created from the hugr and includes any extension not in the prelude or in tket's `bundled_extensions`.

This is done to avoid the extra encoding/decoding overhead when passing hugrs across libraries that already have access to the tket extensions, but sometimes we may want encode a portable hugr envelope with all the extensions.

I kept this as default-omit since it's the existing behaviour. We can discuss wether it should include everything by default or not.

### DIFF
--- a/tket-py/src/state/base.rs
+++ b/tket-py/src/state/base.rs
@@ -77,13 +77,22 @@ impl CompilationState {
     /// Encode the circuit as a HUGR envelope.
     ///
     /// If no config is given, it defaults to the default binary envelope.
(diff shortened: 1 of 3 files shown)


cqc-alec (approved):
No strong opinion on what the default value should be. Perhaps if it weren't breaking I'd slightly lean towards making it "false", but only slightly.

PR #1706, opened 2026-06-17
fix: Deduplicate lowering replacement functions by using Visibility::Public

### LINKED ISSUE #1704: [Bug]: Lowering replacement functions are private, causing duplicates when the same op appears multiple times
## Description

When lowering `TketOp`s or cross-platform ops (e.g. `SolOp::PhasedXX` → Helios, `HeliosOp::ZZPhase` → Sol), each occurrence of the same op that requires a multi-op decomposition function inserts a **duplicate** `FuncDefn` into the HUGR rather than reusing a single shared one.

## Steps to Reproduce

1. Build a HUGR containing two or more instances of the same multi-op `TketOp` (e.g. `TketOp::H`) or cross-platform op (e.g. two `SolOp::PhasedXX`).
2. Call `lower_tk2_ops` targeting either platform.
3. Inspect the resulting HUGR for `FuncDefn` nodes with the replacement function name (e.g. `__tk2_helios_h` or `__tk2_sol_to_helios_phasedxx`).

Reproducer tests added in [`7a1277a1`](https://github.com/quantinuum/tket2/commit/7a1277a11d4ecde0a8fe5a1d5f0b5cca78ab6f2a) on branch `fix/dedup-lowering-functions`:
- `test_duplicate_tket_ops_produce_single_replacement_func` (two `H` ops)
- `test_lower_all_ops_to_helios` / `test_lower_all_ops_to_sol` (fixture with two `PhasedXX` and two `ZZPhase`)

## Expected Behaviour

Each unique op decomposition function is emitted exactly once into the HUGR, with all call sites pointing to that single definition.

## Actual Behaviour

A new `FuncDefn` is inserted for every occurrence of the op. E.g. with two `H` gates:

```
assertion `left == right` failed: Expected exactly one __tk2_helios_h replacement function
  left: 2
 right: 1
```

## Root Cause

`build_func`, `build_helios_op_for_sol`, and `build_sol_op_for_helios` all use `FunctionBuilder::new(...)`, which creates a `Visibility::Private` `FuncDefn`. The `NameLinkingPolicy` with `OnMultiDefn::UseTarget` used in `func_as_node_template` only deduplicates **public** functions — private ones bypass the name-lookup entirely and are always re-inserted as fresh copies.

## Fix

Use `FunctionBuilder::new_vis(..., Visibility::Public)` for all three builder functions so the linking policy can deduplicate them on subsequent insertions. Fixed in `fix/dedup-lowering-functions`.

## Component

- tket-qsystem

ss2165:
Replacement functions built for lowering (TketOp decompositions and
cross-platform PhasedXX/ZZPhase decompositions) were created as
Visibility::Private. The NameLinkingPolicy with OnMultiDefn::UseTarget
only deduplicates public FuncDefns; private ones are always re-inserted
as fresh copies.

Use FunctionBuilder::new_vis(..., Visibility::Public) so that when the
same op appears multiple times, insert_link_hugr reuses the existing
definition instead of adding a duplicate.

Closes #1704

### DIFF
--- a/devenv.lock
+++ b/devenv.lock
@@ -3,11 +3,11 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1776245669,
-        "narHash": "sha256-vK5IG4QgfbtYhIo0KlHgTJmDbHZCYU1ELIbHoK/Ap94=",
+        "lastModified": 1781627264,
+        "narHash": "sha256-TPj5d5MUyvuQZjsfDAAoYJ0SB+tNNNBrdCpD8XL0WeU=",
         "owner": "cachix",
         "repo": "devenv",
-        "rev": "99d2072320398af5ce4eaefb11f71e03f0f6d9a8",
+        "rev": "0fe5629a2955141c336b95e384e2c793c01214fb",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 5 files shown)


ferbetanzo (approved):
Thanks!

maximilianruesch (commented):
Should the functions that are now public (and are also public in the final product I presume) be privatised at the end of the procedure?

ss2165:
@maximilianruesch Qsystempass cleans up at the end anyway: https://github.com/Quantinuum/tket2/blob/4209df1130f092d0936de71154bc98c78745a2ac/tket-qsystem/src/lib.rs#L138


Also I've just added a simple TODO refactor commit, no change in functionality (using a library function in HUGR that we had previously copy pasted)

maximilianruesch (approved):

ss2165:
Have added a snapshot test to show that without this fix multiple LLVM definitions were being inserted (often hidden by inlining), and no longer are.

PR #1708, opened 2026-06-17
feat: Add modifier passes to benchmarks

### LINKED ISSUE #1642: Add modifier sample to the benchmarks
Include here: https://github.com/Quantinuum/tket2/blob/7e4cbc15ad7d48e9d3c68eb528f06e665778e7f7/tket/benches/benchmarks a file that loads some small modifier example to use during benchmarking

nicolaassolini-qntm:
closes #1642

This PR adds modifier passes to benchmarks, benchmarking the passes both on a hugr with modifier and without modifiers


tket/benches/benchmarks/modifier_resolver.rs
```diff
@@ -0,0 +1,81 @@
...
+        .resolve_modifiers(true)
+        .simplify_cfgs(false)
+        .remove_tuple_untuple(false)
+        .constant_folding(false)
+        .remove_dead_funcs(false)
+        .inline_dfgs(false)
+        .squash_borrows(false)
+        .remove_redundant_order_edges(false);
+    normalize.run(hugr).unwrap();
+}
```
aborgna-q: I would replace this function with a direct call to `ModifierResolverPass::default()`.

We're not gaining anything from calling it like this, and we're bound to forget updating this call when adding new passes to NormalizeGuppy.

tket/benches/benchmarks/modifier_resolver.rs
```diff
@@ -0,0 +1,81 @@
+//! Benchmark the modifier resolver pass.
```
aborgna-q, line 1: Pls add a comment about what each named benchmark is doing.

I would rename them to be explicit about running modifier pass, e.g. `modifier_pass_double_modifier`, so it's easier to know what they are about from the codspeed overview: 
https://app.codspeed.io/Quantinuum/tket2/branches/na%2F1642-add-modifier-to-benchmarks?utm_source=github&utm_medium=check&utm_content=button

aborgna-q (approved):
Thanks!

PR #1711, opened 2026-06-17
chore: Update guppylang dependencies in example uv scripts 

### LINKED ISSUE #1571: chore: Update examples in `test_files` folder after guppy release with measurement changes
All python examples with 

```
# [tool.uv.sources]
# guppylang = {git = "https://github.com/quantinuum/guppylang", subdirectory = "guppylang", branch = "ts/future-measure"}
```

Should be updated with the appropriate released guppy version once that is possible

tatiana-s:
Closes https://github.com/Quantinuum/tket2/issues/1571


aborgna-q (dismissed):

test_files/guppy_optimization/justfile
```diff
@@ -11,7 +11,7 @@ recompile-hugrs:
     # Find all examples in subdirectories and re-run them.
     for example in $(find . -type f -name "*.py"); do
         echo "Re-generating $example"
-        uv run --no-project "$example"
+        uv run --prerelease=allow --no-project "$example"
```
nicolaassolini-qntm, line 14: `--prerelease=allow` is dangerous: guppylang4a requires guppyinternals>=4a, thus, with this flag, you're executing guppylang with a different version of internals, and this can cause some hidden breaking (at least this is the case with modifier). I suggest fixing the version of both internals and guppy, at least while we are referring to alpha releases. 

While you're updating, you can also update the dependencies to guppy 1.a5 😄

nicolaassolini-qntm (changes requested):

test_files/guppy_optimization/justfile
nicolaassolini-qntm: Update, to avoid:
```
× No solution found when resolving script dependencies:
  ╰─▶ Because only selene-sim<0.3.0a1 is available and guppylang==1.0.0a5 depends on selene-sim>=0.3.0a1, we can conclude that guppylang==1.0.0a5 cannot be used.
      And because you require guppylang==1.0.0a5, we can conclude that your requirements are unsatisfiable.
```
we still need `--prerelease=allow`, but after fixing the specif internals version.

nicolaassolini-qntm (approved):
:shipit:

PR #1718, opened 2026-06-18
chore: update hugr to 0.28.2

hsemenenko:


aborgna-q:
qis-compiler check fails due to outdated LLVM snapshots.
Can you run `just update-snapshots` ?

aborgna-q (approved):

PR #1719, opened 2026-06-18
test: add cross compilation test cases

ferbetanzo:
Adding a few cross compilation test cases

- Lowering generic and platform specific barriers.
- Re-lowering hugr that contains platform runtime barriers is not allowed.
- Cross platform compilation roundtrip.
- Lowering a hugr that contains all supported TketOp, SharedOP and platform-specific ops.


tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -1154,27 +1211,229 @@ mod test {
...
+
+        // Lowering succeeds
+        lower_tk2_ops(&mut h, Preserve::Public, platform).unwrap();
+        h.validate().unwrap();
+
+        // The <platform>::RuntimeBarrier op should still be present (not remapped).
+        let barrier_still_present = h.nodes().any(|n| {
+            h.get_optype(n)
+                .as_extension_op()
+                .is_some_and(|op| op.def().extension_id() == ext_id)
```
aborgna-q: Could we check that there is an actual barrier op?

aborgna-q (approved):

PR #1720, opened 2026-06-18
perf(qsystem): pre-load lowerer with all replacements (#1717)

### LINKED ISSUE #1717: [Performance]: Qsystem lowering adds same op to ReplaceTypes many times
### Performance Issue

tket-qsystem/src/extension/qsystem/lower.rs lazily constructs the lowerer by first collecting all the ops that need to be replaced then for each one adding the implementation function to the lowerer. 

Given the ops that needs replacing are known and far fewer in number than the typical number of times those ops will be present in an input HUGR, we should instead simply pre-load the lowerer with _all_ lowerings and let it handle it. This saves at least one traversal to find the ops, and should simplify the code significantly. There should be no change in functionality or interfaces.

### Steps to Reproduce

Create a HUGR with many repeated tket.quantum ops and construct a benchmark with it

### Performance Metrics

```markdown

```

### Expected Performance

_No response_

### Component

Rust library

### Environment

```markdown

```

### Optimization Suggestions

_No response_

### Additional Context

_No response_

ss2165:
Closes #1717.

## What

`lower_tk2_ops` previously built the `ReplaceTypes` lowerer *lazily*: it
traversed the HUGR to collect every node needing replacement, then looped over
**every occurrence** and re-called `lowerer.set_replace_op(...)` (re-inserting a
template into a `BTreeMap` per occurrence). The set of distinct ops to lower is
small and known ahead of time, so this was redundant work.

Now all replacements (every `TketOp`, the cross-platform remaps, the legacy
`tket.qsystem` ops and the measurement replacements) are registered **once** up
front via two new helpers, `register_tket_op_replacements` and
`register_cross_platform_replacements`. A single traversal remains, but only to
collect the returned replaced-node list and to perform the per-node barrier
insertion (which mutates the HUGR via `InsertCut` and cannot be a pure
`ReplaceTypes` registration). The per-op caches and the `apply_cross_platform_*`
helpers are gone.

No functional or interface change: `lower_tk2_ops` still returns `Vec<Node>`,
and the full existing test suite (148 tests) passes unchanged.

## Benchmark

Added a criterion benchmark to `tket-qsystem` (mirroring `tket/benches`) that
lowers HUGRs with many repeated quantum ops (layers of `H`/`CX`/`Rx` +
`Reset`), 8 qubits, for both platforms.

| benchmark | baseline | this PR | change |
|---|---|---|---|
| helios / 10 layers | 2.78 ms | 3.17 ms | **+14%** |
| helios / 100 layers | 25.2 ms | 21.2 ms | **-15.7%** |
| helios / 1000 layers | 665 ms | 605 ms | **-9.1%** |
| sol / 10 layers | 2.28 ms | 3.31 ms | **+45%** |
| sol / 100 layers | 24.5 ms | 21.2 ms | **-13.5%** |
| sol / 1000 layers | 654 ms | 603 ms | **-7.9%** |

## Assessment / trade-off

- Pre-loading helps realistically-sized HUGRs (the issue's target: "many
  repeated ops"), saving ~8-16%.
- The small-input "regression" is a **fixed ~1 ms cost**: we now build all ~20
  multi-op decomposition templates up front even when only a few op kinds are
  present. It's negligible in absolute terms and irrelevant for whole-program
  lowering, but it does dominate the tiny 10-layer cases.
- The super-linear scaling (100->1000 ~= 28x) **persists**, which shows the
  dominant cost is `ReplaceTypes::run` itself, not the per-occurrence
  `set_replace_op` churn this PR removes. The classification traversal is
  <1% of runtime.

I deliberately did **not** pursue an interface-breaking follow-up (dropping the
`Vec<Node>` return to remove the traversal): barrier lowering needs a per-node
pass regardless of the return type, and the traversal isn't the bottleneck, so
the break would buy nothing. If we want a larger win, the lever is
`ReplaceTypes::run` / barrier insertion, not this registration path.

## Commits

1. `bench(qsystem)` - the benchmark + baseline numbers.
2. `refactor(qsystem)` - the pre-load refactor.

### DIFF
--- a/tket-qsystem/benches/benchmarks/generators.rs
+++ b/tket-qsystem/benches/benchmarks/generators.rs
@@ -1,0 +1,55 @@
+//! Helpers for building HUGRs that exercise the QSystem lowering pass.
+
+use hugr::CircuitUnit;
+use hugr::Hugr;
+use hugr::Wire;
+use hugr::builder::{Dataflow, DataflowHugr, FunctionBuilder};
+use hugr::extension::prelude::qb_t;
+use hugr::types::Signature;
+use tket::TketOp;
+use tket::extension::rotation::rotation_type;
+
+/// Build a module-rooted HUGR containing many repeated `tket.quantum` ops.
+///
+/// Each layer applies, across all qubits:
+/// - `H` on every qubit (multi-op / function-template lowering),
+/// - `CX` between neighbouring qubit pairs (multi-op),
+/// - `Rx` on every qubit, reusing a single copyable rotation input (multi-op),
+/// - `Reset` on every qubit (direct-map lowering).
+///
+/// This mixes the function-template path (the one that previously re-registered
+/// a replacement for every occurrence) with the direct-map path, so the total
+/// op count grows linearly with `layers` while the set of distinct ops stays
+/// tiny.
+pub fn make_quantum_layers(num_qubits: usize, layers: usize) -> Hugr {
+    let qb_row = vec![qb_t(); num_qubits];
+    let mut inputs = qb_row.clone();
+    inputs.push(rotation_type());
+
+    let mut h = FunctionBuilder::new("main", Signature::new(inputs, qb_row.clone())).unwrap();
+
+    let mut wires: Vec<Wire> = h.input_wires().collect();
+    let angle = wires.pop().expect("rotation input wire");
+
+    let mut circ = h.as_circuit(wires);
+    for _ in 0..layers {
+        for q in 0..num_qubits {
+            circ.append(TketOp::H, [q]).unwrap();
+        }
+        for i in 0..num_qubits / 2 {
+            circ.append(TketOp::CX, [2 * i, 2 * i + 1]).unwrap();
+        }
+        for q in 0..num_qubits {
+            circ.append_and_consume(
+                TketOp::Rx,
+                [CircuitUnit::Linear(q), CircuitUnit::Wire(angle)],
+            )
+            .unwrap();
+        }
+        for q in 0..num_qubits {
+            circ.append(TketOp::Reset, [q]).unwrap();
+        }
+    }
+    let qbs = circ.finish();
+    h.finish_hugr_with_outputs(qbs).unwrap()
+}
--- a/tket-qsystem/benches/benchmarks/generators.rs
+++ b/tket-qsystem/benches/benchmarks/generators.rs
@@ -21,7 +21,7 @@ use tket::extension::rotation::rotation_type;
 /// a replacement for every occurrence) with the direct-map path, so the total
 /// op count grows linearly with `layers` while the set of distinct ops stays
 /// tiny.
-pub fn make_quantum_layers(num_qubits: usize, layers: usize) -> Hugr {
+pub fn make_h_cx_rx_reset_layers(num_qubits: usize, layers: usize) -> Hugr {
     let qb_row = vec![qb_t(); num_qubits];
     let mut inputs = qb_row.clone();
     inputs.push(rotation_type());
--- a/tket-qsystem/src/extension/qsystem/lower.rs
+++ b/tket-qsystem/src/extension/qsystem/lower.rs
@@ -12,8 +12,6 @@ use hugr::{
     types::Signature,
 };
 use lazy_static::lazy_static;
-use std::collections::BTreeMap;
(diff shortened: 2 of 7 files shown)


tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -277,56 +276,34 @@ pub fn lower_tk2_ops(
     let mut replaced_nodes = Vec::with_capacity(replacements.len());
```
acl-cqc: The only thing that slightly bothers me here is that these `replaced_nodes` are the nodes that one part of code (building the list) *thinks* another bit of code is going to replace.... but yeah, ReplaceTypes is too much of a closed box to do that any other way.

So the only thing we could do here would be to rename `replaced_nodes` to something else, say `nodes_about_to_replace`, as a slight clarity improvement. I'll leave that up to you to decide...

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -338,6 +315,98 @@ pub fn lower_tk2_ops(
...
+            for s_op in SolOp::iter() {
+                match SharedOp::try_from(s_op) {
+                    Ok(shared) => lowerer.set_replace_op(
+                        &s_op.to_extension_op().expect("valid registered SolOp"),
+                        NodeTemplate::SingleOp(HeliosOp::from(shared).into()),
+                    ),
+                    Err(_) if global => {
+                        debug_assert_eq!(s_op, SolOp::PhasedXX, "unexpected Sol-specific op");
+                        let template =
+                            NodeTemplate::call_to_function(build_sol_op_for_helios()?, &[])
```
acl-cqc: This `build_sol_op_for_helios` seems misnamed - it doesn't take any parameters saying what sol-op, can we rename to `build_sol_phasedxx_for_helios`?

tket-qsystem/benches/benchmarks/generators.rs
```diff
@@ -0,0 +1,55 @@
...
+/// - `H` on every qubit (multi-op / function-template lowering),
+/// - `CX` between neighbouring qubit pairs (multi-op),
+/// - `Rx` on every qubit, reusing a single copyable rotation input (multi-op),
+/// - `Reset` on every qubit (direct-map lowering).
+///
+/// This mixes the function-template path (the one that previously re-registered
+/// a replacement for every occurrence) with the direct-map path, so the total
+/// op count grows linearly with `layers` while the set of distinct ops stays
+/// tiny.
+pub fn make_quantum_layers(num_qubits: usize, layers: usize) -> Hugr {
```
acl-cqc: I might call this `make_h_cx_rx_reset_layers` as I think there are a few other possibilities

tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -277,56 +276,34 @@ pub fn lower_tk2_ops(
...
+            }
             ReplaceOps::Barrier(barrier) => {
-                // Handle barrier replacements
-                //
-                // Only perform the replacement for global passes, as we
-                // cannot define the barrier function for local entrypoint scopes.
-                if let PassScope::Global(_) = &scope {
+                // Only perform the replacement for global passes, as we cannot
+                // define the barrier function for local entrypoint scopes.
+                if global {
```
acl-cqc, line 290: it does seem like one could actually include this barrier *in the replacement registered with the lowerer* with a bit more fun constructing it.

Then maybe we could make ReplaceTypes itself return a list of replaced nodes, which could be quite a nice feature, and we'd not need all of this. (However I guess there is an issue if you replace an op with some Hugr that includes an op that itself gets replaced - the returned Node would not really be meaningful).

acl-cqc (approved):
Yeah. So I think the decision made by this PR is whether the one-time setup cost, of constructing replacements for all ops even those not present in the underlying Hugr, is worth it; and your results do show that cost, but yeah, I agree it's acceptable, the reduction for big hugrs is more significant, and also the code cleanup :), so yes, looks good to me! Thanks :)

tket-qsystem/src/extension/qsystem/lower.rs
ss2165: agree but given the linear traversal is negligible performance wise, I don't think it is worth the churn on barrier code (which would be siginificant)

PR #1721, opened 2026-06-18
fix: Higher-order modifier input tracing through nested containers

### LINKED ISSUE #1710: [Bug]: ModifierResolverPasses are unstable with higher order functions
### Description

There are still some errors

### Steps to Reproduce

running passes on:

```python
# /// script
# requires-python = ">=3.13"
# dependencies = [
#    "guppylang==1.0.0a8",
# ]
# ///
"""Test the use of a higher-order function with complex control flow inside modifiers"""

from pathlib import Path
from sys import argv
from collections.abc import Callable

from guppylang import enable_experimental_features, guppy
from guppylang.std.builtins import Controllable, Unitary, array, control, dagger
from guppylang.std.debug import state_result
from guppylang.std.lang import Function
from guppylang.std.quantum import angle, discard_array, h, qubit, rx, rz

enable_experimental_features()


@guppy
def get_angle(f: float) -> angle:
    return angle(f)


@guppy
def get_get_angle() -> Function[[float], angle]:
    return get_angle


@guppy(unitary=True)
def apply_r(
    f: Unitary[[qubit, angle], None],
    q: array[qubit, 2],
    fun_angle: Callable[[float], angle],
    radiant: float,
) -> None:
    f(q[1], fun_angle(radiant))


@guppy(controllable=True)
def apply_c(
    f: Controllable[[qubit], None],
    g: Unitary[[qubit, angle], None],
    classic_fun: Function[[], Function[[float], angle]],
    q: qubit,
    b: bool,
) -> None:
    n = 3
    if b:
        while n > 0:
            f(q)
            n -= 1
    else:
        get_a = classic_fun()
        angle = get_a(0.25)
        for _ in range(2):
            g(q, get_a(0.25))
            g(q, angle)


@guppy
def main() -> None:
    qs: array[qubit, 2] = array(qubit(), qubit())
    h(qs[0])
    flag = 2 > 10
    with control(qs[0]):
        apply_c(h, rx, get_get_angle, qs[1], True)
        apply_c(h, rx, get_get_angle, qs[1], flag)

    with control(qs[0]), dagger:
        apply_r(rz, qs, get_angle, 0.25)
        apply_r(rz, qs, get_angle, 0.5)

    state_result("r", qs[0], qs[1])
    discard_array(qs)


program = main.compile()
Path(argv[0]).with_suffix(".hugr").write_bytes(program.to_bytes())

```

nicolaassolini-qntm:
closes #1710

Previously, `higher_order_input_modifiers` treated the port index of any reached `Input` node as if it were an input index of the outer `FuncDefn`. That is only valid when the `Input` belongs directly to the function. In cases involving `CFG`, `DataflowBlock`, loops, or conditionals, the reached `Input` can be local to a nested container, so the resolver could rewrite the wrong higher-order function input type and build an invalid HUGR.

This PR changes the resolver to trace the actual value provenance outward through nested containers before recording a higher-order function input requirement. It handles ordinary container inputs, conditional sum fields, and CFG branch payloads, so the recorded requirement refers to the real top-level function input.

Moreover, the resolver now also handles consecutive indirect calls more explicitly in modifier contexts. When the callee of an indirect call is itself produced by another `CallIndirect`, the resolver checks the producer call's signature. If that producer is purely classical, the outer indirect call is left unchanged. If the producer has a quantum signature, resolution now fails since function calls of the form `f(g())` where 'f' and 'g' have quantum signatures, must allocate or consume qubits.

Also includes small improvements in `modifier_resolver.rs`:
- Changed `controls_ref` to return `&[Wire]` instead of `&Vec<Wire>`.
- Improved `port_vector_rev` lookup by using a `HashSet`.
- Simplified `map_insert`.


nicolaassolini-qntm:
the issue has been solved after the refactor in #1827 ([a74c5ec](https://github.com/Quantinuum/tket2/commit/a74c5ecbe8e28a087e3d0d78393b6fedfc9d66ab))

PR #1722, opened 2026-06-18
chore: update snaps to changes from #1706

ss2165:
Just re-indexing/re-labelling noise (note line count change is 0)
Required workflows did not force re-compute


maximilianruesch (commented):
Could you investigate why the workflows did not force these to be reissued? Sounds like something we should fix...

ss2165:
I'm happy to open an issue and investigate separately but this would be a separate change anyway and this PR should merge to make main green, preventing other PRs from failing

maximilianruesch (approved):
> I'm happy to open an issue and investigate separately

Please do

ss2165:
https://github.com/Quantinuum/tket2/issues/1727

PR #1728, opened 2026-06-19
chore: Updated dependecies in modifier test files

nicolaassolini-qntm:
"guppylang==1.0.0a4" requires "guppylang-internals>=1.0.0a4". Since "guppylang-internals" is marked as pre-reselad, if guppy-internals is not cached by uv we may ave an error.


aborgna-q:
I saw tatiana add `--prerelease=allow` to the `uv run` command in the justfile in https://github.com/Quantinuum/tket2/pull/1711.
https://github.com/Quantinuum/tket2/blob/e8b235a88e536f8032215ce310becce5858c3bc8/test_files/guppy_optimization/justfile#L14

Would that help here?

nicolaassolini-qntm:
> I saw tatiana add `--prerelease=allow` to the `uv run` command in the justfile in #1711.
> 
> https://github.com/Quantinuum/tket2/blob/e8b235a88e536f8032215ce310becce5858c3bc8/test_files/guppy_optimization/justfile#L14
> 
> Would that help here?

This breaks stuff in some examples. I guess because it's getting pre-released versions of other packages also. I'm testing now with guppy 1.0.05a

nicolaassolini-qntm:
with 1.0.05a is fine, the error was probably because guppy1.a4 cannot work with internals1.a5 (metadata key changes)

nicolaassolini-qntm:
But I would like to fix both versions rather than having a flag silently update something that may break things

devenv.lock
```diff
@@ -63,4 +63,4 @@
   },
   "root": "root",
   "version": 7
-}
```
aborgna-q, line 66: If this was autogenerated, we may want to keep the missing newline.

(There's an option to ignore the file in the `.pre-commit-config.yml`)

aborgna-q (approved):

devenv.lock
nicolaassolini-qntm: It was auto-generated by `just check`

PR #1682, opened 2026-06-11
feat: User program entrypoint arguments (MVP)

jake-arkinstall:
Adds operations, lowering, and surgery for introducing runtime argument handling, based on the selene argreader utility.

This PR requires a lot of love and attention - it is presently just an MVP, knocking out the core parts to get functional snapshots building. I will happily offload cleanup, refactoring to canonical usage, and testing to others with more experience in the tket2 repository.

To try out the entrypoint snapshot with selene, I will provide the code in a comment to this PR.

PR #1731, opened 2026-06-19
feat!: runtime entrypoint arguments via generic tket.argreader op

ss2165:
## Runtime entrypoint arguments (`tket.argreader`)

Based on #1682 (originally by @jake-arkinstall). 

### What this does
Lets a compiled user program read runtime arguments at the entrypoint, so a program can be **compiled once and run many times with different parameters** (the motivating use case is variational workloads like QAOA, where today every parameter update forces a recompile).

Dependant guppy PR: https://github.com/Quantinuum/guppylang/pull/1890

- The argreader extensionexposes one generic `read_arg<Tag, T>` op that is parametrised by string tag (like `result`) and the type it outputs. 
- `tket-qsystem`: argreader extension definition (`extension/argreader.rs`) and LLVM lowering (`llvm/argreader.rs`, `llvm/prelude.rs`). Roughly mirrors `result` codegen. However, this new op can create new arrays so needs allocation in ArrayLowering.
- `tket-exts`: ships the `tket.argreader` extension JSON + Python bindings.
- `qis-compiler`: wires the argreader plugin into codegen; regenerated test HUGRs and LLVM/selene snapshots (the large diff is almost entirely regenerated snapshot fixtures).

BREAKING CHANGE: ArrayLowering trait has new required method alloc_array for allocating a fresh array.


qis-compiler/rust/lib.rs
```diff
@@ -217,10 +222,10 @@ fn get_entry_point_name(namer: &Namer, hugr: &impl HugrView<Node = Node>) -> Res
             .as_func_defn()
             .ok_or_else(|| anyhow!("Entry point node is not a function definition"))?;
         if func_defn.inner_signature().input_count() != 0 {
-            return Err(anyhow!(
+            bail!(
```
maximilianruesch: Just a random driveby? This does not seem related to the PR at all (it is even almost 100 lines apart from other changes in this file...

qis-compiler/rust/lib.rs
```diff
@@ -135,11 +136,15 @@ fn codegen_extensions(platform: qsystem::QSystemPlatform) -> CodegenExtsMap<'sta
         .add_extension(ResultsCodegenExtension::new(
             SeleneHeapArrayCodegen::LOWERING,
         ))
-        .add_extension(RotationCodegenExtension::new(pcg))
+        .add_extension(RotationCodegenExtension::new(pcg.clone()))
```
maximilianruesch:
```suggestion
        .add_extension(RotationCodegenExtension::new(pcg))
```
I do not see you using `pcg` after this point, so I do not think this clone is needed here.

tket-qsystem/src/llvm/prelude.rs
```diff
@@ -14,7 +14,7 @@ use inkwell::types::{BasicType, IntType};
 use inkwell::values::{BasicValue, BasicValueEnum, FunctionValue, StructValue};
 use tket::hugr::{HugrView, Node};
 
-#[derive(Clone)]
+#[derive(Clone, Default)]
```
maximilianruesch: Unrelated change much?

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// A signed 64-bit integer argument (i64 / log-width 6 only).
+    Int,
+    /// A 64-bit floating-point argument.
+    F64,
+    /// An array of booleans with a fixed length.
+    ArrBool(u64),
+    /// An array of signed 64-bit integers with a fixed length.
+    ArrInt(u64),
+    /// An array of 64-bit floats with a fixed length.
+    ArrF64(u64),
```
maximilianruesch, line 43: Any reason these are `u64` instead of `usize`?

tket-exts/src/tket_exts/tket/argreader.py
```diff
@@ -0,0 +1,50 @@
...
+from ._util import TketExtension, load_extension
+
+
+class ArgReaderExtension(TketExtension):
+    """Operations for reading runtime entrypoint arguments."""
+
+    @functools.cache
+    def __call__(self) -> Extension:
+        """Returns the argreader extension"""
+        return load_extension("tket.argreader")
```
maximilianruesch: How about `tket.arguments` instead? Fits more into the existing names, and feels less like a "I am an object (or a factory for objects) that can read arguments" OOP vibe. Unless that is what you are going for.

tket-qsystem/src/llvm/argreader.rs
```diff
@@ -0,0 +1,423 @@
...
+
+/// The concrete variant of an argument type, used to select the extern function.
+///
+/// Produced by [`classify_arg_type`]; lives here because it is purely a codegen concept.
+#[derive(Debug, Clone, PartialEq)]
+enum ArgKind {
+    /// A boolean argument.
+    Bool,
+    /// A signed 64-bit integer argument (i64 / log-width 6 only).
+    Int,
```
maximilianruesch: Should perhaps be `I64` to communicate both signed and width?

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,423 @@
...
+    }
+}
+
+impl<AL: ArrayLowering + Clone> CodegenExtension for ArgReaderCodegenExtension<AL> {
+    fn add_extension<'a, H: HugrView<Node = Node> + 'a>(
+        self,
+        builder: CodegenExtsBuilder<'a, H>,
+    ) -> CodegenExtsBuilder<'a, H>
+    where
+        Self: 'a,
```
maximilianruesch, line 194:
```suggestion
    fn add_extension<'a, H: HugrView<Node = Node> + 'a>(
        &'a self,
        builder: CodegenExtsBuilder<'a, H>,
    ) -> CodegenExtsBuilder<'a, H>
```
Would this also work? I find the lifetime requirement on a non-borrowed `self` confusing...

tket-qsystem/src/llvm/argreader.rs
```diff
@@ -0,0 +1,423 @@
...
+            let op = ReadArgOp::from_extension_op(args.node().as_ref())?;
+            ArgReaderEmitter(context, self.array_lowering.clone()).emit(args, &op)
+        })
+    }
+}
+
+struct ArgReaderEmitter<'c, 'd, 'e, H: HugrView<Node = Node>, AL: ArrayLowering>(
+    &'d mut EmitFuncContext<'c, 'e, H>,
+    AL,
+);
```
maximilianruesch: Can the lifetimes have more descriptive names?

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    fn void_t(&self) -> VoidType<'c> {
+        self.iw_context().void_type()
+    }
+
+    fn builder(&self) -> &Builder<'c> {
+        self.0.builder()
+    }
+
+    fn get_argreader_func(&self, kind: &ArgKind) -> Result<FunctionValue<'c>> {
```
maximilianruesch, line 247: Where are these external functions implemented? Selene? I gather from code below that the first pointer is always the tag (this could be on a docstring here), but why do we need the second pointer for the arrays? Is it because we cannot return arrays directly, and as such need to pass a pointer to store the result in?

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,423 @@
...
+            self.builder()
+                .build_call(argread_fn, &[tag_ptr.into()], call_name)?
+                .try_as_basic_value()
+                .unwrap_basic()
+        };
+
+        args.outputs.finish(self.builder(), [result])
+    }
+
+    fn emit_array_read(
```
maximilianruesch, line 325: I feel like there is no real reason this function is not inlined. It is not used anywhere else, and short enough to almost be the same length as the call itself.

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,423 @@
...
+    #[case::narrow_int(8, ReadArgOp::new("test_narrow", int_type(TypeArg::BoundedNat(3))), &DEFAULT_HEAP_ARRAY_LOWERING)]
+    #[should_panic(expected = "log-width 6")]
+    #[case::narrow_int_arr(
+        9,
+        ReadArgOp::new("test_narrow_arr", array_type(4, int_type(TypeArg::BoundedNat(3)))),
+        &DEFAULT_HEAP_ARRAY_LOWERING
+    )]
+    fn emit_argreader_codegen(
+        #[case] _i: i32,
+        #[with(_i)] mut llvm_ctx: TestContext,
```
maximilianruesch, line 385: Why do we need different ids for the test context?

tket-qsystem/src/llvm/argreader.rs
```diff
@@ -0,0 +1,423 @@
...
+    #[case::narrow_int_arr(
+        9,
+        ReadArgOp::new("test_narrow_arr", array_type(4, int_type(TypeArg::BoundedNat(3)))),
+        &DEFAULT_HEAP_ARRAY_LOWERING
+    )]
+    fn emit_argreader_codegen(
+        #[case] _i: i32,
+        #[with(_i)] mut llvm_ctx: TestContext,
+        #[case] op: ReadArgOp,
+        #[case] array_lowering: &'static (impl ArrayLowering + Clone),
```
maximilianruesch: This argument seems to always be `&DEFAULT_HEAP_ARRAY_LOWERING`, and I think it can be inlined.

tket-qsystem/src/llvm/argreader.rs
```diff
@@ -0,0 +1,423 @@
...
+    #[test]
+    fn test_classify_int_rejects_narrow() {
+        for log_width in 0u64..=5 {
+            let err = classify_arg_type(&int_type(TypeArg::BoundedNat(log_width))).unwrap_err();
+            assert!(
+                err.to_string().contains("log-width 6"),
+                "log_width={log_width}: {err}"
+            );
+        }
+    }
```
maximilianruesch: This for loop can (and imo should) be an rstest.

tket-qsystem/src/llvm/argreader.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    #[test]
+    fn test_classify_arr_int_rejects_narrow() {
```
maximilianruesch: Why test all above, and not all of the log widths here?

maximilianruesch (changes requested):
A bunch of questions, but they should be quick. Nothing major to comment on imo.

tket-qsystem/src/llvm/prelude.rs
ss2165: fixed and forgot to push oops

tket-qsystem/src/llvm/argument.rs
ss2165: The lengths come in as `TypeArg::BoundedNat(u64)` from the HUGR type args and go straight into inkwell's `IntType::const_int(u64, ..)` / `u32::try_from`. `u64` is the type at both ends, so `usize` would only add platform-dependent casts for no gain. Keeping `u64`.

tket-qsystem/src/llvm/argument.rs
ss2165: This is dictated by the `CodegenExtension` trait, which declares `fn add_extension<'a, H>(self, ..) -> .. where Self: 'a` — `self` by value, with the `where Self: 'a` bound. `&'a self` wouldn't satisfy the trait. Every sibling extension (`result.rs`, `debug.rs`, `qsystem.rs`, …) has the identical signature, so this stays as-is.

tket-qsystem/src/llvm/argreader.rs
ss2165: Keeping these — they're the exact lifetimes used by every sibling emitter (`ResultEmitter<'c, 'd, 'e, ..>`, `DebugEmitter`, …): `'c` the inkwell context, `'d` the borrow of the `EmitFuncContext`, `'e` its internal lifetime. Renaming one emitter to break from that convention would do more harm than good.

tket-qsystem/src/llvm/argument.rs
ss2165: Both right. They're implemented in Selene (the `argreader_get_*` externs). The first pointer is always the NUL-terminated tag (argument name); for the array variants the second pointer is the caller-allocated output buffer Selene writes into — we can't return an array by value across the C ABI — and the third arg is the element count. I've documented this on `get_argreader_func`.

tket-qsystem/src/llvm/argument.rs
ss2165: Keeping it split out: `emit_array_read` isolates the array path (length-bounds check + alloc + extern call) from the scalar path in `emit`, which reads better than folding it into the branch. The single call site is intentional for readability.

tket-qsystem/src/llvm/argument.rs
ss2165: The id seeds the `TestContext` so each `rstest` case writes to its own snapshot file (`…_1.snap` … `…_6.snap`) — without distinct ids the cases collide on one snapshot. It's the same `#[case] _i` + `#[with(_i)]` pattern as the other LLVM tests (`debug.rs`, `futures.rs`, `globals.rs`, `qsystem.rs`). I've added a comment spelling that out.

tket-qsystem/src/llvm/prelude.rs
ss2165: Agreed, unrelated — this was already dropped when I rebased on `main`; `prelude.rs` is back to its original state. Resolving.

tket-exts/src/tket_exts/tket/argreader.py
ss2165: went with "tket.argument" to mirror singular of "tket.result"

tket-exts/src/tket_exts/tket/argument.py
```diff
@@ -0,0 +1,50 @@
+"""Argument reader extension operations."""
```
maximilianruesch, line 1:
```suggestion
"""Argument extension operations."""
```
(although technically also true as of now)

tket-qsystem/src/llvm/argreader.rs
maximilianruesch: > Renaming one emitter to break from that convention would do more harm than good.

If the convention does not offer any benefit, sticking to the convention does not offer any benefit either, so breaking from it does no harm. Disagreed.

tket-qsystem/src/llvm/argument.rs
```diff
@@ -0,0 +1,432 @@
...
+    #[should_panic(expected = "log-width 6")]
+    #[case::narrow_int(8, ReadArgOp::new("test_narrow", int_type(TypeArg::BoundedNat(3))))]
+    #[should_panic(expected = "log-width 6")]
+    #[case::narrow_int_arr(
+        9,
+        ReadArgOp::new("test_narrow_arr", array_type(4, int_type(TypeArg::BoundedNat(3))))
+    )]
+    fn emit_argument_codegen(
+        // `_i` seeds the `TestContext` so each case emits to its own snapshot file
+        // (`…_1.snap` … `…_9.snap`); without distinct ids the cases would collide.
```
maximilianruesch:
```suggestion
        // (`..._1.snap` ... `..._9.snap`); without distinct ids the cases would collide.
```
these are currently non-ascii, which may not be best for some IDEs

tket-qsystem/src/llvm/argument.rs
maximilianruesch: This only partially answered the question (since an implicit question was how `Self: 'a` adds constraints to a non-borrowed argument). But I researched the answer myself now.

tket-qsystem/src/llvm/argument.rs
maximilianruesch: There may be other places where we will need the external functions to be defined, outside of Selene (or rather, associated plugins). But fine for now.

maximilianruesch (approved):
Even though we disagree on one comment, it is not critical to the code path / functionality, so this can theoretically be merged if needed.

PR #1734, opened 2026-06-22
chore: Regenerate qis-compiler tests on guppy v1 alpha

ss2165:


maximilianruesch (approved):
I loathe the circular reference

PR #1739, opened 2026-06-23
feat: Added an error if non quantum tket-op are modified

### LINKED ISSUE #1735: Raise an error if non-unitary tketop are found
Guppy forbid to allocate or measure/consume qubits inside modifier blocks. 
However, when tket2 pass find a `Measure | MeasureFree | QAlloc | TryQAlloc | QFree | Reset` operator, it simply keeps the same operator without doing anything ([tket2/tket/src/modifier/modifier_resolver/tket_op_modify.rs#L476](https://github.com/Quantinuum/tket2/blob/1281928611af7593ef416b66e86156470da0e737/tket/src/modifier/modifier_resolver/tket_op_modify.rs#L476)). 
It is impossible to have a guppy with a modifier applied to a measurement, but it is possible to build an hugr that does that. As far as modifying measurement & co operations in guppy are illegal, I would like to have the same behaviour in tket2, thus I would raise an error during passing instead of doing nothing.

nicolaassolini-qntm:
closes #1735


tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -627,6 +628,36 @@ mod test {
...
+                .define_function("foo", Signature::new_endo([]))
+                .unwrap();
+            let op_node = func.add_child_node(op);
+            let mut resolver = ModifierResolver::new();
+
+            let result = resolver.modify_tket_op(op_node, op, &mut func, &mut vec![]);
+            match result {
+                Err(ModifierResolverErrors::UnResolvable { node, msg, optype }) => {
+                    assert_eq!(node, op_node);
+                    assert_eq!(msg, format!("{op:?} operation cannot be modified"));
```
acl-cqc: Generally I'd try to avoid repeating strings in prod and test.

Also I'm wary of that `ModifierResolverErrors::Unresolvable` both includes a custom string, and formats it into a `#[display(...)]` template - so what the user will actually see here is, IIUC:
`
Modifier {node} applied to the node TketOp::Measure operation cannot be modified cannot be modified.
`

So for starters, remove the "cannot be modified" that you put into `msg: String` when you construct this error.

You might also want to include `optype` in the `#[display(...)]` template, in which case it's not clear you want a `msg` here at all. (Maybe UnResolvable should have `msg: Option<String>` and the `#[display(...)]` append msg if present, or special-case zero-length `msg` vs not.)

And/or rather than string-matching in the test you might want to consider adding a new error variant(s)

acl-cqc (commented):
So I think there's an issue here with the error message, and I'm gonna propose a minimal cleanup of removing the duplicate "cannot be modified" and preferably refactoring the test, as part of this PR.

However, it also looks as if the general format/`#[display(...)]` of Unresolvable is broken and I suggest a followup PR to clean up the error definition.

*Secondly*, I have to say - it'd be good to see the 440+-line `match tket_op {....}` in `fn modify_tket_op` broken up into, say, a separate function for each case, but again, that can be a followup PR

acl-cqc (approved):
`Modifier {node)  applied to the node of type {optype} cannot be modified` isn't great - but probably as good as you can get with that format string, so ok, roll on a followup ;)

PR #1742, opened 2026-06-23
feat!: update to hugr 0.29.0/hugr-py 0.18.0, tone down RedundantOrderEdgesPass

acl-cqc:
RedundantOrderEdges pass was overdue an update because it was still skipping order edges to containers (fearing these might be the order edges which were once specifically required by nonlocal `Ext` edges, but that requirement was removed in hugr 0.28.0).

But, following https://github.com/Quantinuum/hugr/pull/3058 (in 0.29.0 release), we can only safely remove Order edges that are redundant due to other *Order* edges; it becomes incorrect to remove order edges that are redundant due to the presence of *value* edges.

I ended up redoing the pass significantly, the "new" version is about half the length...

BREAKING CHANGE: updates to public dependencies, specifically hugr 0.29.0 and hugr-py 0.18.0


Cargo.toml
```diff
@@ -41,10 +41,10 @@ large_enum_variant = "allow"
...
 
 # Uncomment to use unreleased versions of hugr
-#hugr = { git = "https://github.com/quantinuum/hugr", "rev" = "1787e12" }
-#hugr-core = { git = "https://github.com/quantinuum/hugr", "rev" = "1787e12" }
-#hugr-cli = { git = "https://github.com/quantinuum/hugr", "rev" = "1787e12" }
-#hugr-llvm = { git = "https://github.com/quantinuum/hugr", "rev" = "1787e12" }
+hugr = { git = "https://github.com/quantinuum/hugr", "rev" = "b8ba9ed53b1411ce5f6e9925d1a70772e6c44e65" }
+hugr-core = { git = "https://github.com/quantinuum/hugr", "rev" = "b8ba9ed53b1411ce5f6e9925d1a70772e6c44e65" }
+hugr-cli = { git = "https://github.com/quantinuum/hugr", "rev" = "b8ba9ed53b1411ce5f6e9925d1a70772e6c44e65" }
+hugr-llvm = { git = "https://github.com/quantinuum/hugr", "rev" = "b8ba9ed53b1411ce5f6e9925d1a70772e6c44e65" }
```
aborgna-q: Leaving this as a request for changes until `hugr` gets released.

aborgna-q (changes requested):
LGTM.
Wait for the hugr update before merging.

acl-cqc:
I've merged with #1305, this still fixes one guppy-opt test that was failing after that :).

Also added a couple of comments to `RedundantOrderEdgesPass` about what/why, and where we might go in the future.

acl-cqc:
*patched version overrides removed
*versions updated, including in tket-exts and tket-eccs

aborgna-q (approved):
Change the title to a `fix!` so it goes into the changelog

PR #1743, opened 2026-06-23
docs: Add docs to globals llvm lowering

hsemenenko:


tket-qsystem/src/llvm/globals.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(missing_docs)]
+//! LLVM lowering implementations for "tket.globals" operations.
```
nicolaassolini-qntm, line 1: Maybe add two lines referring to `GlobalsCodegenExtension` `new` and `with_no_global_error`?

nicolaassolini-qntm (approved):
just a suggestion

Issue #1747, opened 2026-06-24
[Bug]: PytketDecodeError (DuplicatedParameter) on guppy_opt/angles.flat.hugr


acl-cqc:
### Description

Since #1305, `angles` variant of `optimize_flattened_guppy` is failing with a PytketDecodeError (as Actual Behaviour, below):

### Steps to Reproduce

Run the indicated test:
https://github.com/Quantinuum/tket2/blob/b3be2bb093d7b6e4b2d50de852aeaf0cafbf244a/tket-qsystem/tests/guppy_opt.rs#L95-L96

### Expected Behaviour

The test should fail with an `xfail` message (the expected gate counts will need updating)

### Actual Behaviour

Test fails with `PytketDecodeError { inner: DuplicatedParameter { param: "p2_0_out0" }, pytket_op: Some("Barrier"), hugr_op: None }`

### Environment

```markdown

```

### Additional Context

Test still fails the same way even with the middle part removed from `fn run_pytket`:
```
@@ -56,14 +56,6 @@ fn run_pytket(h: &mut Hugr, pass_json: &str) {
     let mut encoded =
         EncodedCircuit::new_with_entrypoint(h, h.module_root(), encode_options).unwrap();
 
-    encoded
-        .par_iter_mut()
-        .for_each(|(_region, serial_circuit)| {
-            let mut circuit_ptr = Tket1Circuit::from_serial_circuit(serial_circuit).unwrap();
-            Tket1Pass::run_from_json(pass_json, &mut circuit_ptr).unwrap();
-            *serial_circuit = circuit_ptr.to_serial_circuit().unwrap();
-        });
-
     encoded
         .reassemble_inplace(
             h,
```

acl-cqc:
Ok, so adding this serialization roundtrip in the middle avoids the decode error:
```
@@ -114,6 +113,20 @@ fn optimize_flattened_guppy(#[case] name: &str, #[case] xfail: Option<Vec<(&str,
     // We don't need NormalizeGuppy to "flatten" control flow here, but we still want
     // to get rid of other guppy artifacts.
     NormalizeGuppy::default().run(&mut hugr).unwrap();
+    let p: std::sync::Arc<Path> = Path::new(&format!(
+        "../test_files/guppy_examples/normalized_{}_1747.hugr",
+        name
+    ))
+    .into();
+    let file = std::fs::File::create(&*p).unwrap();
+    hugr.store(file, hugr::envelope::EnvelopeConfig::default())
[...]

PR #1754, opened 2026-06-26
feat!: include InlineFuncsPass in NormalizeGuppy and improve pass ordering

acl-cqc:
We couldn't do this before the [Hugr changes to InlineDFG](https://github.com/quantinuum/hugr/pulls/3058) because the `angles` variant of `optimize_flattened_guppy` (that was [failing](#1747) since #1305) starts to take, like, forever...adding literally 1000s of order edges.

But now, with inlinedfg change + this, that test now pretty much works (modulo Measure/Free ops that we have no way to remove). Indeed almost all the `optimize_flattened_guppy` tests do...

Tested with guppy via monohugr and everything still passed (there are some tests in guppy repo that call NormalizeGuppy but I think none of the asserts they make afterwards are affected by this PR).

API for the heuristic is that the rust bindings take an option (None => don't inline, else a heuristic); the public python API takes an explicit heuristic or a bool (True => default heuristic, False => don't inline).

BREAKING CHANGE: (rust) NormalizeGuppyErrors is now `non_exhaustive`


tket/src/passes/guppy.rs
```diff
@@ -174,6 +192,7 @@ impl<H: HugrMut<Node = Node> + 'static> ComposablePass<H> for NormalizeGuppy {
 
 /// Errors that can occur during the guppy-program normalization process.
 #[derive(derive_more::Error, Debug, derive_more::Display, derive_more::From)]
+#[non_exhaustive]
```
acl-cqc, line 196: Surprised to find this not already there so maybe we don't want this?

tket-py/src/passes.rs
```diff
@@ -88,7 +88,10 @@ create_py_exception!(
 /// - remove_redundant_order_edges: Whether to remove redundant order edges.
 /// - squash_borrows: Whether to squash return-borrow pairs on BorrowArrays.
 #[pyfunction]
-#[pyo3(signature = (circ, *, resolve_modifiers = true, simplify_cfgs = true, remove_tuple_untuple = true, constant_folding = true, remove_dead_funcs = true, inline_dfgs = true, remove_redundant_order_edges = true, squash_borrows = true, scope = None))]
+#[pyo3(signature = (circ, *, resolve_modifiers = true, simplify_cfgs = true,
+    remove_tuple_untuple = true, constant_folding = true, remove_dead_funcs = true,
+    inline_dfgs = true, inline_funcs = Some(Default::default()),
```
acl-cqc, line 93: Seemed like a good opportunity to line wrap these.

Or we could just drop all the defaults? There is only one call, from the wrapper, as this is inside `_passes` not exposed. I'd be happy to do that...

tket-py/test/test_count_ops.py
```diff
@@ -30,7 +30,7 @@ def load_example(example_name: str) -> CompilationState:
     ("loop_conditional", 5),
     ("conditional_loop", 5),
     ("fn_calls", 2),
-    ("repeat_until_success", 19),
+    ("repeat_until_success", 20),
```
acl-cqc, line 33: The count here comes from `CompilationState.num_operations`: https://github.com/Quantinuum/tket2/blob/b3be2bb093d7b6e4b2d50de852aeaf0cafbf244a/tket-py/src/state/base.rs#L177-L180
which thus counts only ops beneath the entrypoint.
The example has 20 gates before NormalizeGuppy; following #1742 NormalizeGuppy started reducing that to 19. The test contains a call to a small function "read"; that contains one "op" and will now be inlined beneath the entrypoint, so this seems expected.

But yeah, this kind of test has limitations, it is pretty opaque....

tket-py/tket/_tket/passes.pyi
```diff
@@ -27,6 +27,7 @@ def normalize_guppy(
     constant_folding: bool = True,
     remove_dead_funcs: bool = True,
     inline_dfgs: bool = True,
+    inline_funcs: inline_funcs.InlineFuncsHeuristic | None = inline_funcs.MaxSize(64),
```
acl-cqc, line 30: This seems to kind of mirror what's in `passes.rs` but alternatively we could drop the default in both

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -91,10 +91,9 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
 
 #[rstest]
 #[case::nested_array("nested_array", None)]
-// Following is https://github.com/Quantinuum/tket2/issues/1747:
-#[should_panic = "PytketDecodeError { inner: DuplicatedParameter"]
+#[should_panic = "xfail"]
```
acl-cqc, line 94: I've written a fair bit about the bug in #1747, but I couldn't isolate the failing Hugr to make a standalone repro due to the serialization weirdness detailed in the bug.

tket-py/src/passes.rs
aborgna-q: Defaults are mostly for documentation, I wouldn't remove them.

Hopefully we can move to pyo3 generated stubs soon, so there's no duplication in the `.pyi`.

tket-py/tket/passes/__init__.py
```diff
@@ -141,6 +144,13 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
         """Run the pass in the CompilationState
 
         TODO: This should be part of a protocol."""
+        inline_funcs_heuristic = (
+            inline_funcs.MaxSize(64)
+            if self.inline_funcs is True
+            else None
+            if self.inline_funcs is False
+            else self.inline_funcs
+        )
```
aborgna-q: Would probably be cleaner as a `match` statement

tket/src/passes/guppy.rs
aborgna-q: Yes!

Though it makes the PR breaking, so the title needs updating

aborgna-q (approved):

Issue #1750, opened 2026-06-25
[Tracking]: Pass sets organization


aborgna-q:
This issue tracks the refactor of meta-pass definitions and organization in
tket, primarily targeted at supporting different kinds of guppylang developers
when compiling their guppy programs and sending them to emulator or hardware
backends.

The goal of this tracking issue is to clearly define the responsibilities of each step, and organize the API accordingly.
There will be some further fine-tuning required for the sub-pass execution order inside each meta-pass, but that is out of scope.

## Initial state of pass sets

There are two main passes run on a guppy-generated HUGR, each composed of multiple smaller passes.

- `NormalizeGuppy` simplifies the control flow structure, and eliminates artifacts from the guppy compiler. This is not tied to a specific system target.
	- Will soon be run automatically when compiling a guppy program (https://github.com/Quantinuum/guppylang/issues/1843).
	- Internally, this runs
	      - [optional] ModifierResolverPass
	      - [optional] NormalizeCFGPass
	      - [optional] UntuplePass
	      - [optional] ConstantFoldPass
	      - [optional] RemoveDeadFuncsPass
	      - [optional] InlineDFGsPass
	      - [optional] BorrowSquashPass
	      - [optional] RedundantOrderEdgesPass
- `QSystemPass` contains lowerings to native hardware operations, and prepares the HUGR to be lowered to LLVM.
	- This includes
	      - ModifierResolverPass
	      - [optional] MonomorphizePass, RemoveDeadFuncsPass
	      - LowerTketToQSystemPass
	      - LowerDropsPass
	      - [optional] hide_non_pub_funcs
	      - [optional] ConstantFoldPass
	      - [optional] force_order

Some other structural passes that are not currently included either:

- `InlineFunctions`

Additionally, we have quantum optimization operations which currently must be
run explicitly:

- `Tket1Pass` (Python only, based on Rust's `EncodedCircuit`s). This pass allows
    users to run any pytket v1 optimization on a HUGR, and as such may be used
    both on generic quantum operations and on native gates.

---

# Proposed changes

Leave `NormalizeGuppy` unchanged in the first iteration. Split `QSystemPass` into two separate steps:

```mermaid
graph TB
    A[Guppy-generated HUGR]:::textOnly --> B[`NormalizeGuppy`]
    B --> C[Generic quantum optimizations]:::textOnly
    C --> D[`QSystemNativePass`]
    D --> E[Native-gate / platform-aware optimizations]:::textOnly
[...]

PR #1758, opened 2026-06-26
feat!: Split QSystemPass into QSRebasePass and QSLLVMPass

### LINKED ISSUE #1467: [Feature]: Remove backwards compatibility for module-rooted Hugrs in `QSystemPass`
### Problem Statement

When passing a module-rooted Hugr, `QSystemPass` looks for a function named `main` and overrides the entrypoint of the returned program.
https://github.com/Quantinuum/tket2/blob/fa2ce9b129895f69fd23429212cfc6fef7f642e7/tket-qsystem/src/lib.rs#L288-L293

This was done for backwards-compatibility with Hugrs defined before we introduced _entrypoints_,
as normally after executing this pass we give it to the LLVM emitter that expects an executable entrypoint.

Can we remove this now? The behaviour is counter-intuitive for users that just want to run a pass on their hugr.
If still needed, we should explicitly set the entrypoint before the LLVM lowering instead of hiding it inside here.

### Proposed Solution

Remove the linked backwards compatibility lines, and remove the line saying that module-rooted hugrs are not supported by the pass.

### Component

Rust library

### Effort Estimate

### Additional Context

_No response_

### LINKED ISSUE #1627: [Feature]: Expose QSystem rebase as its own pass (and other subpasses)
### Problem Statement

At present if you want *just* to rebase to QSystem gates, you have to run a "whole" QSystemPass but with all the other steps set to `false`.

### Proposed Solution

It'd be much better to make this (and perhaps others) available individually

### Component

Python library

### Effort Estimate

None

### Additional Context

_No response_

### LINKED ISSUE #1727: [Bug]: tket-qsystem changes can merge with outdated qis-compiler snaps
### Description

A PR changing tket-qsystem that changes qis-compiler snapshots can merge without snapshots up to date. 

Example: #1706 

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

This may be expected by the file filter based CI runs, if so please close.

### Component

CI/Build System

### Environment

```markdown

```

### Additional Context

_No response_

aborgna-q:
Splits `QSystemPass` into two steps:

- A `QSystemRebasePass` that lowers the generic quantum operations into native operations.
- A `QSystemLLVMPass` that prepares the HUGR to be lowered to LLVMIR.

Native-aware optimizations can be run in between the two.

See https://github.com/Quantinuum/tket2/issues/1750 for a longer writeup.

Closes #1627. Closes #1727. Closes #1467.

BREAKING CHANGE: `QSystemPass` was split into `QSystemRebasePass` and `QSystemLLVMPass`.

### DIFF
--- a/tket-py/tket/_tket/passes.pyi
+++ b/tket-py/tket/_tket/passes.pyi
@@ -125,7 +125,7 @@ def qsystem_llvm_pass(
     force_order: bool = True,
     scope: PassScope | None = None,
 ) -> None:
-    """Runs a rust backed pass to convert qsystem ops to llvm ops.
+    """Runs a rust backed pass to prepare the HUGR for LLVM code generation.
 
     :param constant_fold: Whether to perform constant folding.
     :param monomorphize: Whether to monomorphize generic functions.
--- a/tket-py/tket/_tket/passes.pyi
+++ b/tket-py/tket/_tket/passes.pyi
@@ -104,16 +104,31 @@ def resolve_modifiers(
 
 def qsystem_rebase_pass(
     circ: CompilationState,
-    constant_fold: bool = True,
-    monomorphize: bool = True,
-    force_order: bool = True,
+    *,
+    resolve_modifiers: bool = True,
+    lower_drops: bool = True,
     hide_funcs: bool = True,
     scope: PassScope | None = None,
 ) -> None:
     """Runs a rust backed pass to convert quantum ops to qsystem ops.
 
+    :param resolve_modifiers: Whether to resolve modifier operations.
+    :param lower_drops: Whether to lower drop operations.
+    :param hide_funcs: Make all HUGR functions private.
+    """
+
+def qsystem_llvm_pass(
+    circ: CompilationState,
+    *,
+    constant_fold: bool = True,
+    monomorphize: bool = True,
+    force_order: bool = True,
+    scope: PassScope | None = None,
+) -> None:
+    """Runs a rust backed pass to convert qsystem ops to llvm ops.
+
     :param constant_fold: Whether to perform constant folding.
     :param monomorphize: Whether to monomorphize generic functions.
     :param force_order: Whether to enforce total ordering of all HUGR operations.
-    :param hide_funcs: Make all HUGR functions private.
+    :param scope: A scope to control how the pass is applied to HUGR regions.
     """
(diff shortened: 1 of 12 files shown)


tket-py/tket/_tket/passes.pyi
```diff
@@ -104,16 +104,31 @@ def resolve_modifiers(
...
+    circ: CompilationState,
+    *,
     constant_fold: bool = True,
     monomorphize: bool = True,
     force_order: bool = True,
-    hide_funcs: bool = True,
     scope: PassScope | None = None,
 ) -> None:
-    """Runs a rust backed pass to convert quantum ops to qsystem ops.
+    """Runs a rust backed pass to convert qsystem ops to llvm ops.
```
acl-cqc: This is not about ops but prepping for LLVM lowering - as you've doc'd elsewhere

tket-qsystem/src/passes/rebase.rs
```diff
@@ -0,0 +1,156 @@
...
+/// Errors reported while lowering a HUGR to native QSystem operations.
+#[derive(Error, Debug, Display, From)]
+#[non_exhaustive]
+pub enum QSystemRebasePassError {
+    /// Error while resolving modifier operations.
+    ModifierResolver(ModifierResolverErrors),
+    /// An error from the component [`LowerTketToQSystemPass`] pass.
+    LowerTk2Error(LowerTk2Error),
+    /// An error from the component [`LowerDropsPass`] pass.
+    LinearizeArrayError(ReplaceTypesError),
```
acl-cqc: I'd call this either `LowerDropsError` or `ReplaceTypesError`, or if you're really brave, make it contain `LinearizerError`

acl-cqc (commented):
I have to wonder whether it's really worth exposing QSystemLLVMPass to python, do folks really want to do anything after that? And if we let them call QSysLLVMP in python, we'll only do it *again* later, which should be idempotent, sure.

And, well, it's a shame to have that compatibility pass, but ok ;)

aborgna-q:
> I have to wonder whether it's really worth exposing QSystemLLVMPass to python, do folks really want to do anything after that? And if we let them call QSysLLVMP in python, we'll only do it again later, which should be idempotent, sure.

Let's make it private for now (`_QSystemLLVMPass`). That way we can use it for testing if necessary, but users don't see it in the pass list.

acl-cqc (approved):
Private works well as the pyo3 binds are hidden already, thanks! :)

PR #1768, opened 2026-06-30
fix: RedundantOrderEdges pass finding loops when there are none

aborgna-q:
The pass had an `assert` to fail when an edge is reached multiple times, claiming that would be a cycle.

This PR removes the assert, and adds a test showing a case where the same order edge is reached from multiple paths (it should not affect the rest of the pass, since edges are just added to a map).

### DIFF
--- a/tket/src/passes/redundant_order_edges.rs
+++ b/tket/src/passes/redundant_order_edges.rs
@@ -87,7 +87,6 @@ impl RedundantOrderEdgesPass {
                     for (&other_pred, &other_pred_port) in order_preds.iter() {
                         if reaching_pred.contains(&(other_pred, other_pred_port)) {
                             // `other_pred` reaches predecessor `order_pred` of child, and has a direct edge to child.
-                            assert!(!to_remove.contains_key(&(*order_pred, child))); // this would mean a cycle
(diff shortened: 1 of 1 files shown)


tket/src/passes/redundant_order_edges.rs
```diff
@@ -177,6 +176,8 @@ mod tests {
     #[case([("input", "noop1"), ("noop1", "noop5"), ("input", "noop4"), ("noop4", "noop5"), ("noop5", "output"),
             ("input", "noop5"), ("noop1", "output"), ("noop4", "output")],
            [("input", "noop1"), ("noop1", "noop5"), ("input", "noop4"), ("noop4", "noop5"), ("noop5", "output")])]
+    #[case([("noop1", "noop4"), ("noop4", "noop2"), ("noop2", "noop5"), ("noop4", "noop5"), ("noop1", "noop5")],
```
acl-cqc, line 179: Could also replace noop1->noop4 with noop1->noop2, maybe worth having both

acl-cqc (approved):
Yes, this occurs not just on a cycle, but simply on a DAG merge, thanks @aborgna-q ! :)

PR #1771, opened 2026-06-30
feat: Improve InlineFuncs size heuristic

aborgna-q:
Improves the `MaxSize` heuristic for function inlining by checking transitive caller's sizes.
That prevents us from exponential size grow when multiple small functions call each other.

The check counts an estimate size for the descendants if they were to be inlined, and adds it to the size of the caller function. This prevents a panic in the `tests.integration.std.test_random-test_pcg32_compile.hugr` where we ended up generating too many calls to a function declaration, due to the callers being copied lots of times.

```
pyo3_runtime.PanicException: Outgoing port count exceeds maximum.
portgraph-0.16.1/src/portgraph.rs:548:9
```

Bumps the default MaxSize from 64 to 128 to account for these added counts.

### DIFF
--- a/tket/src/passes/inline_funcs.rs
+++ b/tket/src/passes/inline_funcs.rs
@@ -62,6 +62,14 @@ fn estimated_inline_size<H: HugrView>(
     estimated_inline_size_inner(func, hugr, cache, &mut HashSet::new(), limit)
 }
 
+/// Recursive worker for [`estimated_inline_size`].
+///
+/// Arguments:
+/// - `cache` memoizes completed function estimates across calls in the pass,
+/// - `visiting` tracks the current recursion stack so cyclic call graphs are
+///   treated as over budget.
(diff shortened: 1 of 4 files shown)


tket/src/passes/inline_funcs.rs
```diff
@@ -32,17 +32,72 @@ pub enum InlineFuncsHeuristic {
...
+    hugr: &H,
+    cache: &mut HashMap<H::Node, usize>,
+    visiting: &mut HashSet<H::Node>,
+    limit: usize,
+) -> usize {
+    if let Some(size) = cache.get(&func) {
+        return *size;
+    }
+    if !visiting.insert(func) {
+        return limit.saturating_add(1);
```
nicolaassolini-qntm, line 84: smart way to detect loops, but I would add a comment

nicolaassolini-qntm (approved):
one little thing

PR #1782, opened 2026-07-01
fix: merge_basic_blocks keeps Order edges

### LINKED ISSUE #1779: [Bug]: Simplify CFGs is unsound and makes outputs unreachable
### Description

Running simplify cfgs (most often as part of `NormalizeGuppy`) produces unsound results and makes output statements unreachable / non-executed.

### Steps to Reproduce

```python
from guppylang.emulator import EmulatorBuilder
from guppylang.std.platform import output
from guppylang import guppy


@guppy
def main() -> None:
    if True:
        pass
    output("yah", 1)

pkg = main.compile()
NormalizeGuppy(
    simplify_cfgs=True, # False -> Output there, True -> Output not there
)(pkg.modules[0])
EmulatorBuilder().build(pkg, n_qubits=1).run()
```
Toggling the simplification on / off makes the result `yah` not show up / show up.

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown
guppylang                 1.0.0a8
guppylang-internals       1.0.0a8
hugr                      0.18.0
pytket                    2.18.0
selene-hugr-qis-compiler  0.4.0
tket                      0.15.0
tket-eccs                 0.6.0
tket-exts                 0.14.0
```

### Additional Context

_No response_

acl-cqc:
closes #1779. The main cause was that the Order edges indicating that the `result` op needed to be preserved, were being destroyed by the action of `merge_basic_blocks`.

This is about three lines of "fix" and the rest all tests - both new and updates:
* Added a rust builder version of the testcase in #1779
* Also corresponding rust tests that the other two cases of NormalizeCFGs (moving stuff out of the head/tail of a CFG, keeping the rest of the CFG) are preserving order edges ok (they were).
* Extended `guppy_opt.rs` tests to (a) improve the xfail mechanism and (b) check we are not optimizing any `result` ops away....and we were, but seems they are all there now, phew! :).
* Finally, in `test_count_ops.py` the `repeat_until_success` test had been optimizing away the `result<int>` node (the value returned by the CFG). I wonder whether we can do more here, "count ops" is not really very good....

### DIFF
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -114,14 +114,17 @@ fn optimize_flattened_guppy(#[case] name: &str, #[case] xfail: Option<Vec<(&str,
     // to get rid of other guppy artifacts.
     NormalizeGuppy::default().run(&mut hugr).unwrap();
     run_pytket(&mut hugr, CLIFFORD_SIMP_STR);
-    let should_xfail = xfail.is_some();
-    let expected_counts = match xfail {
-        Some(counts) => counts.into_iter().map(|(k, v)| (k.into(), v)).collect(),
-        None => count_gates(&load_guppy_circuit(name, HugrFileType::Optimized).unwrap()),
-    };
-    assert_eq!(count_gates(&hugr), expected_counts);
-    if should_xfail {
-        panic!("xfail");
+
+    let actual_counts = count_gates(&hugr);
+    let optimized_counts = count_gates(&load_guppy_circuit(name, HugrFileType::Optimized).unwrap());
+
+    if let Some(expected) = xfail {
+        assert_ne!(actual_counts, optimized_counts);
+        let expected = expected.into_iter().map(|(k, v)| (k.into(), v)).collect();
+        assert_eq!(actual_counts, expected);
+        panic!("xfail")
+    } else {
+        assert_eq!(actual_counts, optimized_counts);
     }
 }
 
--- a/tket-qsystem/tests/guppy_opt.rs
+++ b/tket-qsystem/tests/guppy_opt.rs
@@ -7,6 +7,7 @@ use std::fs;
 use std::io::BufReader;
 use std::path::Path;
 use tket::extension::{TKET_EXTENSION_ID, TKET1_EXTENSION_ID};
+use tket_qsystem::extension::result::EXTENSION_ID as RESULT_EXTENSION_ID;
 
 use hugr::{Hugr, HugrView};
 use rstest::rstest;
@@ -76,7 +77,8 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
     let mut counts = HashMap::new();
     for n in h.nodes() {
         if let Some(eop) = h.get_optype(n).as_extension_op()
-            && [TKET_EXTENSION_ID, TKET1_EXTENSION_ID].contains(eop.extension_id())
+            && [TKET_EXTENSION_ID, TKET1_EXTENSION_ID, RESULT_EXTENSION_ID]
+                .contains(eop.extension_id())
         {
             *counts.entry(eop.qualified_id()).or_default() += 1;
         }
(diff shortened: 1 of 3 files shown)


tket/src/passes/normalize_cfgs.rs
acl-cqc, line 59: We could avoid doing these and just wait for a later InlineDFGsPass, but would need to update a bunch of tests - see 79036fb389e8a1ce646f1c1e329f94520e6e17ab

tket-qsystem/tests/guppy_opt.rs
```diff
@@ -93,15 +95,18 @@ fn count_gates(h: &impl HugrView) -> HashMap<SmolStr, usize> {
...
 ]))]
 #[should_panic = "xfail"]
 #[case::simple_cx("simple_cx", Some(vec![
-    ("tket.quantum.QAlloc", 2), ("tket.quantum.MeasureFree", 2),
+    ("tket.quantum.QAlloc", 2), ("tket.quantum.MeasureFree", 2), ("tket.result.result_bool", 2)
 ]))]
 #[case::nested("nested", None)]
 #[case::ranges("ranges", None)]
-#[case::false_branch("false_branch", None)]
+#[should_panic = "xfail"]
```
acl-cqc, line 106: Yes, so this was passing because we were removing result + qalloc + measurefree, matching no expected ops. When now we keep the result, we have no way of removing qalloc+measurefree, but the "optimized" guppy/hugr doesn't have them, hence we have to xfail

tket/src/passes/normalize_cfgs.rs
```diff
@@ -394,8 +394,7 @@ fn mk_rep<H: HugrView>(
         .collect::<Vec<_>>();
 
     let dfg_order_out = replacement.get_optype(dfg1).other_output_port().unwrap();
-    let order_srcs = (dfg1_outs.is_empty()).then_some((dfg1, dfg_order_out));
-    // Do not add Order edges between DFGs unless there are no value edges
+    let order_srcs = Some((dfg1, dfg_order_out));
```
acl-cqc, line 397: This one-line change exactly parallels the RedundantOrderEdgesPass change in #1768 (following https://github.com/Quantinuum/hugr/pull/3058)

tket/src/passes/normalize_cfgs.rs
```diff
@@ -1017,4 +1021,223 @@ mod test {
...
+        assert_eq!(
+            h.linked_outputs(res2, res2_in).collect_vec(),
+            vec![(res1, res1_out), (inp, inp_out)]
+        );
+    }
+
+    #[rstest]
+    fn order_edges_loop_after() {
+        // BB -> loop_bb -> Exit
+        //        /   \
```
nicolaassolini-qntm: ?
```suggestion
        // BB1 -> loop_bb -> Exit
        //        /   \
```

tket/src/passes/normalize_cfgs.rs
```diff
@@ -1017,4 +1021,223 @@ mod test {
...
+        assert_eq!(h.get_optype(res), &res_op);
+        assert!(h.entrypoint_optype().is_cfg()); // remains, but result nodes in DFG outside:
+        let parent = h.get_parent(res.node()).unwrap();
+        assert!(h.get_optype(parent).is_dfg());
+        assert_order_single_chain(&h, [inp, h.entrypoint(), parent, outp]);
+        let [inp, outp] = h.get_io(parent).unwrap();
+        assert_order_single_chain(&h, [inp, res, outp]);
+    }
+
+    fn assert_order_single_chain<H: HugrView>(h: &H, nodes: impl IntoIterator<Item = H::Node>) {
```
nicolaassolini-qntm, line 1234: I wuold add small comment to explain what we are checking here

nicolaassolini-qntm (approved):
I only have two small comments

tket/src/passes/normalize_cfgs.rs
acl-cqc: Yup, renamed to `entry_bb`

PR #1783, opened 2026-07-01
feat: Metadata propagation policies for `ReplaceTypes`

### LINKED ISSUE #1651: Propagate debug info in NodeTemplate::replace
Replacing an op node with a container node leaves the debug location attached to the container, where it has no effect. The location should be propagated to all ops/calls in the container.

cgh-qtnm:
This PR adds  `tket::passes::replace_types::MetadataPropagationPolicy`, which defines how metadata is handled when a node is replaced with a container by `LinkedHugr` or `CompoundOp`. This is motivated by and fixes #1651.

Each `ReplaceTypes` pass has exactly one associated `MetadataPropagationPolicy`. A policy consists of zero or more rules. A rule has two parts: 

1) a function which takes `(replaced_node_optype, replaced_node_metadata, child_optype, child_metadata)` for each child in the replacing container, and returns a list of key-value pairs to add to that child's metadata.

2) a list of metadata keys to remove from the replacement container node.

Notes:
- Updates are applied to both direct and indirect descendants of the new container (ie, we recurse into sub-containers).
- Updates are applied before `process_subtree_opts`, so that propagation composes recursively. See comment in `replace_types.rs`.
- If multiple rules are defined, all updates to child metadata are applied before removing keys from the replacing container.


tket/src/passes.rs
```diff
@@ -72,7 +72,7 @@ pub use redundant_order_edges::RedundantOrderEdgesPass;
 
 // Replace types, ops, and constants across a HUGR.
 pub mod replace_types;
-pub use replace_types::ReplaceTypes;
+pub use replace_types::{MetadataPropagationPolicy, ReplaceTypes};
```
aborgna-q: This is specific to `ReplaceTypes` and the module is `pub`, so I'd avoid the reexport here to keep the docs clear.


```suggestion
pub use replace_types::ReplaceTypes;
```

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+/// called once per descendant of the replacement container. It
+/// returns metadata key-value pairs to write onto that descendant. All returned
+/// pairs are applied unconditionally, so rules are responsible for checking
+/// `inner_meta` if they want to avoid overwriting existing keys.
+///
+/// `remove_from_old` is a list of keys that
+/// are deleted from the container once all update rules has been applied to its
+/// descendants. Use this when a rule fully relocates metadata onto descendants
+/// and wants to avoid leaving a stale copy on the container. Removals from
+/// all rules are deduplicated and applied once at the very end.
```
aborgna-q: These two names are not defined when looking at the struct's definition.

Image

I'd either move the comment to the method docs, or rewrite these docs defining what is a Rule, and how it is used.

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+>;
+
+#[derive(Clone)]
+struct RuleEntry {
+    update_new: UpdateFn,
+    remove_from_old: Vec<String>,
+}
+
+/// Defines how metadata is propagated from a replaced op node to the inner
+/// nodes of its replacement container.
```
aborgna-q: Does the replacement container node get the original node's metadata?

That should be mentioned here, and include the `remove_from_old` behaviour.

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+            }
+        },
+    );
+    (update, vec![DEBUGINFO_META_KEY.to_string()])
+}
+
+/// Returns a [`MetadataPropagationPolicy`] containing only the built-in
+/// `core.debug_info` rule (see [`default_debuginfo_rule`]). This is the
+/// default policy for [`ReplaceTypes::default`](super::ReplaceTypes).
+pub fn default_debuginfo_policy() -> MetadataPropagationPolicy {
```
aborgna-q: Do we need to expose this, if we already have it as the `Default` impl?
(We can add docs to trait implementations too)

I'm also thinking we may want to add more default rules in the future, so the `debuginfo` may get outdated.

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+/// `inner_meta` if they want to avoid overwriting existing keys.
+///
+/// `remove_from_old` is a list of keys that
+/// are deleted from the container once all update rules has been applied to its
+/// descendants. Use this when a rule fully relocates metadata onto descendants
+/// and wants to avoid leaving a stale copy on the container. Removals from
+/// all rules are deduplicated and applied once at the very end.
+///
+/// Rules are applied in the order they were added.
+#[derive(Clone)]
```
aborgna-q: Add `Debug` too. Rule entry may need a custom impl due to the `dyn`.

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+///
+/// The rule copies `core.debug_info` from a replaced `Call` or `ExtensionOp`
+/// onto every `Call`/`ExtensionOp` descendant of the replacement container
+/// that does not already carry the key, and marks `core.debug_info` for
+/// removal from the container so backends don't see a stale entry on the new
+/// `DFG`/`CFG`/etc.
+///
+/// Exposed as a free function so callers can install it at any position
+/// among their own rules (rule order is significant).
+pub fn default_debuginfo_rule() -> (UpdateFn, Vec<String>) {
```
aborgna-q: We already have a struct definition for the return type.

Let's make `RuleEntry` public, and have this as a method rather than a free-floating function.
(We may be able to hide the `UpdateFn` alias too)

tket/src/passes/replace_types/metadata.rs
```diff
@@ -0,0 +1,733 @@
...
+        self.rules.is_empty()
+    }
+
+    /// Adds a propagation rule.
+    ///
+    /// `update_new` receives `(old_optype, old_meta, inner_optype, inner_meta)`:
+    /// - `old_optype` — optype of the original node that was replaced
+    /// - `old_meta` — metadata of the original node
+    /// - `inner_optype` — optype of a descendant of the replacement container
+    /// - `inner_meta` — current metadata of that descendant
```
aborgna-q: We should specify that `inner_meta` does not include changes done by previous rules. (But perhaps it should?)

aborgna-q (commented):
Nice! I mostly have docs & API design comments.

tket/src/passes/replace_types/metadata.rs
cgh-qtnm: It was a workaround to allow composition with other policies in arbitrary order. But I think it would be cleaner to just make the Rule vec directly mutable.

tket/src/passes/replace_types/metadata.rs
cgh-qtnm: yes, I think it should.

aborgna-q (approved):
Nice, thanks!

PR #1784, opened 2026-07-02
fix: IndirectCall nodes solving in ModifierResolverPass

### LINKED ISSUE #1778: [Feature]: Use signature instead of metadata to filter functions
### Problem Statement

Relying on metadata is not robust for handwritten or optimized hugr, we should check for quantum signature

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

### LINKED ISSUE #1781: [Bug]: ModifierResolverPass fail on Indirect Call node
### Description

ModifierResolverPass fails with:

```python
@guppy(unitary=True)
def foo(q: qubit) -> None:
  h(q)
  with dagger:
    h(q)
  with dagger:
    h(q)

@guppy
def main() -> None:
  q1 = qubit()
  q2 = qubit()
  with dagger:
    foo(q1)
    with control(q1):
      foo(q2)
    cx(q1, q2)
    with control(q1):
      foo(q2)

  discard(q1)
  discard(q2)

modifier_resolver = tket.passes.ModifierResolverPass()
package = main.with_minimal_opt().with_optimization(modifier_resolver).compile()
```

nicolaassolini-qntm:
closes #1778
closes #1781 

As #1778 points out, using metadata to choose which functions need to be modified is not robust enough. Addressing this issue also closes #1781, since the error is related to an imprecision in guppy unitary metadata generation (https://github.com/Quantinuum/guppylang/issues/1984).
While addressing #1778 some changes in `modify_indirect_call` were needed to keep the passes sound when non-quantum functions are modified.


tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -318,22 +318,19 @@ impl<N: HugrNode> ModifierResolver<N> {
...
-            if trace.len() > 1 {
-                return Err(ModifierResolverErrors::unresolvable(
-                    trace[0],
-                    "Cannot modify indirect call.",
-                    indir_call.clone().into(),
-                ));
+        // Modify the function
+        let modified_fn = if trace.len() > 1 {
+            // If the function is target of a modifier chain, we must apply the modifiers to it
+            self.modify_fn(h, func).unwrap()
```
aborgna-q:
```suggestion
            self.modify_fn(h, func).map_err(wrap_resolver_err)?
```

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -332,10 +332,13 @@ impl<N: HugrNode> ModifierResolver<N> {
         h: &mut impl HugrMut<Node = N>,
         func: N,
     ) -> Result<Option<N>, ModifierResolverErrors<N>> {
-        let satisfies = ModifierFlags::from_metadata(h, func)
-            .is_some_and(|flags| flags.satisfies(&self.modifiers));
```
aborgna-q, line 336: We no longer use the UnitaryFlag metadata here. What's the plan now?

Are we dropping the metadata completely? If so, we need to deprecate here and update guppy. There's also leftover code dealing with moving the flags around in this repo.

An alternative would be to keep the metadata as an override logic, but I don't think it would be useful.

nicolaassolini-qntm: > We no longer use the UnitaryFlag metadata here. What's the plan now?

I don't know. I can open an issue, I guess some discussion is required.

tket/src/modifier/modifier_resolver/dfg_modify.rs
aborgna-q: Leave a TODO here explaining the situation with modifier metadata, and linking to #1790

My main issue if we leave the flag around is that we're not testing their valid anymore, so there won't be any checks that other passes don't leave semantically invalid flags when rewriting the hugr.

So I would leave a warning in the metadata definition about it, saying it's ignored and may be incorrect.

nicolaassolini-qntm: Done it, I also removed a useless `TODO` left before `modify_fn_if_needed`

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -312,30 +312,34 @@ impl<N: HugrNode> ModifierResolver<N> {
...
-    //      if only dagger, just check signature
-    //
-    // Also, it may be better to check with the usage (how it is instantiated).
+    /// When the function definition or a concrete call instantiation contains qubits,
+    /// the function needs to be modified. The concrete signature matters for polymorphic
+    /// functions whose uninstantiated definition may not reveal quantum data.
+    /// force is used to indicate that the function is being called after a modifiers chain,
+    /// so it must be modified even if the concrete signature does not contain qubits.
+    ///
+    /// NOTE: When a polymorphic functions has a polymorphic input, the function is considered
```
aborgna-q:
```suggestion
    /// NOTE: When a polymorphic function has a polymorphic input, the function is considered
```

aborgna-q (approved):
:shipit:

PR #1787, opened 2026-07-03
feat: Target platform option in PytketHugrPass

### LINKED ISSUE #1619: [Feature]: Select target extensions in pytket decoder from python
### Problem Statement

When using the pytket encoder/decoder, operatios found in the pytket circuit get extracted back into hugr operations using a configured list of extension decoders.
These are tried in order; if we have both the `tket.quantum` and `tket.qsystem` extension decoders and we try to decode a pytket `Rz` operation then whichever decoder got registered first will generate their own version of `rz`.

This is problematic when we do things like calling `SingleQubitSquash` from python, expecting the resulting Hugr to remain in the `qsystem` gate set. There is currently no way to change the set of decoders from the python API.

This will become even more annoying once https://github.com/Quantinuum/tket2/pull/1567/ is merged, and we have three competing extensions to worry about.

### Proposed Solution

We should consider exposing a configuration enum on the python API that can pick between different sets (and different order) of extension decoders.

aborgna-q:
Closes #1619

Adds a `target` option to `PytketHugrPass` and to `CompilationState.from_tket1` so calling a pass doesn't unexpectedly rebase quantum operations between `tket.quantum` and the native `qsystem` ones.

### DIFF
--- a/tket-py/src/passes/tket1.rs
+++ b/tket-py/src/passes/tket1.rs
@@ -1,14 +1,16 @@
 //! Passes that call to tket1-passes using the tket-c-api.
 
+use std::str::FromStr;
+
 use rayon::iter::ParallelIterator;
 use std::sync::Arc;
 use tket::passes::composable::PassScope;
-use tket_qsystem::QSystemPlatform;
+use tket_qsystem::PlatformTarget;
 
 use crate::passes::PyPassScope;
+use pyo3::exceptions::PyValueError;
 use pyo3::prelude::*;
 use tket::serialize::pytket::{EncodeOptions, EncodedCircuit};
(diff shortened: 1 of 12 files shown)


cqc-alec (approved):
Looks good, though it's not quite clear to me why we need the string forms instead of using enums throughout. Is this a limitation of pyo3?

aborgna-q:
> (...) it's not quite clear to me why we need the string forms instead of using enums throughout. Is this a limitation of pyo3?

We can define a enum with `pyo3` _derives_ and have it exposed to python.
This works fine, but causes a bit of friction with some python tools since the definitions are not quite like pure-python ones (same reason as why we wrap the structs).

Strings here are the simples way to do it. We should be able to remove the duplicated `.pyi` definitions too, with pyo3's experimental typing hints support.

Issue #1415, opened 2026-02-23
[Feature]: WASM wheels for TKET 2


aidanCQ:
### Problem Statement

I don't expect this to be easily solveable or achievable in the near term but having WASM wheels would improve web interoperability. I appreciate pytket links against c++ libs so compiling those may be a headache.

### Effort Estimate

None

jpacold:
Hi -- I've been working on this intermittently because it seemed like an interesting build-systems exercise. [This branch of `tket`](https://github.com/Quantinuum/tket/compare/main...jpacold:tket:emscripten) has a new conan recipe and a one-line code change needed to build `tket-c-api` and its dependencies with `emcc`. I built a `tket2` pyodide wheel that links against them, along with wheels for `hugr`, `tket_eccs`, and `tket_exts`. I confirmed that I can use Node to:

* import `tket` (failed at first due to a dependency on `pytket`, but [#1507](https://github.com/Quantinuum/tket2/pull/1507) [link fixed] resolved this 🙂)
* run a `RemoveRedundancies` pass

I have a couple of questions before continuing (assuming there is still a need for this):

* What would CI/CD look like for this? For example, where would `tket-c-api.a` and its dependencies live?
* There is some multithreading (for example Rayon's `par_iter_mut`) that will not work with WASM. Would you rather have a feature flag that switches to single-threaded implementations, or just document any impacted methods as unsupported?

aborgna-q:
Hi @jpacold. Thanks for the comment, we weren't expecting to have such a clean solution!

Here in the tket2 CI we use pre-compiled tket-c-api builds from the https://github.com/Quantinuum/hugrverse-env repository. (These are artifacts that rarely need rebuilding, so we use that as static storage).
We also don't use conan over there, just building with cmake instead.

I'll investigate if we can build the emscripten image there. Would you mind contributing the `Path.cpp` fix to https://github.com/Quantinuum/tket?
I think the conan config would also be valuable, even if we don't depend on it for CI.

aborgna-q:
> There is some multithreading (for example Rayon's par_iter_mut) that will not work with WASM. Would you rather have a feature flag that switches to single-threaded implementations, or just document any impacted methods as unsupported?

Yes. We'll probably want to move those definition to default features, that we disable for the wasm builds.

jpacold:
Sorry I took this long to follow up -- it's been a busy few months, including getting up to speed at the Broomfield office :)

[...]

PR #1788, opened 2026-07-03
ci: Build WebAssembly (wasm32-unknown-emscripten) wheels

aborgna-q:
Adds a `wasm` job to the Python wheels workflow that cross-compiles the tket Python extension to `wasm32-unknown-emscripten`, producing pyodide-compatible WebAssembly wheels. Implements Quantinuum/tket2#1415.

Targets **Emscripten 4.0.9 / Python 3.13** (pyodide `pyemscripten_2025_0` ABI), to match hugr's wasm wheels

---

I tested these locally with emscripten, and in a guppy-playpond instance.
Note that the `pytket` python wheel is not available in WASM, but `tket`'s pytket-based pass work correctly as they use an internally compiled `tket-c-api`.

Wheel building run: https://github.com/Quantinuum/tket2/actions/runs/33096483912/job/98602665474


.github/workflows/python-wheels.yml
```diff
@@ -251,6 +264,79 @@ jobs:
...
+        uses: actions/setup-node@v6
+        with:
+          node-version: "18"
+      - name: Setup emsdk manually
+        # The emscripten version must match both the version used to build the
+        # tket-c-api wasm target in hugrverse-env and the pyodide ABI.
+        # See <https://doc.rust-lang.org/beta/rustc/platform-support/wasm32-unknown-emscripten.html>
+        # and <https://github.com/pyodide/pyodide/blob/main/docs/development/abi.md>
+        run: |
+          git clone --depth 1 https://github.com/emscripten-core/emsdk.git
```
doug-q: I think it would be good to pin a commit here

tket-py/src/passes/tket1.rs
```diff
@@ -70,15 +71,17 @@ pub(crate) fn tket1_pass(
...
-        .try_for_each(|(_, circ)| -> Result<(), tket1_passes::PassError> {
-            let mut tk1_circ = tket1_passes::Tket1Circuit::from_serial_circuit(circ)?;
-            tket1_passes::Tket1Pass::run_from_json(pass_json, &mut tk1_circ)?;
-            *circ = tk1_circ.to_serial_circuit()?;
-            Ok(())
-        })
-        .convert_pyerrs()?;
+    // Emscripten wheels are built with `rayon` disabled, so we need to run the
+    // pass sequentially in that case.
+    if cfg!(target_os = "emscripten") {
```
doug-q: This is a weird way to do it. Rayon doesn't look like it's disabled in Cargo.toml.

doug-q (approved):
Looks good, just a couple comments

.github/workflows/python-wheels.yml
```diff
@@ -251,6 +264,79 @@ jobs:
...
+    if: ${{ needs.check-tag.outputs.run == 'true' }}
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+      - uses: mozilla-actions/sccache-action@v0.0.10
+      - uses: actions/setup-python@v6
+        with:
+          python-version: ${{ env.WASM_PYTHON_VERSION }}
+      - name: Install nightly Rust toolchain
+        uses: dtolnay/rust-toolchain@master
```
cqc-alec:
```suggestion
        uses: dtolnay/rust-toolchain@stable
```

cqc-alec (approved):

tket-py/src/passes/tket1.rs
aborgna-q: Yeah, it's mostly that running `par_iter_mut` on emscripten fails at runtime:

```
The global thread pool has not been initialized:
ThreadPoolBuildError {
    kind: IOError(Os {
        code: 6,
        kind: WouldBlock,
        message: "Resource temporarily unavailable"
    })
}
```

I'll update the comment

PR #1792, opened 2026-07-03
fix: ModifierResolverPass preserve function edges into classical DAG

### LINKED ISSUE #1789: [Bug]: Modifying higher order classical function after normalization produce invalid hugrs
### Description

simplify_cfgs in NormalizeGuppy is transforming a CFG with a dataflowblock in a DFG. This causes the ModifierResolverPass to produce invalid hugr when a classical function is modified. This is linked to #1774 

### Steps to Reproduce

```python


@guppy
def inner(mk_struct: "Callable[[int], int]", x: int) -> int:
    return mk_struct(x)


@guppy
def foo(i: int) -> int:
    return i + 1


@guppy
def main() -> None:
    c1 = qubit()

    # testing that a classical higher order function can be called inside a modified context
    with dagger, control(c1):
        inner(foo, 2)

    discard(c1)


program = main.compile()

normalize = NormalizeGuppy(
    remove_tuple_untuple=False,
    remove_dead_funcs=False,
    remove_redundant_order_edges=False,
    squash_borrows=False,
    inline_funcs=False,
    resolve_modifiers=True,
    simplify_cfgs=True,
    constant_folding=False,
    inline_dfgs=False,
)
program = normalize(program.modules[0])


```

nicolaassolini-qntm:
closes #1789 

When a DFG classical block is found, the DFG is copied as it is. When the DFG contained a non-local edge (a function call), an invalid hugr was generated since the edge was not copied.


tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -456,22 +456,50 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+        dfg: &mut impl Container,
     ) -> Result<Node, ModifierResolverErrors<N>> {
-        // Some containers have qubits in their signature but only pass them
-        // through while doing classical work. Copying the whole subtree keeps
-        // those classical dependencies intact instead of trying to dagger the
-        // boundary one port at a time.
-        let insertion_result = new_dfg.add_hugr_view(&h.with_entrypoint(n));
+        let nodes = h
+            .descendants(n)
+            .chain(iter::once(n))
```
acl-cqc: Isn't the first element of `descendants(n)` the root node `n`? Or do you need this to come last?

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -456,22 +456,50 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+            .iter()
+            .flat_map(|node| {
+                h.node_inputs(*node)
+                    .filter(|port| {
+                        matches!(
+                            h.get_optype(*node).port_kind(*port),
+                            Some(EdgeKind::Function(_))
+                        )
+                    })
+                    .filter_map(|port| {
```
acl-cqc: You've already filtered the inports down to those of kind `EdgeKind::Function`, and the only source of such edges is a FuncDefn/FuncDecl; and FuncDefn/FuncDecl nodes [can only be children of the module root](#2256), not inside a DFG. (Yes, that PR badly titled - it applies to Decls too.)

But, there are also *constant* edges, which come only from Const nodes, which can be anywhere...

And there are nonlocal value edges...which the PR title suggests you want, but the variable name `static_edges` suggests you don't....please make these consistent....

If you want static edges, you should do this as (either the port is of kind EdgeKind::Function, OR it's EdgeKind::Constant and the single_linked_output has a different parent)...

Or perhaps you do only want Function edges, and are being extremely conservative wrt. handling invalid Hugrs, but if you really need that, I think that's worth a comment or two...(in particular, wtf end result you want ;))

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -456,22 +456,50 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+
+        let insertion_result = dfg.add_hugr_view(&h.with_entrypoint(n));
 
         let new_node = insertion_result.inserted_entrypoint;
         for port in h.all_node_ports(n) {
             self.map_insert(DirWire(n, port), DirWire(new_node, port))?;
         }
+        for (source, old_target, target_port) in static_edges {
+            let Some(new_target) = insertion_result.node_map.get(&old_target).copied() else {
+                return Err(ModifierResolverErrors::unreachable(format!(
```
acl-cqc: this verges on `unwrap`, I mean it really should be there

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -1451,9 +1479,42 @@ mod test {
...
     }
 
+    /// Test pass on a DFG with no quantum signature that calls an external function
+    fn foo_dfg_external_function_call(
+        module: &mut ModuleBuilder<Hugr>,
+        t_num: usize,
+    ) -> FuncID<true> {
+        assert_eq!(t_num, 1);
+
+        let external = {
```
acl-cqc: nit: You can just do
```
let external = module
   .define_function("external_classical_noop", Signature::new_endo([]))
   .unwrap()
   .finish_with_outputs([])
   .unwrap()
```

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -1451,9 +1479,42 @@ mod test {
...
+            func.finish_with_outputs([]).unwrap()
+        };
+
+        let foo_sig = Signature::new_endo([qb_t()]);
+        let mut func = module.define_function("foo", foo_sig).unwrap();
+        func.set_unitary();
+        let q = func.input_wires().next().unwrap();
+        {
+            let mut dfg = func
+                .dfg_builder(Signature::new(type_row![], type_row![]), [])
```
acl-cqc: suggest `new_endo([])` again

tket/src/modifier/modifier_resolver/dfg_modify.rs
acl-cqc: However see comment on `insertion_result` above

tket/src/modifier/modifier_resolver/dfg_modify.rs
```diff
@@ -456,22 +456,50 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+                    })
+                    .filter_map(|port| {
+                        h.single_linked_output(*node, port)
+                            .filter(|(source, _)| h.get_parent(*node) != h.get_parent(*source))
+                            .map(|(source, _)| (source, *node, port))
+                    })
+            })
+            .collect::<Vec<_>>();
+
+        let insertion_result = dfg.add_hugr_view(&h.with_entrypoint(n));
```
acl-cqc: Rather than doing `add_hugr_view` and patching up the edges yourself, see `insert_link_hugr_by_node` where you can specify a map from nodes in the source hugr (i.e. the sources of the static edges) to nodes in the target hugr (`NodeLinkingDirective::UseExisting(func)`) - in this case the same Node on both sides of the map. That'll connect the static edges as you want.

acl-cqc (commented):
Looks like a good thing to do, but not sure we're quite doing the right thing, and we can probably do it neater ;)

tket/src/modifier/modifier_resolver/dfg_modify.rs
acl-cqc: Ok so I see a few options:
1. Just EdgeKind::Function. You could assert that the source parent is different. You might also wanna check that for EdgeKind::Constant and EdgeKind::Value, the source parent is the same as the target parent (thus you are OK to skip).

2. Also EdgeKind::Constant, if the source parent is different

Either of those two you could use `insert_hugr_link_by_node` rather than patching up edges yourself.

3. Also EdgeKind::Value, if the source parent is different. This requires manually patching up the edges as you are doing below, and doesn't occur in guppy-generated Hugrs, so maybe go for (1.) or (2.)

If you go for 1/2, the easiest way might be to filter *only* on source_parent != target_parent, and then assert (for every edge that passes filter) that it's EdgeKind::Function (or EdgeKind::Constant).

tket/src/modifier/modifier_resolver/dfg_modify.rs
acl-cqc: Ah, this doesn't work because N != Node... :(

tket/src/modifier/modifier_resolver/dfg_modify.rs
acl-cqc: Pushed!

acl-cqc (approved):
Looks good although see a few nits/simplifications!

Issue #1724, opened 2026-06-18
[Feature]: Support parametrised sum variants for constant folding


maximilianruesch:
### Problem Statement

During #1609, we would have liked to introduce constant folding support for parametrised sum variants (such as variants of `Either`) for cases where we can easily reconstruct the tag tree, i.e. where the exact variant does not have any child values.

But constant folding will happily constant fold these variants, even if the type itself is not copyable, leading to validation errors when inputs are disconnected and cannot just be dropped (like with copyable types).

The code prepared for this was:
```rust
let variants: Result<Vec<_>, _> = sum
                    .sum_type
                    .variants()
                    .map(|rv_row| TypeRow::try_from(rv_row.clone()))
                    .collect();
                variants
                    .map(|vs| hugr.add_node_with_parent(parent, Tag::new(sum.tag, vs)))
                    .ok()
```

### Effort Estimate

None

PR #1793, opened 2026-07-03
fix: prevent Constant Folding disconnecting linear outputs even if constant

acl-cqc:
For a Sum types where some variants are quantum but others are classical, we can deduce a constant value of the classical variant, but "breaking the wire" to insert the constant leads to an unconnected port of the *sum type*, which is linear...

Ideal solution in many cases is probably https://github.com/Quantinuum/tket2/issues/1796, but that may not apply in all cases. Same problem could also happen for CustomConsts of linear type (BorrowArray?) if we constant-fold those.

For now the workaround is to avoid breaking any linear wire. This is overly conservative; we could insert consts and break such wires if the sources can also be removed, but that requires traversal done only in DCE.

First commit adds a test where the Hugr fails validation after constant-folding; second fixes.


tket/src/passes/const_fold.rs
```diff
@@ -142,7 +142,14 @@ impl<H: HugrMut<Node = Node> + 'static> ComposablePass<H> for ConstantFoldPass {
         assert_eq!(root_descs.next(), Some(root), "Could not skip root");
         let wires_to_break = root_descs
             .filter(|n| !hugr.get_optype(*n).is_load_constant()) // no point in adding another Const!
-            .flat_map(|n| hugr.out_value_types(n).map(move |(outp, _ty)| (n, outp)))
+            .flat_map(|n| {
+                hugr.out_value_types(n)
+                    // Do not consider breaking wires from outports of linear type.
+                    // Overly conservative: see https://github.com/Quantinuum/tket2/issues/1794,
+                    // https://github.com/Quantinuum/tket2/issues/1795,
+                    // https://github.com/Quantinuum/hugr/issues/2245
```
acl-cqc: This one needs moving to TKET2 repo.

maximilianruesch:
Note that this may be related / partially solving https://github.com/Quantinuum/tket2/issues/1724

tket/src/passes/const_fold.rs
```diff
@@ -142,7 +142,14 @@ impl<H: HugrMut<Node = Node> + 'static> ComposablePass<H> for ConstantFoldPass {
...
         let wires_to_break = root_descs
             .filter(|n| !hugr.get_optype(*n).is_load_constant()) // no point in adding another Const!
-            .flat_map(|n| hugr.out_value_types(n).map(move |(outp, _ty)| (n, outp)))
+            .flat_map(|n| {
+                hugr.out_value_types(n)
+                    // Do not consider breaking wires from outports of linear type.
+                    // Overly conservative: see https://github.com/Quantinuum/tket2/issues/1794,
+                    // https://github.com/Quantinuum/tket2/issues/1795,
+                    // https://github.com/Quantinuum/tket2/issues/1796
+                    .flat_map(move |(outp, ty)| ty.copyable().then_some((n, outp)))
```
aborgna-q: nit: It could also have been a `.filter(...).map(...)`, but works that same anyways.

aborgna-q (approved):

aborgna-q:
> Note that this may be related / partially solving https://github.com/Quantinuum/tket2/issues/1724

I think so too, we should do some testing.

tket/src/passes/const_fold.rs
acl-cqc: Yeah, longer but clearer, so done

PR #1806, opened 2026-07-06
perf: Improve rule evaluation order in dataflow analysis

aborgna-q:
Fixes a big slowdown due to rule order in the dataflow relationship definitions.
The analysis was run by running ConstantFolding on a big real-world HUGR (81k nodes with complex structure), and using `ascent`'s `#![measure_rule_times]` to extract detailed information on the rule runtimes.

Multiple rules in the dataflow analysis definition started with `node(n)`.
This meant the `node(n)` relation was first enumerated in it's entirety (taking ~1ms) before trimming down the list with the following rules.
In the original code, `node` got called 23k times on each analysis, consuming a total of `17.8s`.

This PR reorders the rule definitions, so they start with smaller sets that are relevant to the delta when updating the lattice relations.

This PR (after #1805) reduced the test times from 17.8s to 295ms.

### DIFF
--- a/tket/src/passes/dataflow/datalog.rs
+++ b/tket/src/passes/dataflow/datalog.rs
@@ -202,11 +202,10 @@ fn run_datalog<V: AbstractValue, H: HugrView>(
 
         // Interpret leaf ops
         out_wire_value(n, p, v) <--
-           node(n),
+           node_in_value_row(n, vs),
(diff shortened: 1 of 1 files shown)


doug-q (approved):
Nice Agustin. This makes  a lot of sense to me.

Issue #1780, opened 2026-07-01
Rename `NormalizeGuppy` to `Normalize`


cqc-alec:
The name arose because it was originally intended to clean up artifacts introduced by guppy compilation, but its applicability is more general so it doesn't make much sense to have "Guppy" in the name.

The scope of this pass is basic simplification of the classical structure, without touching the quantum structure and without doing costly search-based optimization. `Normalize` describes this well enough.

PR #1807, opened 2026-07-06
chore: Add benchmarks for NormalizeGuppy

aborgna-q:
Adds some benchmarks for NormalizeGuppy, so we can track perf changes in codspeed.

Adds a t-factory example extracted from guppy's example notebooks, so we have a hugr with a bit of everything for the pass to run on.

Named the benchmark `normalize` rather than `normalize_guppy` in preparation for #1780, so we don't lose the bench history when renaming it.


test_files/guppy_examples/t_factory.py
```diff
@@ -0,0 +1,78 @@
+# /// script
+# requires-python = ">=3.13"
+# dependencies = [
+#     "guppylang==1.0.0a5",
+#     "guppylang-internals==1.0.0a5",
```
aborgna-q, line 5: Using the same guppy version as other examples. We'll update all to the `rc` later.

tket/benches/benchmarks/normalize.rs
```diff
@@ -0,0 +1,56 @@
...
+        ),
+    ];
+
+    for (name, bytes) in examples {
+        let hugr = load_hugr(bytes);
+        group.bench_function(name, |b| {
+            b.iter_batched(
+                || hugr.clone(),
+                |mut hugr| {
+                    NormalizeGuppy::default().run(&mut hugr).unwrap();
```
cqc-alec, line 40: Wonder if it's worth being explicit about the options for `NormalizeGuppy`, in case the defaults change or others are added. But probably no big deal as we should be aware when that happens.

cqc-alec (approved):

tket/benches/benchmarks/normalize.rs
aborgna-q: Yeah, I think we'll want to see the change in runtime for the default options primarily.

PR #1808, opened 2026-07-06
fix: `ModifierResolverErrors::UnResolvable` display error message

### LINKED ISSUE #1748: [Bug]: `ModifierResolverErrors` display message
### Description

`ModifierResolverErrors`, in particular `UnResolvable` are quite imprecise, and something contains repetitions.

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1748

This Pr also slightly refactor `modify_tket_op` to avoid redundant tests


tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -658,6 +691,72 @@ mod test {
         }
     }
 
+    #[test]
+    /// Test that when no modifiers are applied (e.g. with an even number of daggers), non quantum operations are handled correctly.
+    fn double_dagger_allows_measurement_function() {
```
doug-q: This sounds like a misfeature. 

it only makes sense looking at the program globally, because you have to see all callsites of all functions. 

I think it would be wiser to disallow any measures in a dagger context

No action required, clearly it is already like this.

doug-q (approved):

tket/src/modifier/modifier_resolver/tket_op_modify.rs
nicolaassolini-qntm: I rephrased the comment, since the reference to even daggers is quite misleading

PR #1809, opened 2026-07-07
refactor!: remove `ModifierFlags` logic from modifier and `dfg_modify` modules

nicolaassolini-qntm:
After https://github.com/Quantinuum/tket2/pull/1784 Unitary flag metadata are no more used by the pass

BREAKING CHANGE: remove the public structure `ModifierFlags` from `tket/src/modifier.rs`


aborgna-q (approved):

aborgna-q:
I just realised the PR was marked as breaking.
There are no breaking changes in the code (`ModifierFlags` was private), so we'll need to override the breaking flag manually in the release PR.

PR #1810, opened 2026-07-07
feat: add pg-tk crate

yao-cqc:
This crate enables conversion between `PauliGraph`s and TKET serialized circuits. This also allows unitary testing for Pauli graphs with only quantum ops.


yao-cqc:
Tableau conversion is not correct, fixing it now

yao-cqc:
> Tableau conversion is not correct, fixing it now

fixed

pg-libs/pg-converters/pg-tk/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "pg-tk"
+version = "0.1.0"
+edition = { workspace = true }
```
cqc-alec, line 4: Need to add `pg-converters/pg-tk` to the `workspace.members` array in `pg-libs/Cargo.toml`.

pg-libs/pg-converters/pg-tk/src/converters.rs
```diff
@@ -0,0 +1,1167 @@
...
+                    optype
+                )));
+            }
+        }
+    }
+    Ok(pg)
+}
+
+
+
```
cqc-alec:
```suggestion

```

pg-libs/pg-converters/pg-tk/src/unitary_comparison.rs
```diff
@@ -0,0 +1,234 @@
...
+            trace.as_str().unwrap_or("<non-string error>")
+        );
+    }
+
+    result["result"]
+        .as_bool()
+        .expect("Unexpected result format from python unitary comparison")
+}
+// add test module and tests
+#[cfg(test)]
```
cqc-alec, line 183: These tests will all fail on CI (once `pg-tk` is added to the workspace) because of the pytket dependency. Probably need a special step in the workflow to install pytket before running the tests.

cqc-alec (changes requested):

pg-libs/pg-converters/pg-tk/src/unitary_comparison.rs
yao-cqc: I’ve added a `pyproject.toml` inside the `pg-libs` folder. It includes the dev dependencies, pytket and ruff. The `members` list is empty for now, and I’ll add the members later.

I’ve also updated the commit hooks so they run inside the pg-libs folder and use the Python binary there.

pg-libs/pg-converters/pg-tk/Cargo.toml
yao-cqc: done

cqc-alec (approved):
Thanks!

PR #1813, opened 2026-07-07
fix: NormalizeCFGs: no need to add Order edges for new Ext edges

acl-cqc:
NormalizeCFGs can move the entry block outside the CFG, turning `Dom` edges from nodes in the entry block (to other basic blocks in the CFG) into `Ext` edges from outside. It used to add Order edges for those Ext edges, but this has been unnecessary since Quantinuum/hugr#2951 (hugr-v0.28.0).

So, don't add them...


aborgna-q (approved):
-.-

PR #1816, opened 2026-07-09
feat: Decode global phases from pytket circuits

### LINKED ISSUE #2001: [Bug]: Control modifier does not work correctly on imported pytket circuit
If we use `guppy.load_pytket()` to import a pytket circuit as a guppy function and then apply `control` to it, the result will be incorrect if the pytket circuit contains any global phase. This is because we do not take account of global phase when importing from pytket. The tket2 global-phase extension could be used to capture phase.

Note that pytket circuits may express global phase either by the `OpType.Phase` zero-qubit phase operation, or by the `phase` property on the circuit (or both).

aborgna-q:
Fixes https://github.com/Quantinuum/guppylang/issues/2001.

Now pytket circuit phase parameters and `Phase` commands are decoded as `tket.global_phase` ops.

Also added an encoder from `tket.global_phase` ops into pytket phase commands. This is not a clean roundtrip if the original circuit had a phase parameter, but it should be semantically equivalent.

The `PytketPhaseExpr` metadata is now deprecated, since we no longer encode the pytket phase there.
We still read its value when producing pytket circuits, for backwards compat with older hugrs.

### DIFF
--- a/tket/src/serialize/pytket/decoder/param/parser.rs
+++ b/tket/src/serialize/pytket/decoder/param/parser.rs
@@ -18,8 +18,6 @@ use crate::serialize::pytket::decoder::ParameterType;
 ///
 /// The leafs of the AST are either a constant value, a variable name, or an
 /// unrecognized sympy expression.
-///
-/// Return type of [`parse_pytket_param`].
 #[derive(Debug, Display, Clone, PartialEq)]
 pub enum PytketParam<'a> {
     /// A constant value that can be loaded directly.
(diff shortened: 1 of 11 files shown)


tket/src/serialize/pytket/encoder.rs
```diff
@@ -170,7 +170,9 @@ impl<H: HugrView> PytketEncoderContext<H> {
         .filter(|name| !name.is_empty())
         .cloned();
 
-        // Recover other parameters stored in the metadata
+        // Backwards compatibility only: The global phase parameter
+        // of pytket circuits is now encoded as  explicit `tket.global_phase` operations instead.
```
cqc-alec:
```suggestion
        // of pytket circuits is now encoded as explicit `tket.global_phase` operations instead.
```

tket/src/serialize/pytket/decoder/param/parser.rs
```diff
@@ -47,18 +46,64 @@ pub enum PytketParam<'a> {
...
+    pub fn is_zero(&self) -> bool {
+        matches!(self, PytketParam::Constant(value) if *value == 0.0)
+    }
+}
+
+impl<'a> PartialEq for PytketParam<'a> {
+    fn eq(&self, other: &Self) -> bool {
+        match (self, other) {
+            (Self::Constant(l0), Self::Constant(r0)) => {
+                l0.relative_eq(r0, f64::EPSILON, f64::EPSILON)
```
cqc-alec, line 74: I'm a bit confused about what this is, the docs only say "Approximate equality using both the absolute difference and relative based comparisons.". What does this mean mathematically?

cqc-alec (approved):

tket/src/serialize/pytket/decoder/param/parser.rs
aborgna-q: It's checking
$(|a - b| < \mathtt{eps}_{abs}) \land (|a - b| < \max(a, b) \times \mathtt{eps}_r )$

Here I set both the absolute and relative epsilons to `f64::EPSILON ~= 1e-16`. Should we change the relative one to something bigger?

cqc-alec: Thanks. seems OK to me.

PR #1817, opened 2026-07-09
fix: ModifierPasses producing invalid hugr with non quantum DAG

### LINKED ISSUE #1814: [Bug]: Empty modifier block produces invalid hugrs after cfg simplification
### Description

After CFG simplification, CFG blocks are converted into DFGs. For nested modifiers, if the body of a block is empty (or contains only classical computation), the corresponding DFG is skipped because it is treated as a classical function. As a result, the inner modifier blocks are never resolved, which leads to invalid HUGRs after the modifier cleanup pass.

### Proposed Solution

There are two possible approaches:

* Update the worklist logic in `modifier_resolver.rs` so that nested DFGs are still visited, even if the enclosing DFG itself does not require modification:
  https://github.com/Quantinuum/tket2/blob/f52b89af936e13dd57de4f0d8a70644d266b8ca7/tket/src/modifier/modifier_resolver.rs#L1629

* When DFG modification is skipped in `dfg_modify.rs`, still run the modifier resolution passes on the DFG body:
  https://github.com/Quantinuum/tket2/blob/f52b89af936e13dd57de4f0d8a70644d266b8ca7/tket/src/modifier/modifier_resolver/dfg_modify.rs#L518

**Note:** Checking whether a DFG contains modifiers is not sufficient. A DFG may represent an otherwise classical computation while still containing modifiers associated with qubit allocation, measurement, or other operations that are unrelated to the enclosing modifier. Each DFG should therefore be treated as an independent unit, with no assumptions about its relationship to the surrounding context, and processed in the same way as an external classical function.

### Steps to Reproduce

run 
```python
@guppy
def bar(q: qubit, q2: qubit) -> None:
    with control(q2):
        with dagger:
            pass

p = bar.compile_function()
```
with NormalizeGuppy flags on

nicolaassolini-qntm:
closes #1814


tket/src/modifier/modifier_resolver.rs
```diff
@@ -2428,7 +2468,7 @@ mod tests {
     fn test_resolve(h: &mut Hugr) {
         assert_matches!(h.validate(), Ok(()));
 
-        let entrypoint = h.entrypoint();
+        let entrypoint = PassScope::default().root(&h).unwrap_or(h.entrypoint());
```
nicolaassolini-qntm: I did this change since having a different scope between the test and the default of the passes was causing incosistency

aborgna-q (dismissed):
LGTM

aborgna-q (approved):

PR #1824, opened 2026-07-10
refactor: rename `NormalizeGuppy` to `Normalize`

### LINKED ISSUE #1780: Rename `NormalizeGuppy` to `Normalize`

nicolaassolini-qntm:
closes #1780

BEGIN_COMMIT_OVERRIDE
feat: rename NormalizeGuppy to Normalize (#1824)
feat: Disable pytket encoding for tuple wires (#1846)
END_COMMIT_OVERRIDE


tket/src/passes.rs
```diff
@@ -34,6 +34,8 @@ pub mod force_order;
 
 // Normalize the structure of Guppy-generated programs.
```
cqc-alec:
```suggestion
// Normalize the structure of programs.
```

tket-py/tket/passes/__init__.py
cqc-alec:
```suggestion
    """Flatten the structure of a program to enable additional optimisations.
```

tket/src/passes/normalize.rs
cqc-alec, line 1: Can we rename this file to `normalize.rs`?

cqc-alec (changes requested):

tket/src/passes/normalize.rs
```diff
@@ -214,6 +214,14 @@ pub enum NormalizeGuppyErrors {
     RedundantOrderEdges(HugrError),
 }
 
+/// Deprecated alias for [`Normalize`].
+#[deprecated(since = "0.22.0", note = "Use Normalize instead.")]
+pub type NormalizeGuppy = Normalize;
```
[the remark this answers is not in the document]
aborgna-q: It should be "since 0.21.2"

tket/src/passes/normalize.rs
```diff
@@ -214,6 +214,14 @@ pub enum NormalizeGuppyErrors {
     RedundantOrderEdges(HugrError),
 }
 
+/// Deprecated alias for [`Normalize`].
+#[deprecated(since = "0.21.2", note = "Use Normalize instead.")]
```
cqc-alec: Since this is now technically breaking at the rust level, we probably have to wait for 0.22.0. (Ditto below.)

tket/src/passes.rs
```diff
@@ -32,9 +32,11 @@ pub use dead_funcs::{RemoveDeadFuncsError, RemoveDeadFuncsPass};
 // Force a topological order on nodes.
 pub mod force_order;
 
-// Normalize the structure of Guppy-generated programs.
-pub mod guppy;
-pub use guppy::NormalizeGuppy;
+// Normalize the structure of programs.
+pub mod normalize;
```
aborgna-q, line 37: This is what's failing the semver checks.

You should leave a stub `guppy.rs` module and move the alias definitions there.

tket/src/passes/normalize.rs
cqc-alec: Actually Agustin's suggestion is better.

cqc-alec (approved):

PR #1827, opened 2026-07-10
refactor: Higher-order function handling in `ModifierResolverPass`

### LINKED ISSUE #1761: [Refactor]: ModifierResolverPass to deal with new guppy `Unitary` compilation
### Problem Statement

https://github.com/Quantinuum/guppylang/pull/1951 refactor the way higher-order unitary functions are handled. Now the indirect call nodes do not depend on inputs but point directly to the function definition.

### Proposed Solution

Clean up the ModifierResolverPass to delete and refactor the part that was supposed to handle input higher-order functions dynamically. This involves almost exclusively reverting the changes made by https://github.com/Quantinuum/tket2/pull/1634 and ensuring that indirect calls are always handled correctly. 

### Additional Context

While refactoring, we should also solve the bug present here:  https://github.com/Quantinuum/tket2/issues/1710 and ~the one that happened while compiling~ (solved!):

```python
@guppy(unitary=True)
    def apply_unitary(f: Unitary[[qubit], None], ctrl: qubit, q: qubit) -> None:
        with dagger:
            with control(ctrl):
                f(q)

  @guppy(unitary=True)
  def foo(q: qubit) -> None:
      pass

  @guppy
  def main(q1: qubit, q2: qubit) -> None:
      apply_unitary(h, q1, q2)
      apply_unitary(foo, q1, q2)

main.compile_function()
```

nicolaassolini-qntm:
closes #1761

After https://github.com/Quantinuum/guppylang/pull/1951 ([8cbe5fd](https://github.com/Quantinuum/guppylang/commit/8cbe5fd3673e1605b599994a5341aadf3bbba816)), all the modifiable quantum functions are statically known; thus, no special handling of the `IndirectCall and `Input` nodes is needed.


tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -255,21 +163,13 @@ impl<N: HugrNode> ModifierResolver<N> {
                         fn_optype.clone(),
                     ));
                 };
-                // TODO: We want some machinery to prevent generating a lot of copies of modified functions
```
nicolaassolini-qntm, line 258: I removed this comment since it has been handled by #1634

test_files/modifier_examples/cfg_in_ctrl.py
```diff
@@ -1,8 +1,10 @@
 # /// script
 # requires-python = ">=3.13"
 # dependencies = [
-#    "guppylang==1.0.0a8",
+#    "guppylang",
 # ]
+# [tool.uv.sources]
+# guppylang = {git = "https://github.com/quantinuum/guppylang", subdirectory = "guppylang", rev = "cc0977af918569ef9747bb07ca17fdfcbe8376e9"}
```
cqc-alec: I guess we can't use an rc release yet? (Same comment in other test files.)

cqc-alec (commented):
LGTM. Some of the comments need fixing up as pointed out by AI.

test_files/modifier_examples/cfg_in_ctrl.py
nicolaassolini-qntm: we can I have update the files!

cqc-alec (approved):

PR #1834, opened 2026-07-14
fix: Unpack values from inside tuples when decoding

aborgna-q:
When decoding a pytket, we try to match requested register input values to an existing hugr wire.

Sometimes those register values are part of a tuple in the original hugr, so we need to add an `UnpackTuple` operation before returning the matched wire.

This PR adds that automatic handling.

Required for https://github.com/Quantinuum/guppylang/issues/2041


aborgna-q:
Closing in favour of #1846

PR #1838, opened 2026-07-14
refactor: Remove leftover todos in `ModifierResolverPass`

### LINKED ISSUE #1551: [Refactor]: Cleanup the modifier pass code
### Problem Statement

There's a bunch of leftover TODOs that require context about the problem to understand.
We should make a cleanup pass over the code before enabling it as a production-ready hugr pass.

Check the modifier code in guppy too, as it has some TODOs over there.

nicolaassolini-qntm:
closes https://github.com/Quantinuum/tket2/issues/1551

This PR removes leftover TODOS. The still valid TODO has been converted into an issue.

### DIFF
--- a/tket/src/modifier/modifier_resolver.rs
+++ b/tket/src/modifier/modifier_resolver.rs
@@ -1055,6 +1055,7 @@ impl<N: HugrNode> ModifierResolver<N> {
             // Some other Hugr extension operation.
             // Here, we do not know what is the modified version.
             // We try to place the original operation.
+            // TODO: Determine when we should raise an error
             // (see https://github.com/Quantinuum/tket2/issues/1828)
             self.modify_dataflow_op(h, op_node, optype, new_dfg)
         }
@@ -1350,6 +1351,11 @@ pub fn resolve_modifier_with_entrypoints_and_scope(
     // (e.g. intermediate nodes in a chain whose last modifier was the one rewritten).
     // Walk the same reachable set again and delete any surviving modifier nodes,
     // together with every downstream node that consumes their output.
+    //
+    // NOTE:
+    // This might be insufficient as a cleanup since the resolution procedure might
+    // generate nodes that are not reachable from the entry points.
+    // If more thorough cleanup is needed, we should run dead code elimination.
     let mut deletelist = entry_points.clone();
     let mut visited = FxHashSet::default();
     while let Some(node) = deletelist.pop_front() {
--- a/tket/src/modifier/modifier_resolver.rs
+++ b/tket/src/modifier/modifier_resolver.rs
@@ -637,7 +637,7 @@ impl<N: HugrNode> ModifierResolver<N> {
         node
     }
 
-    /// connects all the wires in the builder.
+    /// Connects all the wires in the builder.
     fn connect_all(
         &mut self,
         h: &impl HugrView<Node = N>,
@@ -647,7 +647,8 @@ impl<N: HugrNode> ModifierResolver<N> {
         for out_node in h.children(parent) {
             for out_port in h.node_outputs(out_node) {
                 if let Some(EdgeKind::StateOrder) = h.get_optype(out_node).port_kind(out_port) {
-                    // TODO: Currently, we just ignore StateOrder edges.
+                    // NICOLA TODO: Open an issue, discuss with Alan
+                    // Currently, we just ignore StateOrder edges.
                     // This might be OK when the dagger is applied since StateOrder is not managable then.
                     // However, if not, we should preserve the StateOrder edges.
                     // This could be done in two ways:
@@ -670,11 +671,7 @@ impl<N: HugrNode> ModifierResolver<N> {
 }
 
 impl<N: HugrNode> ModifierResolver<N> {
-    // FIXME: Shouldn't we check that there is a caller of the modified function?
-    // We don't want to modify a function that is loaded and modified but never called.
-    // When more than one modifier is chained, after the last modifier is resolved,
-    // we delete the last modifier node, but the previous modifiers are not deleted.
-    // If the second last modifier was only called by the last modifier, that will not be called anymore.
+    // NICOLA: I would like not to do deadcode elimination here
     fn verify(&self, h: &impl HugrView<Node = N>, n: N) -> Result<(), ModifierError<N>> {
         // Check if the node is a modifier, modifying an operation.
         let optype = h.get_optype(n);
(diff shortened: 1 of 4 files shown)


tket/src/modifier/modifier_resolver.rs
```diff
@@ -664,16 +666,13 @@ impl<N: HugrNode> ModifierResolver<N> {
...
     }
 }
 
 impl<N: HugrNode> ModifierResolver<N> {
-    // FIXME: Shouldn't we check that there is a caller of the modified function?
-    // We don't want to modify a function that is loaded and modified but never called.
-    // When more than one modifier is chained, after the last modifier is resolved,
-    // we delete the last modifier node, but the previous modifiers are not deleted.
-    // If the second last modifier was only called by the last modifier, that will not be called anymore.
+    // NICOLA: I would like not to do deadcode elimination here
```
nicolaassolini-qntm: I would like not to do deadcode elimination or any kind of similar optimizaton here. Also, this may be a super edge case, not possible from guppy

tket/src/modifier/modifier_resolver/call_modify.rs
```diff
@@ -245,9 +245,9 @@ impl<N: HugrNode> ModifierResolver<N> {
         new_dfg.hugr_mut().connect(new_load, 0, new_call_node, 0);
         self.map_insert_none((n, IncomingPort::from(0)).into())?;
 
-        // FIXME: Forgetting all the nodes in the chain so that we don't have to worry about mapping the edges.
+        // Forgetting all the nodes in the chain so that we don't have to worry about mapping the edges.
         // Otherwise, there would be edges in the original graph that have no corresponding edges in the new graph.
-        // However, this could remove wires referenced by other nodes that are not in the chain.
+        // NOTE: this could remove wires referenced by other nodes that are not in the chain.
```
cqc-alec, line 250: Is this not a problem?

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -92,7 +92,8 @@ impl<N: HugrNode> ModifierResolver<N> {
                             } else if i == qubits + control {
                                 Some((halfturn, IncomingPort::from(0)).into())
                             } else {
-                                // FIXME: forget state order
+                                // TODO: forget state order
+                                // see (https://github.com/Quantinuum/tket2/issues/1828)
```
cqc-alec: This looks like the wrong github issue.

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -348,7 +349,8 @@ impl<N: HugrNode> ModifierResolver<N> {
                     mem::swap(&mut incoming, &mut outgoing)
                 }
                 incoming.push((halfturns, IncomingPort::from(0)).into());
-                // FIXME: Ignoring StateOrder
+                // TODO: Ignoring StateOrder
```
cqc-alec: "Ignoring StateOrder" reads like a normal comment rather than a TODO. What is the TODO? (Don't ignore them? Document why it's OK to ignore them?)

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -348,7 +349,8 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+                // see https://github.com/Quantinuum/tket2/issues/1828
```
cqc-alec: Wrong issue?

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -463,7 +465,8 @@ impl<N: HugrNode> ModifierResolver<N> {
                 assert_eq!(control, self.control_num());
                 self.modifiers.dagger = dagger;
 
-                // TODO: This does not handle invisible wires
+                // TODO: This does not handle StateOrder wires
```
cqc-alec: Again please rephrase as a TODO.

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -463,7 +465,8 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+                // (see https://github.com/Quantinuum/tket2/issues/1828)
```
cqc-alec: Wrong issue?

tket/src/modifier/modifier_resolver.rs
```diff
@@ -664,16 +665,12 @@ impl<N: HugrNode> ModifierResolver<N> {
...
         // FIXME: StateOrder is not preserved here.
+        // see: https://github.com/Quantinuum/tket2/issues/1836
         Ok(())
     }
 }
 
 impl<N: HugrNode> ModifierResolver<N> {
-    // FIXME: Shouldn't we check that there is a caller of the modified function?
-    // We don't want to modify a function that is loaded and modified but never called.
-    // When more than one modifier is chained, after the last modifier is resolved,
```
cqc-alec, line 674: Is this not a problem? If not, worth explaining why.

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1058,8 +1055,7 @@ impl<N: HugrNode> ModifierResolver<N> {
             // Some other Hugr extension operation.
             // Here, we do not know what is the modified version.
             // We try to place the original operation.
-            // TODO: Revisit whether unknown extension operations should return
-            // an explicit error instead of falling back to the original operation.
+            // (see https://github.com/Quantinuum/tket2/issues/1828)
```
cqc-alec, line 1059: Why delete the TODO comment? Best keep a TODO for searchability if there is something to do.

tket/src/modifier/modifier_resolver.rs
```diff
@@ -1354,10 +1350,6 @@ pub fn resolve_modifier_with_entrypoints_and_scope(
     // (e.g. intermediate nodes in a chain whose last modifier was the one rewritten).
     // Walk the same reachable set again and delete any surviving modifier nodes,
     // together with every downstream node that consumes their output.
-    // TODO:
-    // This might be insufficient as a cleanup since the resolution procedure might
```
cqc-alec, line 1358: I think it's worth keeping this comment, even if we don't consider it a TODO.

tket/src/modifier/modifier_resolver/call_modify.rs
nicolaassolini-qntm: Not really, modifier nodes should be only part of the chain, a case where a node in the chain is connected to somenthing external is not possible from guppy (and does not make any sense at the moment).

tket/src/modifier/modifier_resolver.rs
nicolaassolini-qntm: > Shouldn't we check that there is a caller of the modified function?

I think we should not do this kind of checking at this point. This can be a sort of dead code elimination, which may be unwanted.

>  we delete the last modifier node, but the previous modifiers are not deleted.

This is not true, probably the comment is old

> If the second last modifier was only called by the last modifier, that will not be called anymore.

I dont know what is the meaning of this, but chained modifiers are handle properly now

nicolaassolini-qntm:
Sorry for the wrong issue links, I messed up while copy-pasting the links

cqc-alec (approved):

PR #1845, opened 2026-07-15
feat: pg-kernels

yao-cqc:
This PR contains three crates:

1. `pg-utils`: a small utility crate that currently provides a function for evaluating Clifford angles, and a function for checking if a value is approximately 0 modulo n.
2. `pg-ir-kernels`: provides simple rewrites for operations in a `PauliGraph`. These rewrites are not optimised. For example, commutativity is checked by counting anti-commuting Pauli letters one by one. However, they are useful for prototyping and are also suitable for less performance-sensitive routines. The `PGTableau` trait defines how a Clifford tableau implementation interacts with `PauliGraph` operations.
3. `pg-bitpacked`: implements Clifford conjugation on binary-encoded Paulis.

I have also added a few workarounds to the `justfile` for a local development issue caused by having two separate Python projects share the same devenv environment. We should be able to remove these once the two projects are better integrated.


pg-libs/pg-kernels/pg-ir-kernels/src/tableau_trait.rs
```diff
@@ -0,0 +1,302 @@
...
+                }
+                vec![Op::ConditionalBox {
+                    data: ConditionalBoxData::new(
+                        new_ops,
+                        data.get_conditional_bits().clone(),
+                        data.get_conditional_values().clone(),
+                    ),
+                }]
+            }
+            _ => panic!(
```
cqc-alec: Suggest using `unreachable!` here.

pg-libs/pg-kernels/pg-ir-kernels/src/ops.rs
```diff
@@ -0,0 +1,728 @@
...
+/// # Returns
+///
+/// - `bool` - `true` if the operation is a Clifford operation, `false` otherwise.
+pub fn is_clifford(op: &Op) -> bool {
+    match op {
+        Op::Gate { data } => {
+            data.get_conditional_bits().is_empty()
+                && (is_clifford_gate_type(data)
+                    || (is_rotation_gate_type(data)
+                        && data.get_params().iter().all(|&p| cliff_angle(p).is_some())))
```
cqc-alec: This will miss some cases where `PhasedX` is Clifford. Specifically, `PhasedX(a,b)` is Clifford if `a` is a multiple of pi and `b` is a multiple of pi/4.

pg-libs/pg-kernels/pg-ir-kernels/src/ops.rs
```diff
@@ -0,0 +1,728 @@
...
+        }
+    }
+
+    fn merge_ops<T: PGTableau>(&mut self, i: usize, j: usize) {
+        if !self.can_merge(i, j) {
+            panic!("Ops that cannot be merged");
+        }
+        let op_i = &self.get_ops()[i];
+        let op_j = &self.get_ops()[j];
+        // True if both clifford
```
cqc-alec: Comment looks like a copy-paste from above.

pg-libs/pg-kernels/pg-ir-kernels/src/ops.rs
```diff
@@ -0,0 +1,728 @@
...
+                        ),
+                    }],
+                )
+            }
+            _ => panic!("Unsupported op type for merging"),
+        }
+    }
+}
+
+// Here we only test methods that do not require tableau. Remaining methods will be tested in a separate test crate.
```
cqc-alec, line 538: What is the separate test crate? Is this TODO?

pg-libs/pg-kernels/pg-bitpacked/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+/// ```
+pub fn u64s_to_sign_flips(packed_sign_flip_bits: &[u64], n_signs: usize) -> Vec<bool> {
+    let mut signs = Vec::with_capacity(n_signs);
+    let mut decoded = 0;
+    for encoding in packed_sign_flip_bits {
+        for i in 0..64 {
+            if decoded == n_signs {
+                return signs;
+            }
+            let bits = ((encoding >> i) & 0b1) as u8;
```
cqc-alec: Is the `as u8` needed?

pg-libs/pg-kernels/pg-bitpacked/src/lib.rs
```diff
@@ -0,0 +1,376 @@
...
+            let sign = match bits {
+                0 => false,
+                1 => true,
+                _ => unreachable!(),
+            };
+            signs.push(sign);
```
cqc-alec:
```suggestion
            signs.push(bits == 1);
```

pg-libs/pg-kernels/pg-bitpacked/src/slice.rs
```diff
@@ -0,0 +1,513 @@
+use super::Pauli;
+use crate::kernels::*;
+use crate::u8_encodings::{XX_U8, XY_U8, XZ_U8, YX_U8, YY_U8, YZ_U8, ZX_U8, ZY_U8, ZZ_U8};
+
+// ------------ 1Q Clifford gates that mutate the Pauli operators --------------
+
+macro_rules! make_slice_sq_gate {
+    ($slice_fn:ident, $gate_fn:ident) => {
+        #[doc = concat!(
+                                            "Scalar slice version of [`",
```
cqc-alec: Why the large indentations in these macros?

pg-libs/pg-utils/pg-utils/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+/// # Returns
+///
+/// - `Option<u8>` - The Clifford angle is expressed as a number of quarter turns (0, 1, 2, 3) if the value is a valid Clifford angle, otherwise `None`.
+///
+pub fn cliff_angle(value: f64) -> Option<u8> {
+    const EPS: f64 = 2e-9;
+    let mut x = value % 2.0;
+    if x < 0.0 {
+        x += 2.0;
+    }
```
cqc-alec:
```suggestion
    let mut x = value.rem_euclid(2.0);
```
This just does the right thing.

cqc-alec (commented):
Looks good, just a few comments.

pg-libs/pg-kernels/pg-ir-kernels/src/ops.rs
yao-cqc: The separate test crate depends on both `pg-ir-kernels` and `pg-qm-tableau`. It already exists, but can only be added once both crates are in.

pg-libs/pg-kernels/pg-bitpacked/src/slice.rs
yao-cqc: Good catch. This is a known issue with rustfmt
https://github.com/rust-lang/rustfmt/issues/5489
I've fixed the indentation, and added `#[rustfmt::skip]` for these macros.

pg-libs/pg-kernels/pg-ir-kernels/src/ops.rs
yao-cqc: Thanks! I also added the case that the gate is an identity when `a` is a multiple of 2pi

pg-libs/pg-utils/pg-utils/src/lib.rs
```diff
@@ -26,6 +25,25 @@ pub fn cliff_angle(value: f64) -> Option<u8> {
...
+/// - `n` (`f64`) - The modulus value.
+///
+/// # Returns
+///
+/// - `bool` - `true` if the value is approximately 0 modulo n, otherwise `false`.
+///
+pub fn equiv_0(value: f64, n: f64) -> bool {
+    if n == 0.0 {
+        return false;
+    }
```
cqc-alec: I think if this function is called with `n == 0.0` it's a bug, so fine for `rem_euclid()` to panic?

cqc-alec (approved):

pg-libs/pg-utils/pg-utils/src/lib.rs
yao-cqc: Good idea! Done.
`rem_euclid` does not panic on 0.0, so we need to panic explicitly.

PR #1848, opened 2026-07-15
feat: Frontend support for Helios platform configuration

cgh-qtnm:
This PR adds a new `_set_platform_config` function to tket-py under the path `tket.platform.helios`. This function sets some metadata on the root of a compiled HUGR which controls Helios-specific hardware features. The metadata object and key are defined in `tket_qsystem::extension::qsystem::helios`. Code to consume and apply the platform configuration is not included here.

As noted in the function docstring, this interface is unstable and subject to change without notice.

This is ported from the original implementation in https://github.com/Quantinuum/guppylang/pull/1807.


tket-py/tket/platform/helios.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def set_platform_config(
+    hugr: Package | Hugr[Any],
+    squash_rxys: bool = True,
+    enable_dd: bool = False,
+) -> None:
+    """EXPERIMENTAL: Set Helios-specific job configuration options on a compiled HUGR.
+    This is currently provided for development purposes only - options may not work as
+    expected (or at all) and the interface may change in future releases.
```
aborgna-q, line 26: Since this is an unstable API, should we mark it as `_hidden` for now?

cgh-qtnm:
With the addition of `leakage_repump` the set of supported config features is now complete. I think this PR can move independently of the backend support, it won't break anything if we release this first.

tket-py/test/test_helios_config.py
```diff
@@ -1,73 +1,76 @@
...
         config = module[module.module_root].metadata[HELIOS_CONFIG_META_KEY]
         assert config == {
             "squash_rxys": False,
             "enable_dd": True,
+            "leakage_repump": True,
         }
 
 
-def test_set_platform_config_independent_copies() -> None:
+def test__set_platform_config_independent_copies() -> None:
```
ss2165: beautiful find and replace

tket-py/tket/platform/helios.py
```diff
@@ -0,0 +1,50 @@
...
+
+from typing import Any
+
+from hugr.hugr.base import Hugr
+from hugr.package import Package
+
+from .._tket import metadata as _metadata
+
+#: Metadata key under which the Helios platform configuration is stored.
+HELIOS_CONFIG_META_KEY = _metadata.HELIOS_CONFIG_KEY
```
aborgna-q: This should be a `Metadata` subclass instead, plus a python enum for the options. Add the definition to `metadata.py`.

tket-qsystem/src/extension/qsystem/helios.rs
```diff
@@ -233,6 +233,23 @@ impl MakeOpDef for RuntimeBarrierDef {
     }
 }
 
+/// Platform config options exposed by `tket.platform.helios.set_platform_config`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
```
aborgna-q: - impl `Default` too?
- Should it be `non_exhaustive`?

tket-qsystem/src/extension/qsystem/helios.rs
```diff
@@ -233,6 +233,23 @@ impl MakeOpDef for RuntimeBarrierDef {
...
+pub struct HeliosPlatformConfig {
+    /// Whether to combine single-qubit gates at runtime (independent of any
+    /// compile-time squashing).
+    pub squash_rxys: bool,
+    /// Whether to enable dynamical decoupling.
+    pub enable_dd: bool,
+    /// Whether to enable leakage repump.
+    pub leakage_repump: bool,
```
aborgna-q, line 254: We probably want to `#[serde(default)]` the fields (or `#[serde(default="true")]`), to be lenient towards missing flags in user code.

tket-qsystem/src/extension/qsystem/helios.rs
```diff
@@ -651,4 +671,24 @@ mod test {
     fn test_cast() {
         crate::extension::qsystem::common::test_utils::assert_cast_roundtrip::<HeliosOp>();
     }
+
+    #[test]
+    fn roundtrip_helios_public_config() {
```
aborgna-q: Is this a unit test for serde? Seems unnecessary.

tket-py/tket/platform/helios.py
cgh-qtnm: I made it a metadata subclass, but I don't understand what you mean about the enum for the options.

aborgna-q: Sorry, "dataclass" rather than enum.

We can use a specialized value type for the python Metadata protocol instead of a generic `dict[str, bool]`.

aborgna-q (approved):

PR #1850, opened 2026-07-16
fix: Parameters used before declaration in pytket decoder

aborgna-q:
Fixes an edge-case when decoding a pytket circuit where parameters were used before being produced.

Required for https://github.com/Quantinuum/guppylang/issues/2041.

---

Opaque subgraphs from a hugr may be encoded in two ways when producing pytket circuits:
- As a pytket Barrier command, if they contained register input/outputs (so their order is well preserved).
- As an external hugr subgraph in the rust context.

While pytket sets a strict order between commands, we have no way to impose an order between them and the subgraphs in the hugr.

This is not normally a problem, except that different subgraphs may be connected via parameter wires that can be produced and consumed in both kinds of subgraphs.

Since we process all the external subgraphs at the beginning of the decoding process, if we find a parameter input we should register it somewhere such that if a pytket barrier declares it as an output we are able to connect the ports correctly.

This PR does so by adding an internal `WireTracker::reserve_forward_parameter` call.

### DIFF
--- a/tket-py/tket/_rewrite.py
+++ b/tket-py/tket/_rewrite.py
@@ -3,7 +3,7 @@ from ._tket.rewrite import ECCRewriter, CircuitRewrite, Subcircuit
 
 __all__ = [
     "default_ecc_rewriter",
-    # Bindings.
+    # Bindings
     # TODO: Wrap these in Python classes.
     "ECCRewriter",
     "CircuitRewrite",
--- a/tket/src/serialize/pytket/decoder/wires.rs
+++ b/tket/src/serialize/pytket/decoder/wires.rs
@@ -3,13 +3,13 @@
 use std::collections::{BTreeMap, VecDeque};
 use std::sync::Arc;
 
-use hugr::builder::{DFGBuilder, Dataflow as _};
+use hugr::builder::{Container as _, DFGBuilder, Dataflow as _};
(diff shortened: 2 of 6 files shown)


tket/src/serialize/pytket/decoder/wires.rs
```diff
@@ -1036,24 +1042,93 @@ impl WireTracker {
...
+    /// its consumers to the produced value, and removes the temporary node.
+    ///
+    /// Otherwise, this creates a new binding and records the name in the
+    /// context. In that case the `builder` is not modified.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`PytketDecodeErrorInner::DuplicatedParameter`] if the name is
+    /// already bound and is not an unresolved forward reference.
+    pub(super) fn bind_parameter(
```
aborgna-q, line 1058: Renamed this internal method, I think `bind_parameter` is clearer than `register_input_parameter`.

cqc-alec (approved):

tket-py/tket/_rewrite.py
```diff
@@ -3,7 +3,7 @@
 
 __all__ = [
     "default_ecc_rewriter",
-    # Bindings.
+    # Bindings
```
aborgna-q, line 6: Added a dummy change in `tket-py` so this triggers a changelog line / release pr.

PR #1864, opened 2026-07-20
feat: Prepare use of QIS compiler as a Rust crate

maximilianruesch:


qis-compiler/Cargo.toml
```diff
@@ -30,7 +31,8 @@ tket-qsystem = { path = "../tket-qsystem", version = "0.27.1", features = [
 workspace = true
 
 [features]
-default = []
+default = ["py"]
```
aborgna-q: Given how the `pyo3` build is only required when building with maturin, I think we can make this non-default.

(It's not a breaking change if it was never published as a crate).

aborgna-q (approved):

PR #1870, opened 2026-07-22
feat: Add crate pg-qm-tableau for qubit major Clifford tableaus

yao-cqc:
This PR adds a qubit-major Clifford tableau implementation to `pg-libs`.

# Basic APIs

The tableau provides standard APIs for post-composition, pre-composition, Pauli string conjugation, and inversion.

# `PGTableau` trait

The tableau implements the `PGTableau` trait, allowing it to be used as a backend for `PauliGraph` rewrites. However, not all methods are implemented efficiently, which is acceptable for now since they are not currently used in performance-critical applications. Additionally, the `PGTableau` implementation is not SIMD-enabled.

# `SIMD`

Enabling the `simd` feature provides the `SimdTableau` trait, which exposes SIMD-accelerated implementations of selected tableau methods.

# `pg-ir-kernels-tests` crate

This PR also enables testing for the previously merged `pg-ir-kernels` crate https://github.com/Quantinuum/tket2/pull/1845. I have included it here because it provides examples of how this tableau can be used for rewrites.


pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+    }
+
+    /// Postcompose a single-qubit basis change gate.
+    /// The basis change is specified by a Pauli operator.
+    ///
+    /// # Arguments
+    ///
+    /// * `axis` - The Pauli axis for the basis change (X, Y, or Z)
+    /// * `q` - Index of the qubit to apply the basis change to
+    /// * `dagger` - Whether to apply the inverse (daggered) version of the gate
```
willsimmons1465: Seeing this method in use feels weird because I'd expect `axis=Pauli::Y` to work algebraically similar to `Pauli::X` or `Pauli::Z` in implementing Ry(pi/2), but it subtly sneaks in an extra Pauli to get to a H. I get it from the design of trying to collect the implementations for the gates we have shorthand Ops for, but that is not obvious from the method signature and docstring. Could you make this a bit clearer in the docstring?

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+                        data.get_args()[1],
+                    ),
+                    GateType::SWAP => self.postcompose_swap(data.get_args()[0], data.get_args()[1]),
+                    GateType::RX => {
+                        // TODO not efficient
+                        let mut string = vec![Pauli::I; self.get_n_qubits()];
+                        string[data.get_args()[0]] = Pauli::X;
+                        self.postcompose_op(&Op::Rotation {
+                            data: (RotationData::new(string, data.get_params()[0])),
+                        })
```
willsimmons1465: If you're looking for a more efficient implementation, match the half-pis mod 4 and defer to V, X, Vdg, or nothing.
This might get weirder for RY since RY(pi/2) is not what is implemented by `postcompost_basis_change(Pauli::Y, _)`. More motivation for handling Hadamards separately from pi/2 rotations

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+                    GateType::PHASEDX => {
+                        let alpha = data.get_params()[0];
+                        let beta = data.get_params()[1];
+                        // handle special cases
+                        // if alpha is a multiple of 2pi, then this is identity
+                        if equiv_0(alpha, 2.0) {
+                            return;
+                        }
+                        // if alpha is a multiple of pi and beta is a multiple of pi/4, then this is a Clifford gate
+                        if equiv_0(alpha, 1.0) && equiv_0(beta, 0.25) {
```
willsimmons1465: Since we have already checked `is_clifford(op)`, once we have `equiv_0(alpha, 1.0)` we are guaranteed to have `equiv_0(beta, 0.25)` (since this also holds for all Clifford angles

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+                                .with_params(vec![-beta]),
+                        });
+                        self.postcompose_op(&Op::Gate {
+                            data: GateData::new(GateType::RX, data.get_args().clone())
+                                .with_params(vec![alpha]),
+                        });
+                        self.postcompose_op(&Op::Gate {
+                            data: GateData::new(GateType::RZ, data.get_args().clone())
+                                .with_params(vec![beta]),
+                        });
```
willsimmons1465: At this point, by `is_clifford(op)` and the special cases above, we know `alpha` is an odd number of half-pis, and `beta` is some number of half-pis. Similar to the RX/RY/RZ case, this might then be more efficiently handled by a LUT to select between V/Vdg/RY(pi/2)/RY(-pi/2)

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+impl PGTableau for Tableau {
+    fn eye(n_qubits: usize) -> Self {
+        Tableau::eye(n_qubits)
+    }
+    fn get_n_qubits(&self) -> usize {
+        self.get_n_qubits()
+    }
+    fn x(&self, qubit: usize) -> (Vec<Pauli>, bool) {
+        let (paulis, sign) = self.get_x_col_paulis(qubit);
+        (paulis, !sign)
```
willsimmons1465: I would argue that the `PGTableau` trait should stick with standard conventions for symplectic tableau representations of 0 for +1 and 1 for -1 (matching the relationship between measurement outcomes and eigenvalues). In that case, the negation here would not be needed

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+                _ => Pauli::Z,
+            };
+            tableau.postcompose_tqe(g0, g1, q0, q1);
+        }
+        tableau
+    }
+
+    /// get a column by index
+    // TODO:  consider reusable scratch buffers for columns to avoid repeated allocations
+    fn get_col(&self, col_index: usize) -> (Vec<u64>, Vec<u64>, bool) {
```
willsimmons1465: As in our previous discussions, the attribution of row/column to image/preimage of a Pauli in a unitary tableau is not always consistent across our works, so it would help to avoid them even as names in code where possible in favour of semantically-relevant terms like image/preimage for unitary tableaus or string/qubit for state tableaus. Likely to cause some developer confusion if others have to debug this later on (i.e. I would personally get confused and be constantly fighting my preconceived notions), but change is very optional here as it only occurs in a class internals and not in the API so not too critical

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+    fn postcompose_tableau(&mut self, other: &Self) {
+        self.compose(other);
+    }
+
+    fn conjugate_string(&self, paulis: &[Pauli]) -> (Vec<Pauli>, bool) {
+        let (z_vec, x_vec) = paulis_to_u64s(paulis);
+        let (new_z_vec, new_x_vec, neg_sign) = self.apply_to_pauli(&z_vec, &x_vec);
+        (
+            u64s_to_paulis(&new_z_vec, &new_x_vec, self.get_n_qubits()),
+            !neg_sign,
```
willsimmons1465: Same preference on sign convention here

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+                row[2 * col_idx + 1] = data.get_x_outputs()[col_idx].0[row_idx];
+            }
+            let (zb_row, xb_row) = paulis_to_u64s(&row);
+            zb_rows.push(zb_row);
+            xb_rows.push(xb_row);
+        }
+        let mut neg_signs = vec![false; 2 * n_qubits];
+        for col_idx in 0..n_qubits {
+            neg_signs[2 * col_idx] = !data.get_z_outputs()[col_idx].1;
+            neg_signs[2 * col_idx + 1] = !data.get_x_outputs()[col_idx].1;
```
willsimmons1465: The API for `TableauData` should probably have a comment that specifies its sign convention

pg-libs/pg-qm-tableau/src/converter.rs
```diff
@@ -0,0 +1,660 @@
...
+
+    #[rstest]
+    #[case(0.0, 0.0)]
+    #[case(2.0, 1.7)]
+    #[case(0.0, 1.7)]
+    #[case(-2.0, 1.7)]
+    #[case(-1.0, 0.5)]
+    #[case(3.0, 0.25)]
+    #[case(1.0, -0.25)]
+    fn test_postcompose_phasedx(#[case] alpha: f64, #[case] beta: f64) {
```
willsimmons1465, line 492: Please add a case with alpha being a non-Pauli angle, i.e. .5 mod 1

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+        z_res.copy_to_slice(z_chunk0);
+        x_res.copy_to_slice(x_chunk0);
+        let z0x1 = z0 & x1;
+        let x0z1 = z1 & x0;
+        let anti_commute = z0x1 ^ x0z1;
+        let anti_cyclic = (x0z1 & !z0 & !x1) | (z0x1 & (x0 ^ z1));
+        n_anti_commute_sites += anti_commute.count_ones().reduce_sum();
+        n_anti_cyclic_sites += anti_cyclic.count_ones().reduce_sum();
+    }
+    // TODO: Use SIMD for the remaining elements.
```
willsimmons1465, line 154: I think it's fine to commit without SIMD here for now. If we are serious about fixing this, I think it should probably become a GH issue rather than a TODO in code

There are a few other TODOs in this file as well about SIMD/parallelisation

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+                    x1_col_index,
+                    &zb_x1_col,
+                    &xb_x1_col,
+                    s_x0_col ^ s_x1_col ^ (phase == 2),
+                );
+            }
+            (Pauli::Y, Pauli::Y) => {
+                // z0 <- z0*y1 = iz0*x1*z1
+                let mut zb_z0_col_copy = zb_z0_col.clone();
+                let mut xb_z0_col_copy = xb_z0_col.clone();
```
willsimmons1465: Copies are not needed

y0' = i(x0 * y1) * (z0 * y1) = i x0 * z0 * y1 * y1 = i x0 * z0 = y0

If anything, storing y1 to reuse it between the z0 and x0 updates cuts down one string multiplication, and similarly y0 to reuse it between the z1 and x1 updates for another one.

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+                0 => Pauli::X,
+                1 => Pauli::Y,
+                _ => Pauli::Z,
+            };
+            let g1 = match rng.random_range(0..3) {
+                0 => Pauli::X,
+                1 => Pauli::Y,
+                _ => Pauli::Z,
+            };
+            tableau.postcompose_tqe(g0, g1, i, i + 1);
```
willsimmons1465: Why does the ladder use pairs of TQE gates here rather than just one TQE per step?

pg-libs/pg-qm-tableau/src/lib.rs
willsimmons1465: Or, given that `postcompose_half_pis` exists, I would prefer using that and providing an explicit `postcompose_h` which handles it separately

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+        self.postcompose_pauli_with(pauli, q, slice_x_gate, slice_y_gate, slice_z_gate);
+    }
+
+    /// Return image of X on the ith qubit. Negative sign indicates a phase of +1.
+    pub fn get_x_col_paulis(&self, i: usize) -> (Vec<Pauli>, bool) {
+        let res = self.get_col(2 * i + 1);
+        (u64s_to_paulis(&res.0, &res.1, self.n_qubits), res.2)
+    }
+
+    /// Return image of Z on the ith qubit. Negative sign indicates a phase of +1.
```
willsimmons1465: Similar comment on sign conventions. Especially here, "negative sign" meaning +1 is very confusing

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+    }
+
+    /// Multi-threaded version of apply_to_pauli_with
+    /// the conjugation of z-bits and x-bits, as well as the counting of ys, are done in parallel
+    fn apply_to_pauli_mt_with(
+        &self,
+        zb_pauli: &[u64],
+        xb_pauli: &[u64],
+        string_mul_cols: StringMulFn,
+        string_mul_merge: StringMulFn,
```
willsimmons1465: Both uses of this give identical methods for `string_mul_cols` and `string_mul_merge`. Are you anticipating a situation where we might need them to be separate or could we drop it down to just one?

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -0,0 +1,2020 @@
...
+    /// Return image of Z on the ith qubit. Negative sign indicates a phase of +1.
+    pub fn get_z_col_paulis(&self, i: usize) -> (Vec<Pauli>, bool) {
+        let res = self.get_col(2 * i);
+        (u64s_to_paulis(&res.0, &res.1, self.n_qubits), res.2)
+    }
+}
+
+/// Trait for SIMD operations on Tableau
+#[cfg(feature = "simd")]
+pub trait SimdTableau {
```
willsimmons1465, line 1659: I don't think I get why the SIMD versions need to have a separate interface. Would it not be easier for them to have the same names in the API as non-SIMD variants so we can swap between non-SIMD and SIMD purely with compiler flags and not relying on the programmer changing every call in their code to add/remove `_simd`?

willsimmons1465 (changes requested):
Nice work! Mostly seems pretty much as expected for implementing a tableau data structure. Most of my comments are the usual discussion on conflicting naming/notational conventions and finding ways to phrase the code and API to avoid these becoming issues in the future - for such comments, if you wish to ignore them because this is an internal module that we don't expect users to interact with and should later be replaced by a zixy integration, then feel free to just mark them resolved as appropriate

pg-libs/pg-qm-tableau/src/lib.rs
yao-cqc: I agree that this would provide a cleaner API. However, the current API was chosen mainly to make benchmarking and comparison easier, as the SIMD feature is still very experimental. I plan to do another round of benchmarking and update the library to the latest nightly toolchain. It might be better to revisit this issue once that investigation is complete.

pg-libs/pg-qm-tableau/src/lib.rs
yao-cqc: As mentioned in another thread, the SIMD feature is experimental and needs a broader investigation. These TODOs are therefore not currently planned, and I would prefer to leave them here. There are also some TODOs that are not related to SIMD or parallelisation. I honestly think they are pretty low priority, and we might never need to work on them.

yao-cqc (commented):
Hi Will, thanks for the review. These are genuinely great comments! I’ve tried to address all of them, but please let me know if I’ve missed anything. For reviewing, I think the best approach is to check the commits individually, as a large chunk of the diff is just renaming stuff. The project now uses `image`/`qubit_slice` instead of `column`/`row` when addressing a tableau. Signs are also named more consistently as `sign_bit` and follow the convention that `true` means −1 and `false` means +1.

pg-libs/pg-qm-tableau/src/lib.rs
```diff
@@ -413,70 +413,56 @@ impl Tableau {
...
+                    + x0_phase_new
+                    + 1;
+                // z1 <- z1*y0
+                let z1_phase_new =
+                    (string_mul(&mut zb_z1_col, &mut xb_z1_col, &zb_y0_col, &xb_y0_col)
+                        + z1_phase
+                        + y0_phase)
+                        % 4;
+                self.set_col(z1_col_index, &zb_z1_col, &xb_z1_col, z1_phase_new == 2);
+                // x1 <- x1*y0 = (-iy1*z1)*y0 = -iy1* (z1*y0)
```
willsimmons1465, line 456: Just to check my understanding here, is the reason why we can't just do `x1 <- x1*y0` directly because `let mut <P>b_y1_col = <P>b_x1_col;` doesn't copy the data of `x1` but instead moves/renames it, and therefore `<P>b_x1_col` is out of scope at this point?

willsimmons1465 (approved):
Excellent, thanks for the clean split of changes across commits as well. Looks good to go!

maximilianruesch:
I will take a look at this PR soon!

maximilianruesch (approved):
The change in `.github` is fine for me. The changes in `pg-libs` have been reviewed by @Quantinuum/pg-libs, and are specifically due for refactoring and reassembly once integration is done 🫡

pg-libs/pg-qm-tableau/src/lib.rs
yao-cqc: exactly, it’s just to avoid cloning x1.

PR #1872, opened 2026-07-22
docs: Show documentation for `tket.extensions` module

### LINKED ISSUE #1844: Generated docs for tket extensions are empty
https://quantinuum.github.io/tket2/generated/tket.extensions.html#

cqc-alec:
Fixes #1844 .

I could not find any way to make sphinx's `autosummary` include the module-level constants (`argument` etc) imported from `tket_exts`, so this is a compromise soltution:
- Export the class definitions from `tket_exts` (`ArgumentExtension` etc) so that these and their methods are documented.
- Extend the module documentation so that it explains how the extensions can be imported directly.

The main `extensions` page now looks like this:

image

And if you click on `ArgumentExtension` (for example) you see this:

image

### DIFF
--- a/tket-py/tket/extensions/__init__.py
+++ b/tket-py/tket/extensions/__init__.py
@@ -1,3 +1,5 @@
+"""HUGR extension definitions for tket circuits."""
+
 from tket_exts import (
     argument,
     debug,
--- a/tket-py/tket/extensions/__init__.py
+++ b/tket-py/tket/extensions/__init__.py
@@ -21,6 +21,13 @@ Extensions may be imported directly as follows:
 >>> result_extension = ext.result()
 >>> rotation_extension = ext.rotation()
 >>> wasm_extension = ext.wasm()
+>>>
+>>> read_arg_op = argument_extension.get_op("read_arg")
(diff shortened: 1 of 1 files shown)


tket-py/tket/extensions/__init__.py
```diff
@@ -1,3 +1,28 @@
...
+>>> modifier_extension = ext.modifier()
+>>> qsystem_extension = ext.qsystem()
+>>> qsystem_helios_extension = ext.qsystem_helios()
+>>> qsystem_sol_extension = ext.qsystem_sol()
+>>> qsystem_random_extension = ext.qsystem_random()
+>>> qsystem_utils_extension = ext.qsystem_utils()
+>>> quantum_extension = ext.quantum()
+>>> result_extension = ext.result()
+>>> rotation_extension = ext.rotation()
+>>> wasm_extension = ext.wasm()
```
maximilianruesch, line 23: If you make the members show up as in the general comment, this would do with one example (which might be a bit more elaborate, showing what you can do with the object you receive from the call).

maximilianruesch (changes requested):
The reason you were not able to make the constants show up is that Sphinx treats them as general `members` rather than `functions` (since they are technically class instances with a `__call__` dunder method), and those are not listed in the template at https://github.com/Quantinuum/tket2/blob/main/tket-py/docs/_templates/autosummary/module.rst.

You could consider a custom template for this module that lists the members (I do not think they should be included in general, since they contain duplicates of the entries in `functions` etc.), which would result in
image

But I also understand if you want to stick to the current solution (which is arguably higher maintenance though.

maximilianruesch (approved):

PR #1886, opened 2026-07-28
chore: new tket-qsim unitary simulator for unit testing

### LINKED ISSUE #1648: [Test]: Add a mini-executor for validating rewrite tests
### Problem Statement

We cannot use selene in the local tests to validate that rewrite passes do not change the semantics.
This means that most pass tests are not testing for behaviour changes.

We should implement a minimal executor/interpreter that supports small <10 qubit states and interprets only the operations that we need to test.

This should be only used for internal testing, and should not be published as a crate.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

ferbetanzo:
Closes #1648. Includes the following:

- New `tket-qsim` crate that allows to simulate (purely  unitary) hugrs. It defines the required `<tket|helios|sol>_ops` unitary matrices but compares them against `roqoqo`'s gate definitions.
- New semantics preserving tests for cross-compilation.


tket-qsystem/src/extension/qsystem/lower.rs
```diff
@@ -1812,4 +1812,209 @@ mod test {
...
+        InlineFunctionsPass::default()
+            .with_heuristic(InlineFuncsHeuristic::All)
+            .run(&mut h)
+            .unwrap();
+        InlineDFGsPass::default().run(&mut h).unwrap();
+        h.validate().unwrap();
+
+        let unitary_after = simulate_circuit(&h, &params).unwrap();
+        assert!(
+            unitary_before.approx_eq_up_to_global_phase(&unitary_after, 1e-10),
```
doug-q, line 1969: I would prefer to use the https://docs.rs/approx/latest/approx/ crate for this

tket-qsim/src/matrix.rs
```diff
@@ -0,0 +1,321 @@
+use num_complex::Complex64;
+use std::fmt;
+
+/// A dense unitary matrix of dimension `2^num_qubits × 2^num_qubits`.
+///
+/// Stored in row-major order.
+#[derive(Clone)]
+pub struct UnitaryMatrix {
```
doug-q, line 8: I would prefer to use the https://www.nalgebra.rs/docs/ crate for this

doug-q (commented):
Thanks for this. Apologies but I do not think we should proceed with this approach. Happy to discuss.

I think that we can't use selene because of dependency loops, is that right?

Problems with this approach  
* It relys on the hugrs under test to get optimised into single-dfg functions
* lowerings that contain control flow cannot be handled
* lowerings that include ancillas or mid circuit measurements cannot be handled
* The hugrs that are tested are fixed. This approach is not going to catch corner cases (e.g. bugs where special angles cause big problems)

I don't think this is the right solution to verifying our transformations.

I suggest
* Each op should be verified independently
* The expected unitary should be specified local to the test
* I don't think we should build an interpreter here. (I would be ok on depending on a separate, complete(ish) interpreter)
* A better approach, IMO, would be to
  * depend on selene in such a way as it doesn't  create circular dependencies. This is surely possible because we don't need to run any tket transformations in selene (we can do it beforehand)
  * For parameterised gates use property testing to explore the parameter space and catch corner cases

PR #1888, opened 2026-07-28
fix: Swap angles in PhasedXX decomposition

### LINKED ISSUE #1887: `PhasedXX` decomposition in qsystem extension is incorrect
In the definition of `build_phasedxx` (in `tket-qsystem/src/extension/qsystem/helios.rs`) the angles `angle1` and `angle2` are the wrong way round.

cqc-alec:
Fixes #1887 .

### DIFF
--- a/tket-qsystem/src/extension/qsystem/helios.rs
+++ b/tket-qsystem/src/extension/qsystem/helios.rs
@@ -534,12 +534,12 @@ where
         let pi_minus_2 = pi_mul_f64(self, -0.5);
         let [added] = self
             .inner
-            .add_dataflow_op(FloatOps::fadd, [pi_minus_2, angle1])?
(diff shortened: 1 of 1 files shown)


ferbetanzo (approved):
LGTM. Maybe just add a small unit test to apeace codecov? It will certainly be tested once the cross-compilation semantics preserving tests are merged.

aborgna-q (approved):
Nine catch @ferbetanzo

cqc-alec:
> LGTM. Maybe just add a small unit test to apeace codecov? It will certainly be tested once the cross-compilation semantics preserving tests are merged.

At the moment we can't effectively unit-test this since it's an arbitrary definition, so I will just merge as is and wait for your semantic tests.

PR #1894, opened 2026-07-30
feat: Extract multiple pytket circuits from a hugr region

### LINKED ISSUE #1856: [Feat]: Refactor handling of opaque subgraphs in pytket decoder
### Description

This is a rewrite proposal for the logic governing how unsupported subgraphs of a hugr region are encoded in the pytket circuit, and tracked in the context.

### Current encoding

Suppose we have a circuit like
```
input: q (qubit)

q = h(q)
q = unsupported_1(q)
q = h(q)

output: q
```

We currently support this by encoding `unsupported_1` in a mid-circuit barrier over `q`.

But this is not always possible. In this example we have unsupported operations that are connected via non-register (i.e not qubit/bit) wires:

```
input: q (qubit)

rot: rotation = unsupported_0
q = rz(q, rot)
q: qubit, t: opaque = unsupported_1(q)
q = h(q)
t: opaque = unsupported_2(t)

output: q
```
The middle unsupported op can still be encoded in a barrier, but we now have two other ops that cannot be linked to a pytket registry so they cannot be put in a barrier in a sensible way.

We can instead keep the unsupported subgraphs in the encoding context (in `EncodedCircuit`), and re-insert them afterwards while decoding the circuit back into a hugr region.

That works because the subgraphs can be ordered either _before_ the circuit or _after_ it, so they don't introduce ordering dependencies between the pytket commands themselves.

### The problem

Now, the problematic case is when a mid-circuit unsupported op cannot be encoded as a barrier;

```
input: q0, q1

q1: qubit, t: opaque = unsupported_0(q1)
q1 = h(q1)
r: rotation = unsupported_1(t)
q2 = rz(q2, r)

output: q
```

Here the first `unsupported_0` op can be included as a barrier on `q1`.

For `unsupported_1` however, it cannot be mapped to a barrier directly even though it's introducing a strict ordering between the `h` and the `rz` ops.
(We could do a barrier on `q1` and `q2` here, but computing that in the general case gets expensive).

### Proposed solution

There is no nice way to implement this. The simplest solution would be adding the `rz`  from the previous example to the unsupported subgraph and emit a barrier, but that could greatly reduce the circuit exposed to pytket in some cases and limit the possible optimizations too much.

What I propose instead is to move from producing one circuit per region to instead allow partitioning the circuit into many when needed.

We should still keep the inline-opaque-barriers where possible, as splitting the circuits is also a big detriment to optimizations.

This will probably be an API breaking change, since we'll return multiple circuits per hugr region.

aborgna-q:
Fixes #1856 by extracting multiple circuits per hugr region, separated by unsupported graphs that must be ordered in-between them.

`EncodedCircuit` has new `contains_region` and `get_circuits` methods.
The old `get_circuit` method that returns a single circuit has been marked as deprecated.

This is a requirement for re-enabling default quantum optimizations in guppy.
I tested the fix on `guppylang` & `guppylang-docs`.

BEGIN_COMMIT_OVERRIDE
fix: Panic when applying a `PytketHugrPass` on complex regions (#1894)
END_COMMIT_OVERRIDE

### DIFF
--- a/tket/src/serialize/pytket/circuit.rs
+++ b/tket/src/serialize/pytket/circuit.rs
@@ -38,12 +38,13 @@ use super::opaque::OpaqueSubgraphs;
 /// [`EncodedCircuit::ensure_standalone`].
 #[derive(Debug, Clone)]
 pub struct EncodedCircuit<Node: HugrNode> {
-    /// Circuits encoded from independent dataflow regions in the HUGR.
+    /// Circuit segments grouped by their originating dataflow region.
     ///
     /// These correspond to sections of the HUGR that can be optimized
     /// independently.
     ///
-    /// Ordered as a depth-first pre-order.
+    /// Regions are ordered as a depth-first pre-order; segments retain their
+    /// execution order within each region.
     circuits: IndexMap<Node, EncodedCircuitInfo>,
     /// Sets of subgraphs in the HUGR that have been encoded as opaque barriers
     /// in the pytket circuit.
@@ -53,13 +54,30 @@ pub struct EncodedCircuit<Node: HugrNode> {
     opaque_subgraphs: OpaqueSubgraphs<Node>,
 }
 
+/// Identifies one pytket circuit segment within an encoded HUGR region.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct EncodedCircuitId<Node> {
+    /// The HUGR dataflow region from which the segment was encoded.
+    pub region: Node,
+    /// The segment's zero-based position within the region.
+    pub segment: usize,
+}
+
 /// Information stored about a pytket circuit encoded from a HUGR region.
 #[derive(Debug, Default, Clone)]
 pub(super) struct EncodedCircuitInfo {
-    /// The serial circuit encoded from the region.
-    pub serial_circuit: SerialCircuit,
-    /// Information about any unsupported nodes in the region that could not be encoded as a pytket command.
-    pub additional_nodes_and_wires: AdditionalNodesAndWires,
+    /// Runs of pytket commands encoded from the region.
+    pub serial_circuits: Vec<SerialCircuit>,
+    /// Opaque subgraphs that must be restored before, between, or after the
+    /// serial circuits.
+    ///
+    /// This always contains one more entry than [`Self::serial_circuits`].
+    pub boundaries: Vec<EncodedCircuitBoundary>,
+    /// List of wires that directly connected the input node to the output node in the encoded region,
+    /// and were not encoded in [`serial_circuits`].
+    ///
+    /// We just store the input nodes's output port and output node's input port here.
+    pub straight_through_wires: Vec<StraightThroughWire>,
     /// List of parameters in the pytket circuit in the order they appear in the
     /// hugr input.
     ///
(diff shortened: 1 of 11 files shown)


tket/src/serialize/pytket/decoder.rs
```diff
@@ -583,74 +583,26 @@ impl<'h> PytketDecoderContext<'h> {
...
-            .additional_subgraphs
+        boundaries
             .iter()
+            .flat_map(|boundary| &boundary.external_subgraphs)
             .flat_map(|subgraph| &subgraph.params)
             .map(|param| PytketParam::parse(param))
             .for_each(|param| {
                 param.visit_input_variables(&mut |name| {
-                    // `pi` is loaded as a constant rather than routed as a parameter.
                     if name != "pi" {
```
nicolaassolini-qntm, line 623: Is this the better way of checking this?

tket/src/serialize/pytket/decoder.rs
```diff
@@ -663,6 +615,79 @@ impl<'h> PytketDecoderContext<'h> {
...
+    ) -> Result<(), PytketDecodeError> {
+        let config = self.config().clone();
+        for com in commands {
+            let op_type = com.op.op_type;
+            self.process_command(com, config.as_ref())
+                .map_err(|e| e.pytket_op(&op_type))?;
+        }
+
+        Ok(())
+    }
```
nicolaassolini-qntm: for readability I would move this function above under `run_decoder`

tket/src/serialize/pytket/tests.rs
```diff
@@ -29,12 +29,12 @@ use crate::serialize::pytket::PytketEncodeError;
...
-    DecodeInsertionTarget, DecodeOptions, EncodeOptions, EncodedCircuit, PytketDecodeError,
-    PytketDecodeErrorInner, PytketDecoderConfig, PytketEncodeOpError, PytketEncoderConfig,
-    default_decoder_config, default_encoder_config,
+    DecodeInsertionTarget, DecodeOptions, EncodeOptions, EncodedCircuit, EncodedCircuitId,
+    PytketDecodeError, PytketDecodeErrorInner, PytketDecoderConfig, PytketEncodeOpError,
+    PytketEncoderConfig, default_decoder_config, default_encoder_config,
 };
 use hugr::hugr::hugrmut::HugrMut;
-use hugr::ops::handle::FuncID;
+use hugr::ops::handle::{FuncID, NodeHandle};
```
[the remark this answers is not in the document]
nicolaassolini-qntm: false right?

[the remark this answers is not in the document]
aborgna-q: Bad bot. That's a trait

tket/src/serialize/pytket/encoder.rs
```diff
@@ -745,12 +824,25 @@ impl<H: HugrView> PytketEncoderContext<H> {
         subencoder.run_encoder(hugr, node)?;
 
         let (info, opaque_subgraphs) = subencoder.finish(hugr, node)?;
-        if !info.output_params.is_empty() {
+        self.opaque_subgraphs = opaque_subgraphs;
+        if !info.output_params.is_empty()
+            || info.serial_circuits.len() != 1
```
nicolaassolini-qntm: why this check?

nicolaassolini-qntm (commented):
Just a some questions

nicolaassolini-qntm:
Also, code coverage is pretty bad, at first glance it seems that the uncovered code is mostly errors or failing paths, but it may be worth a double check.

tket/src/serialize/pytket/decoder.rs
aborgna-q: We really should update the pytket parser to avoid these hardcoded cases.

It shouldn't be in this PR, so I opened https://github.com/Quantinuum/tket2/issues/1900 and left a TODO here.

tket/src/serialize/pytket/encoder.rs
aborgna-q: This was a compatibility check for emitting pytket circboxes.
Note that this codepath was disabled, and we don't plan to emit boxes anymore.

I just removed the unused block of code.

tket/src/serialize/pytket/circuit.rs
```diff
@@ -657,93 +657,87 @@ impl<Node: HugrNode> EncodedCircuit<Node> {
...
     }
 
-    /// Returns a mutable iterator over every circuit segment and its ID.
-    pub fn iter_mut(
-        &mut self,
-    ) -> impl Iterator<Item = (EncodedCircuitId<Node>, &mut SerialCircuit)> {
+    /// Returns a mutable iterator over every circuit segment and its region.
+    ///
+    /// A region containing multiple segments appears once for each segment.
+    // TODO: Update signature to return `(EncodedCircuitId<Node>, &mut SerialCircuit)` in a breaking release.
```
nicolaassolini-qntm, line 677: Is there an issue related to these TODOs?

nicolaassolini-qntm (approved):

tket/src/serialize/pytket/circuit.rs
aborgna-q: Added links to https://github.com/Quantinuum/tket2/issues/1901

PR #1909, opened 2026-08-05
perf: improve Cliffordize performance

### LINKED ISSUE #1767: [Performance]: Cliffordize / RuleMatcher.apply_exhaustive scales quadratically with rewrites in a region
### Performance Issue

Hi, and thanks for the scoped `RuleMatcher.apply_exhaustive` helper and `Cliffordize` pass from #1654.

This report comes from my local work on a PR attempt for the non-Clifford-to-Clifford rewrite work during UnitaryHack 2026 / #1315. I was not able to submit my PR because of the UnitaryHack 2026 PR limit, so I continued testing the current `main` locally and found one performance/scalability issue that seems worth tracking, plus one smaller pass-result observation. I did not find a correctness problem with the merged pass in the cases I tested.

### Summary

`RuleMatcher.apply_exhaustive(...)`, as used by `Cliffordize`, currently applies one rewrite, rebuilds the whole `Circuit` view, restarts the matcher, then repeats until no match remains. For a region with `K` matching gates, this gives `K` full scans of the region. When `K` is proportional to the region size, runtime scales approximately quadratically.

I realize re-matching after each edit may be deliberate: it avoids stale matches, overlapping rewrites, and convexity/validity problems. So I am not assuming that simply applying all matches from one scan is always safe. The issue is that the current fixed-point strategy has a clear O(N^2) cost for the independent `T -> S` / `Tdg -> Sdg` replacements that `Cliffordize` performs.

I searched open and closed issues/PRs on 2026-06-28 for `Cliffordize quadratic`, `apply_exhaustive`, `RuleMatcher performance`, `find_matches_iter`, and `Cliffordize`. I found only #1315, #1653, #1654, and #1695 for `Cliffordize`, and no issue tracking this scaling behavior. Related matcher work exists in #1208, #833, #983, #1017, #1057, and #838, but I do not think this duplicates any of them.

### Steps to Reproduce

I tested against current `main` at `b3be2bb093d7b6e4b2d50de852aeaf0cafbf244a`.

My local build command was:

```bash
git fetch origin
git checkout origin/main
cd tket-py
VIRTUAL_ENV=../.venv PATH="../.venv/bin:$PATH" uvx maturin develop --release --uv
cd ..
```

Because my environment did not have `patchelf`, I ran with `LD_LIBRARY_PATH`:

```bash
LD_LIBRARY_PATH="$PWD/target/maturin" .venv/bin/python cliffordize_scaling.py
```

Script:

```python
# cliffordize_scaling.py
import time

from tket._state.build import H, OneQbGate, from_coms
from tket.passes import Cliffordize

T = OneQbGate("T")

print("Sequential T gates")
print(f"{'N':>6} {'rewrites':>9} {'time(s)':>9} {'time/N^2':>10} {'ratio':>8}")
prev = None
for n in (100, 200, 400, 800, 1600, 3200, 6400):
    # Circuit construction is deliberately outside the timed region.
    hugr = from_coms(*[T(0) for _ in range(n)]).to_python().modules[0]

    t0 = time.perf_counter()
    res = Cliffordize().run(hugr, inplace=False)
    dt = time.perf_counter() - t0

    ratio = f"x{dt / prev:.1f}" if prev else ""
    print(f"{n:>6} {res.results[-1][1]:>9} {dt:>9.4f} {dt/(n*n):>10.2e} {ratio:>8}")
    prev = dt

print()
print("One T in an otherwise-Clifford region")
print(f"{'N':>6} {'rewrites':>9} {'time(s)':>9} {'time/N':>10} {'ratio':>8}")
prev = None
for n in (200, 400, 800, 1600, 3200, 6400):
    hugr = from_coms(*([H(0) for _ in range(n - 1)] + [T(0)])).to_python().modules[0]

    t0 = time.perf_counter()
    res = Cliffordize().run(hugr, inplace=False)
    dt = time.perf_counter() - t0

    ratio = f"x{dt / prev:.1f}" if prev else ""
    print(f"{n:>6} {res.results[-1][1]:>9} {dt:>9.4f} {dt/n:>10.2e} {ratio:>8}")
    prev = dt
```

### Performance Metrics

```markdown

```

### Expected Performance

**Expected:** For independent `T -> S` replacements, I expected the fixed-point pass to avoid doing one full region scan per rewritten gate, or at least to have sub-quadratic scaling. A single rewrite in a growing region should be O(N), and many independent rewrites ideally would not multiply that by N.

**Actual:** Runtime grows close to quadratically with the number of rewrites in the region.

Optimized local build:

```text
Sequential T gates
     N  rewrites   time(s)   time/N^2    ratio
   100       100    0.0478   4.78e-06
   200       200    0.0697   1.74e-06     x1.5
   400       400    0.1954   1.22e-06     x2.8
   800       800    0.6499   1.02e-06     x3.3
  1600      1600    2.2862   8.93e-07     x3.5
  3200      3200    8.5589   8.36e-07     x3.7
  6400      6400   32.7053   7.98e-07     x3.8

One T in an otherwise-Clifford region
     N  rewrites   time(s)     time/N    ratio
   200         1    0.0375   1.88e-04
   400         1    0.0661   1.65e-04     x1.8
   800         1    0.1217   1.52e-04     x1.8
  1600         1    0.2574   1.61e-04     x2.1
  3200         1    0.5230   1.63e-04     x2.0
  6400         1    1.1539   1.80e-04     x2.2
```

The second table is the control case: one rewrite in a growing region scales roughly linearly. The first table rewrites every gate and shows the additional multiplication by the number of rewrites.

I also tested an interleaved two-qubit pattern (`T, H, T, CX` per layer) and saw the same shape:

```text
layers   gates  rewrites   time(s)    ratio
   100     400       200    0.1851
   200     800       400    0.4487     x2.4
   400    1600       800    1.4741     x3.3
   800    3200      1600    5.2027     x3.5
```

### Where it seems to come from

The loop in `tket-py/src/pattern.rs` applies one rewrite at a time:

```rust
while let Some(rewrite) = self.find_match(target)? {
    target
        .apply_rewrite(rewrite)
        .context("Could not apply exhaustive rule rewrite")?;
    rewrite_count += 1;
}
```

`find_match` rebuilds the `Circuit` view and takes only the first match:

```rust
pub fn find_match(&self, target: &CompilationState) -> PyResult<Option<PyCircuitRewrite>> {
    let circ = Circuit::try_new(&target.hugr)
        .map_err(|error| PyValueError::new_err(error.to_string()))?;
    let Some(pmatch) = self.matcher.find_matches_iter(&circ).next() else {
        return Ok(None);
    };
    Ok(Some(self.match_to_rewrite(pmatch, &circ)?))
}
```

So for `K` rewrites in one region, the current implementation pays for `K` `Circuit::try_new(...)` calls and `K` matcher restarts. `find_matches` already exists and returns all matches from one scan, which suggests the matcher can enumerate the independent `T`/`Tdg` matches cheaply; the expensive part is the repeated full restart.

### Small related observation: `PassResult.modified`

This is lower priority and may be intentional/shared behavior, but `Cliffordize` reports `modified=True` even when it performs zero rewrites:

```python
from tket._state.build import H, OneQbGate, from_coms
from tket.passes import Cliffordize

S = OneQbGate("S")
hugr = from_coms(H(0), S(0)).to_python().modules[0]
res = Cliffordize().run(hugr, inplace=False)

print(res.results)
print(res.modified)
```

Actual:

```text
[('Cliffordize', 0)]
True
```

I think this comes from the shared `PassResult.for_pass(..., modified=True)` default rather than from `Cliffordize` specifically. It is just more visible here because `Cliffordize` also returns the rewrite count. If the intended contract is that `modified=False` means the output is unchanged, then a zero-rewrite `Cliffordize` run seems like it could report `modified=False`. If the current behavior is intentional, this note can be ignored or split out.

### Verification

The existing `Cliffordize` tests pass for me on the same optimized build:

```bash
LD_LIBRARY_PATH="$PWD/target/maturin" .venv/bin/python -m pytest tket-py/test/test_cliffordize.py -q
```

```text
5 passed, 10 warnings in 0.61s
```

I also checked basic nested/control-flow/module cases locally and did not find a correctness issue. This report is about scalability of the current fixed-point matching strategy, not about the correctness of #1654.

### Component

Rust library

### Environment

```markdown
nvcr.io/nvidia/cuda:12.9.2-cudnn-devel-ubuntu24.04
Host: 128 GB RAM, 24 thread AMD CPU
```

### Optimization Suggestions

_No response_

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1767

This PR introduces a new `RuleMatcher.apply_all_matches_once` method that allows us to apply a rule in linear time.


tket-py/src/pattern.rs
nicolaassolini-qntm: This function is not used now. Not sure if we have to keep it.

tket-py/src/pattern.rs
```diff
@@ -144,8 +144,11 @@ impl RuleMatcher {
             .collect()
     }
 
-    /// Apply the first matching rule repeatedly within each circuit-compatible
-    /// region in the selected scope.
+    /// Apply the first matching rule repeatedly within each circuit-compatible region.
```
aborgna-q: Keep the first line of docs short, since it gets rendered as a summary sometimes.

```suggestion
    /// Apply the first matching rule repeatedly within each circuit-compatible region.
    ///
```

tket-py/src/pattern.rs
aborgna-q: If it's dead code we should remove it.

tket-py/src/pattern.rs
```diff
@@ -184,6 +187,75 @@ impl RuleMatcher {
...
+                    target
+                        .apply_rewrite(rewrite)
+                        .context("Could not apply exhaustive rule rewrite")?;
+                    rewrite_count += 1;
+                }
+                // We assume that applying a rewrite does not introduce new matches for the same rule
+                assert!(
+                    self.find_match(target)?.is_none(),
+                    "Applying rewrites should not introduce new matches for the same rule"
+                );
```
aborgna-q: Isn't this going against the aim of avoiding repeated traversals?

The assert is more than an assumption.

nicolaassolini-qntm: My idea is that this function must be used only when by searching once you can solve all the rules, that's why there is the assertion. We can remove it or update the doc string to make this behaviour clearer

aborgna-q (approved):

PR #1911, opened 2026-08-06
fix: `wire_inout` port mapping with interleaved classical and quantum wires under `dagger` modifier

nicolaassolini-qntm:
The previous `wire_inout` implementation assumed that quantum inputs and outputs occupied contiguous, matching positions.

However, after a controlled operation has been resolved, a classical input or output may appear between quantum ports.

For example:

```text
inputs:  [qubit, usize, qubit]
outputs: [qubit,        qubit]
```

In this case, the quantum inputs and outputs are still aligned: the first input qubit corresponds to the first output qubit, but there is a gap corresponding to the `usize` input. As a result, the resolution process could connect a classical value to a `qubit` port, producing an invalid HUGR.

The updated implementation:

* Preserves the original positions of classical input and output mappings.
* Collects quantum input and output indices independently.
* Reverses only the quantum data flow when applying `dagger`.


tket/src/modifier/modifier_resolver.rs
```diff
@@ -935,85 +935,81 @@ impl<N: HugrNode> ModifierResolver<N> {
         &mut self,
         (old_in, old_out): (N, N),
         (new_in, new_out): (Node, Node),
-        (mut inputs, mut outputs): (
+        (inputs, outputs): (
```
doug-q, line 953: Pre-existing, but perhaps worth fixing:

* It's better to take `impl IntoIterator<...>` as arguments than `impl Iterator`

* It's weird that old nodes are  `N`s and new nodes are `Node`s.

* This is a complicated function, I think work a docstring (even though fn is private) specifying what the parameters mean.

doug-q (approved):
Looks good. If you wanted to you could improve this function

tket/src/modifier/modifier_resolver.rs
nicolaassolini-qntm: I addressed points 1 and 3

> It's weird that old nodes are Ns and new nodes are Nodes.

Yes, but we would need a major refactor to change this.

PR #1931, opened 2026-08-17
feat: pg canonical form

yao-cqc:
This crate provides `CanonicalFormPass`. The canonical form it produces is defined in the README.


pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,1055 @@
...
+                        // If the pass is moving backward, we need to add the gates in reverse order.
+                        if !forward {
+                            decomposed_gates.reverse();
+                        }
+                        for sub_gate in decomposed_gates {
+                            let sub_gate = sub_gate.with_conditional(
+                                data.get_conditional_bits().clone(),
+                                data.get_conditional_values().clone(),
+                            );
+                            // We set cliff_eval to false so the gates are not composed into the tableau.
```
chloe-quantinuum: I'm not sure I understand why we use `cliff_eval` to prevent gates from being composed into the tableau. `cliff_eval` seems to be a boolean used to store the preference of the caller of the pass, rather than to ensure the logic is correct. If we don't want conditional gates to be composed into the tableau, isn't this already handled anyway by `sub_gates` inheriting their parent's `conditional_bits` and the check on line 23?

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,1055 @@
...
+            if let Op::ConditionalBox { data } = op {
+                let mut inner_ops = data.get_ops().clone();
+                inner_ops.reverse();
+                *data = ConditionalBoxData::new(
+                    inner_ops,
+                    data.get_conditional_bits().clone(),
+                    data.get_conditional_values().clone(),
+                );
+            }
+        }
```
chloe-quantinuum: It looks like this block is used to reverse the order of inner ops in conditional boxes if `!forward`, but doesn't line 499 already do this, or do these cover different cases? If the intended behaviour is for the ops to be reversed twice, ie. identity, could we remove this and line 499?

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,1055 @@
...
+fn add_rotation(
+    pg: &mut PauliGraph,
+    s: Vec<Pauli>,
+    theta: f64,
+    tab: &mut QubitMajorTableau,
+    data: &GateData,
+    forward: bool,
+    cliff_eval: bool,
+) {
+    let cliff_theta = cliff_angle(theta);
```
chloe-quantinuum: It doesn't look like the angle stored in this variable ever actually gets used. The condition in the following if statement might be more readable if we used `let is_clifford = cliff_angle(theta).is_some();` instead

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,1055 @@
...
+            }
+        }
+        Op::Rotation { data } => {
+            let (s, sign_bit) = tab.conjugate_string(data.get_string());
+            let theta = if sign_bit {
+                -data.get_angle()
+            } else {
+                data.get_angle()
+            };
+            let cliff_theta = cliff_angle(theta);
```
chloe-quantinuum: Same as line 22, the angle doesn't seem to ever actually be used so may be more readable to have `let is_clifford = cliff_angle(theta).is_some();`

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,1055 @@
...
+        Op::Reset { data } => {
+            let (z_string, mut z_sign_bit) = tab.conjugate_string(data.get_first_string());
+            let (x_string, mut x_sign_bit) = tab.conjugate_string(data.get_second_string());
+            z_sign_bit ^= data.get_first_sign_bit();
+            x_sign_bit ^= data.get_second_sign_bit();
+            pg.add_op(Op::Reset {
+                data: ResetData::new(z_string, x_string, z_sign_bit, x_sign_bit),
+            });
+        }
+        Op::ConditionalBox { data } => {
```
chloe-quantinuum: There appears to be a lot of overlap between the `Op::Rotation`, `Op::Measure`, and `Op::Reset` branches inside `Op::ConditionalBox` and those outside it. Would it be worth pulling out some of this duplicate code into helper functions? 

(there also seems to be quite a bit of overlap between those `Op::Rotation` branches and `add_rotation()`, though the different types in this case might make unification annoying)

chloe-quantinuum (commented):
Looks great, just a few comments about minor stylistic choices and potential code duplication

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
yao-cqc: Good point!

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
yao-cqc: You are right. I have updated the comment and pass through the cliff_eval

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
yao-cqc: The new code always processes each conditional box as a whole without messing with the order of its internal ops, so only the outer order is reversed at the end.

pg-libs/pg-passes/pg-canonical-form/src/lib.rs
```diff
@@ -0,0 +1,813 @@
...
+    } else {
+        new_pg.add_op(Op::Tableau {
+            data: TableauData::from(current_tableau),
+        });
+        // reverse the ops
+        let mut ops = new_pg.get_ops().clone();
+        ops.reverse();
+        new_pg = PauliGraph::new(pg.get_n_qubits()).with_ops(ops);
+    }
+    // Merge adjacent conditional boxes once both outer and inner ops are in circuit order.
```
yao-cqc, line 317: This is needed to fix a test regression caused by the new handling of conditional gates. It should be relatively cheap.

chloe-quantinuum (approved):
The refactor looks great!

PR #1943, opened 2026-08-20
feat: pg-optimise

yao-cqc:
This PR introduces two passes: `GroupCommutingOpsPass` and `RotationMergingPass`. See the README for details.


pg-libs/pg-passes/pg-optimise/README.md
```diff
@@ -0,0 +1,17 @@
...
+- `GroupCommutingOpsPass`
+Groups mutually commuting operations by inserting `SetBoundary` operations.
+Commutation is checked for rotations, measurements, resets, and conditional
+boxes; other operation types are treated as non-commuting.
+
+- `RotationMergingPass`
+Greedily merges Pauli rotations with the same Pauli string when they commute
+with every operation between them. Clifford rotations are folded into
+tableaux.
+Consider running `CanonicalFormPass` before either optimisation pass so that
```
cqc-alec, line 17:
```suggestion
tableaux.

Consider running `CanonicalFormPass` before either optimisation pass so that
```

pg-libs/pg-passes/pg-optimise/README.md
```diff
@@ -0,0 +1,17 @@
+# pg-optimise
+
+This crate contains 2 optimisation passes for `PauliGraph`s.
+
+## Passes
+
+- `GroupCommutingOpsPass`
```
cqc-alec, line 7: I think we should try to be consistent about using noun phrases or verb phrases in pass names. Unfortunately tket2 is already inconsistent, so not sure what to suggest!

pg-libs/pg-passes/pg-optimise/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+        }
+        bits.push(word);
+    }
+    bits
+}
+
+/// Checks whether two bit-packed Pauli strings commute.
+///
+/// The strings must use opposite encodings: one must have been packed with
+/// `z_first` set to `true`, and the other with it set to `false`.
```
cqc-alec, line 41: This seems like a weird condition. Is there a good reason for it? Can we not just settle on one convention (either X or Z first), and have this function handle it?

pg-libs/pg-passes/pg-optimise/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+            .get_paulis()
+            .map(|ps| ps.iter().map(|p| bitpack_paulis(p, z_first)).collect());
+        Self {
+            op,
+            strings,
+            classical_access,
+            z_first,
+        }
+    }
+    fn flip_to_zx(&mut self) {
```
cqc-alec: Would `coerce_to_zx` be a better name? Since it might already be Z-first? (And in that case, should we short-circuit the manipulation?)

pg-libs/pg-passes/pg-optimise/src/lib.rs
```diff
@@ -0,0 +1,316 @@
...
+#[cfg(test)]
+mod tests {
+    use pg_core::{MeasureData, ResetData, RotationData};
+
+    use super::*;
+
+    #[test]
+    fn test_commute_bitpacked() {
+        let p0 = vec![Pauli::X, Pauli::I, Pauli::Z];
+        let p1 = vec![Pauli::X, Pauli::I, Pauli::Z];
```
cqc-alec: These are the same, is that intentional?

pg-libs/pg-passes/pg-optimise/src/rotation_merging.rs
```diff
@@ -0,0 +1,645 @@
...
+        });
+
+        pg.add_op(Op::Rotation {
+            data: RotationData::new(vec![Pauli::X, Pauli::X], 0.4),
+        });
+        pg.add_op(Op::Rotation {
+            data: RotationData::new(vec![Pauli::X, Pauli::I], 0.4),
+        });
+        let merged_pg = rotation_merging(&pg);
+        assert_eq!(merged_pg.get_ops().len(), 2);
```
cqc-alec, line 327: Why is this not 1?

pg-libs/pg-passes/pg-optimise/src/rotation_merging.rs
```diff
@@ -0,0 +1,645 @@
...
+            },
+            Op::Rotation {
+                data: RotationData::new(vec![Pauli::X], 0.125),
+            },
+        ]);
+
+        let merged_pg = rotation_merging(&pg);
+
+        assert_eq!(merged_pg, pg);
+        assert!(compare_unitaries_via_tk(&pg, &merged_pg));
```
cqc-alec: This is redundant given the assertion above.

cqc-alec (commented):
Looks good! Just a few comments and queries.

pg-libs/pg-passes/pg-optimise/src/lib.rs
yao-cqc: Yes, I agree that it feels a bit awkward, but I think this requires the fewest bitwise operations. If both encodings use the same convention, we would need some extra bit shifts.

pg-libs/pg-passes/pg-optimise/src/rotation_merging.rs
yao-cqc: oh, because there is a tableau at the end

pg-libs/pg-passes/pg-optimise/README.md
yao-cqc: Do you prefer noun phrases or verb phrases? I could rename it to something like `CommutationPartitionPass`. What do you think?

cqc-alec: Since the codebase is already inconsistent in this regard and there doesn't seem to be a consensus on which convention to adopt, I am OK to leave it as it is for now.

pg-libs/pg-passes/pg-optimise/src/lib.rs
cqc-alec: Is this a big deal? I would be surprised if a few extra bit shifts were significant for performance. I think the most important thing is API simplicity and maintainability, and this kink does introduce a level of complexity which I'd rather avoid.

cqc-alec (commented):
Thanks, my only remaining concern is the `x_first`/`z_first` parameter, which I'd rather avoid unless there is a clear performance hit.

cqc-alec (approved):
Thanks!

Issue #1947, opened 2026-08-21
Investigate preservation of order edges between gates


nicolaassolini-qntm:
#1836 should have fixed the order edge preservation. However, order edges between gates are uncommon, and it is not easy to spot potential bugs. We should deep test everything.

Seems that substitution as [`CY`](https://github.com/Quantinuum/tket2/blob/f8309ac6171d851ef114d780f5e39d6a37630bb4/tket/src/modifier/modifier_resolver/tket_op_modify.rs#L175) may lose some order edges.

PR #1948, opened 2026-08-21
fix: Wrong decomposition of `CnZ` gate

### LINKED ISSUE #1944: [Bug]: `CnZ` decomposition
### Description

In the modifier resolver passes, when the decomposition of a CnZ is required, we map it to a CRz(pi). However, these two operations are equal up to a global phase, and this introduces an error when controlled.
We should change the decomposition.

### Steps to Reproduce

```python
from guppylang.std.angles import angle
from guppylang.std.quantum import discard, x, discard_array, crz, cz, z
from guppylang.std.debug import state_result

@guppy
def main_cr() -> None:
    q = qubit()
    cs = array(qubit())
    h(cs[0])
    x(q)
    crz(cs[0], q, angle(1))
    state_result("r", cs[0], q)
    discard(q)
    discard_array(cs)

@guppy
def main_cz() -> None:
    q = qubit()
    cs = array(qubit())
    h(cs[0])
    x(q)
    cz(cs[0], q)
    state_result("r", cs[0], q)
    discard(q)
    discard_array(cs)

@guppy
def main_ctrl_z() -> None:
    q = qubit()
    cs = array(qubit() for _ in range(3))
    h(cs[0])
    x(cs[1])
    x(cs[2])
    x(q)
    with control(cs[0], cs[1], cs[2]):
        z(q)
    state_result("r", cs[0], cs[1], cs[2], q)
    discard(q)
    discard_array(cs)

from pathlib import Path

import numpy as np
import numpy.typing as npt

def format_statevector(
    state: npt.NDArray[np.complexfloating], threshold: float = 1e-6
) -> str:
    """Pretty-print a statevector, omitting amplitudes below *threshold*.

    Each basis state is shown as a zero-padded binary string, e.g.::

        000 -> 0.7071+0.j, 111 -> 0.7071+0.j
    """
    n_qubits = int(np.round(np.log2(len(state))))
    parts = []
    for idx, amp in enumerate(state):
        if abs(amp) > threshold:
            label = format(idx, f"0{n_qubits}b")
            parts.append(f"\t{label} -> {amp:.4g}")
    return "\n".join(parts) if parts else "all amplitudes below threshold"

print("Native CR(PI)")
state = main_cr.emulator(n_qubits=2).statevector_sim().run()
res = state.partial_state_dicts()[0]["r"].as_single_state()
print(f"{format_statevector(res)}")
print("===\nNative Cz")
print(format_statevector(main_cz.emulator(n_qubits=2).statevector_sim().run().partial_state_dicts()[0]["r"].as_single_state()))
print("===\n with control: z")
print(format_statevector(main_ctrl_z.emulator(n_qubits=4).statevector_sim().run().partial_state_dicts()[0]["r"].as_single_state()))
```
output:
```
Native CR(PI)
	01 -> 0.7071+0j
	11 -> 0+0.7071j
===
Native Cz
	01 -> 0.7071+0j
	11 -> -0.7071+0j
===
 with control: z
	0111 -> 0.7071+0j
	1111 -> 0+0.7071j
```

### Actual Behaviour

Basically, what we do is
`CnZ -> CnRz(pi)`

```

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by nicolaassolini-qntm**
closes #1944

While fixing the issue, I also refactored a bit of `modify_tket_op` to avoid overlapping in the match conditions.

I haven't added any tests since there is no easy way to test these changes (we need https://github.com/Quantinuum/tket2/pull/1886 to properly test the `modify_tket_op` function)

### CODE REVIEW

File: tket/src/modifier/modifier_resolver.rs [hunk 1]
```diff
@@ -1780,7 +1780,7 @@ mod tests {
                     .is_some_and(|function| function.func_name().starts_with("__modified__"))
             })
             .unwrap();
-        assert_eq!(state_order_edge_count(&h, modified_function), 26);
+        assert_eq!(state_order_edge_count(&h, modified_function), 16);
```
nicolaassolini-qntm, line 1783: Order edges seem fine; further investigations are supposed to be done in https://github.com/Quantinuum/tket2/issues/1947

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -59,10 +59,11 @@ impl<N: HugrNode> ModifierResolver<N> {
         }
         match tket_op {
             X | CX | Toffoli | Y | CY | Z | CZ | S | Sdg | T | Tdg | V | Vdg | H
-                if (control == 0)
-                    || (control < 3 && tket_op == X)
+                if (control == 0 && dagger)
```
aborgna-q, line 62: This was implied already by the `if control == 0 && !dagger` higher up, but I agree that it's better to make it explicit 👍

tket/src/modifier/modifier_resolver/tket_op_modify.rs
```diff
@@ -59,10 +59,11 @@ impl<N: HugrNode> ModifierResolver<N> {
...
+                    || ((1..3).contains(&control) && tket_op == X)
```
aborgna-q: - `1..=2` would be clearer than `1..3` here.
- Even better, just be explicit


```suggestion
                    || ((control == 1 && tket_op == X)
                    || ((control == 2 && tket_op == X)
```

aborgna-q (approved):

PR #1965, opened 2026-09-03
feat(rs)!: Update hugr to 0.30.1

aborgna-q:
BREAKING CHANGE: Updated public dependency `hugr` to `0.30.0`.

Bumps the dependency version.
The main noise here is replacing calls to `.signature()` that only used it to query info about the ports with the new efficient port query APIs from https://github.com/Quantinuum/hugr/pull/3149.


tket/benches/benchmarks/normalize.rs
```diff
@@ -4,10 +4,11 @@ use std::hint::black_box;
 
 use criterion::{AxisScale, BatchSize, Criterion, PlotConfiguration, criterion_group};
 use hugr::Hugr;
+use tket::extension::REGISTRY;
 use tket::passes::{ComposablePass, Normalize};
 
 fn load_hugr(bytes: &[u8]) -> Hugr {
-    Hugr::load(bytes, None).unwrap()
+    Hugr::load(bytes, Some(&REGISTRY)).unwrap()
```
maximilianruesch: Can you please explain this sudden change?

tket/src/serialize/pytket/opaque/subgraph.rs
```diff
@@ -64,16 +63,12 @@ impl<N: HugrNode> OpaqueSubgraph<N> {
 
         for &node in &nodes {
             let op = hugr.get_optype(node);
-            let Some(signature) = op.dataflow_signature() else {
-                continue;
-            };
```
maximilianruesch, line 69: This skip removal may be a semantic change. What is its replacement?

tket/src/serialize/pytket/extension/prelude.rs
```diff
@@ -35,12 +35,12 @@ impl<H: HugrView> PytketEmitter<H> for PreludeEmitter {
...
-                sig.input()
-                    .iter()
-                    .chain(sig.output().iter())
-                    .any(|ty| encoder.config().type_to_pytket(ty).is_none())
-            }) {
+            if hugr
+                .value_types(node, Direction::Incoming)
+                .chain(hugr.value_types(node, Direction::Outgoing))
+                .map(|(_, ty)| ty)
+                .any(|ty| encoder.config().type_to_pytket(&ty).is_none())
```
maximilianruesch:
```suggestion
                .any(|(_, ty)| encoder.config().type_to_pytket(&ty).is_none())
```

tket/src/passes/untuple.rs
```diff
@@ -166,7 +166,7 @@ fn make_rewrite<'h, T: HugrView>(
         return None;
     }
 
-    let tuple_types = op.dataflow_signature().unwrap().input_types().to_vec();
+    let tuple_types = hugr.in_value_types(node).map(|(_, ty)| ty).collect_vec();
```
maximilianruesch, line 169: You may even be able to avoid the `collect_vec`, since the vector is only used in a reference below. I have not looked into the function that consumes the slice, but if it is only iterated once you should be able to just pass the iterator itself.

tket/src/passes/dataflow/datalog.rs
```diff
@@ -190,26 +188,27 @@ fn run_datalog<V: AbstractValue, H: HugrView>(
...
-          if sig.output_ports().contains(p);
+          let op = hugr.get_optype(*n),
+          if p.index() < op.value_output_count();
 
         // Assemble node_in_value_row from in_wire_value's
-        node_in_value_row(n, ValueRow::new(sig.input_count())) <-- node(n), if let Some(sig) = hugr.signature(*n);
-        node_in_value_row(n, ValueRow::new(hugr.signature(*n).unwrap().input_count()).set(p.index(), v.clone())) <-- in_wire_value(n, p, v);
+        node_in_value_row(n, ValueRow::new(op.value_input_count())) <-- node(n),
+          let op = hugr.get_optype(*n),
+          if op.value_input_count() > 0 || op.value_output_count() > 0;
```
maximilianruesch, line 203: I suppose this is an extra performance optimisation?

tket/src/passes/normalize.rs
maximilianruesch, line 1: Is this related to the registry change?

maximilianruesch (changes requested):

tket/benches/benchmarks/normalize.rs
aborgna-q: The update caused the `t_factory` normalize benchmark to start failing, and I assumed this was an oversight in loading the files.

But looking closer into it, the `t_factory` hugr had the extension already bundled. There was a latent bug in hugr that got uncovered by the move to arc-types: https://github.com/Quantinuum/hugr/pull/3242

tket/src/passes/dataflow/datalog.rs
aborgna-q: Yeah, it's to avoid calling `signature().is_some()`.

tket/src/passes/normalize.rs
aborgna-q: It got added when I was investigating the benchmark registry error, yeah. I'll revert it.

tket/src/passes/untuple.rs
aborgna-q: The slice got `to_vec` and put in an op definition further on, so I changed the `remove_pack_unpack` argument directly to a `Vec` instead.

tket/src/serialize/pytket/opaque/subgraph.rs
aborgna-q: The old skip was wrong. It ignored non-dataflow nodes in subgraphs and marked them as SiblingSubgraph-compatible anyways.

This is a semantic change, but it flags this problem at the definition rather than panicking down the line.

maximilianruesch (approved):
🚀

PR #1968, opened 2026-09-03
feat: pg-greedy-synth

yao-cqc:
This PR introduces `GreedySynthPass`, which is designed as a faster version of the `GreedyPauliSimp` pass in pytket. It also includes a simple rebase pass that maps Clifford gates acting on two qubits to CX gates only.

An `ARCHITECTURE.md` file is included to make the codebase easier to understand and maintain.

For now, these two crates have not been added to the workspace, so they are not compiled or checked by CI. This is because two crates they depend on are still being merged.


pg-libs/pg-passes/pg-greedy-synth/src/frontier/mod.rs
```diff
@@ -0,0 +1,1038 @@
...
+            tqe_cost: 0,
+            support_pairs: 0,
+            reduction_gates: 0,
+        }
+    }
+
+    /// Recomputes the cost of a single string as one less than its support size
+    /// and updates the number of TQE candidates that reduce it.
+    fn recompute(&mut self) {
+        assert!(!self.supports.is_empty());
```
sjdilkes: Are Identity rotations , i.e. nothing, rejected earlier in the compilation flow than this? Identity rotations should never be introduced at any point, so wondering why this check is also needed here.

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
```diff
@@ -0,0 +1,268 @@
...
+    window_size: Option<usize>,
+    pool_size: Option<usize>,
+    top_up_size: Option<usize>,
+) -> (usize, usize, usize) {
+    let window_size = window_size.unwrap_or(DEFAULT_WINDOW_SIZE);
+    let pool_size = pool_size.unwrap_or(max(
+        MINIMUM_POOL_SIZE,
+        ((n_qubits * n_qubits) as f64 * 0.2) as usize,
+    ));
+    let top_up_size = top_up_size.unwrap_or(max(MINIMUM_TOP_UP_SIZE, pool_size / n_qubits));
```
sjdilkes, line 76: test_empty_graph::case_1_zero tests for empty zero-qubit graphs, which would cause an issue here with `n_qubits`, as could a global phase rotation - are we confident that these cases can't be hit? Do we need an assertion first or similar

pg-libs/pg-passes/pg-greedy-synth/src/reducer/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+                        FrontSetKind::Conditional(slice.decode_conditional_box(index))
+                    }
+                    _ => FrontSetKind::Ordinary,
+                }
+            }
+        };
+
+        match set_kind {
+            FrontSetKind::BlackBox(data, tableau) => {
+                let mut nested_depth = vec![0; qubit_depth.len()];
```
sjdilkes: Is it intentional that nested synthesis doesn’t contribute to the outer depth state?
- In reducer/mod.rs:206, nested_depth starts from zero and is discarded after synthesising the tableau before a black box
- Recursive conditional synthesis in reducer/mod.rs:227 also tracks its depth independently
- Candidate selection in reducer/gate_pool.rs:135 continues using the outer qubit_depth, which still contains TQEs emitted before these operations

sjdilkes: This seems to include some earlier TQEs in the depth heuristic while ignoring those emitted during nested synthesis, meaning the depth might not be accurate/lower than expected? If the intention is to optimise each barrier-separated segment independently, should the outer depth be reset explicitly? Or we could carry the nested depth forward?

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+        } else {
+            self.single_mask[first.chunk] &= !(1 << first.bit);
+        }
+    }
+
+    /// Logically retires the first visible set.
+    pub(crate) fn progress_set(&mut self) -> bool {
+        assert!(!self.is_empty(), "cannot progress an empty packed slice");
+        self.len -= self.sets[self.front_set].len();
+        self.front_set += 1;
```
sjdilkes, line 424: Does the packed window ever remove data that's no longer needed?
- packed_pg_slice/mod.rs:423-424 reduces `len` and advances `front_set`, but I can’t see where the no longer neeed data is removed from `sets`, the packed rows, masks, or barriers
- packed_pg_slice/mod.rs:220 tracks next_column, which also seems to keep increasing
- reducer/mod.rs:93-95 refills the visible window by appending more sets
So,`window_size bounds the amount of lookahead used for scoring, but perhaps not the underlying memory usage. Is that intentional? If so, it might be worth clarifying

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+            }
+            self.next_push_starts_new_set = true;
+        } else {
+            if self.sets.is_empty() || self.next_push_starts_new_set {
+                self.sets.push(Vec::new());
+                self.next_push_starts_new_set = false;
+            }
+            self.sets.last_mut().unwrap().push(packed);
+        }
+        self.len += 1;
```
sjdilkes, line 700: Could it be beneficial to track the packed size separately from the number of visible operations?
- packed_pg_slice/mod.rs:634-644 packs each operation inside a conditional box into the bit rows
- The box is then counted as one visible operation here, which seems reasonable for len itself
- However, slice.len() is also used for the window limit in reducer/mod.rs:93 and as part of the automatic parallelism estimate in reducer/mod.rs:314

Could that underestimate the amount of packed work for a large conditional box? Perhaps this is intentional, but I wondered whether a separate packed-column or chunk count would be a better input for those decisions

OTOH - you mentioned that conditionals are underperforming currently, so maybe a future issue?

pg-libs/pg-passes/pg-greedy-synth/src/reducer/gate_pool.rs
```diff
@@ -0,0 +1,205 @@
...
+            q0_version: self.qubit_versions[tqe.q0],
+            q1_version: self.qubit_versions[tqe.q1],
+            cost,
+            stop,
+        });
+    }
+
+    /// Invalidates every candidate touching the selected TQE qubits by
+    /// advancing the version for each qubit without scanning the pool.
+    pub(crate) fn invalidate(&mut self, tqe: TQE) {
```
sjdilkes, line 93: Can this pool accumulate stale candidates if the front set is large?
- `gate_pool.rs:93-95` invalidates candidates by updating the qubit versions, so the existing entries remain in the skip set
- `gate_pool.rs:123-130` removes stale entries as they are encountered while looking for the next five current candidates, but can higher-cost stale entries remain untouched if they're never in the bottom 5?
- `reducer/mod.rs:278-283` then samples another pool_size or top_up_size candidates on each iteration
- The pool is cleared for each new front set at `reducer/mod.rs:265`, so this would at least be limited to one set

I assume this is intentional to scanning the pool during every invalidation, but do we know how large it tends to grow? and what's the potential bottleneck from having this set potentially grow very large?

pg-libs/pg-passes/pg-greedy-synth/src/backend/greedy.rs
```diff
@@ -0,0 +1,200 @@
...
+        let mut current_weight = 1.0;
+        let mut final_op_has_tableau_weight = false;
+
+        for (set_offset, set) in slice.op_sets().enumerate() {
+            if set_offset != 0 {
+                current_weight *= super::ALPHA;
+            }
+
+            for (index, view) in set {
+                final_op_has_tableau_weight = matches!(
```
sjdilkes: I might be misunderstanding what `final_op_has_tableau_weight` is meant to represent, but should this depend on the final operation rather than the final set as a whole?
- `backend/greedy.rs:48-51` overwrites the flag for every operation, so only the last operation visited has any effect
- `backend/greedy.rs:65-68` then gives the trailing tableau either its discounted weight or zero based on that flag

I think this means a final commuting set containing a disjoint reset and rotation gets a nonzero tableau weight if the rotation is last, but zero if the reset is last. If those operations can safely be reordered within the set, that would make the heuristic depend on their input order.

Could we add a comment or test explaining why the last operation is the relevant one? Otherwise, should this be calculated from the final set as a whole?

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
```diff
@@ -0,0 +1,268 @@
...
+        return PauliGraph::new(pg.get_n_qubits());
+    }
+
+    let (window_size, pool_size, top_up_size) = resolve_sizes(
+        pg.get_n_qubits(),
+        config.window_size,
+        config.pool_size,
+        config.top_up_size,
+    );
+    if average_rotations_per_set(pg) >= GROUPED_WEIGHTED_SUM_MIN_ROTATIONS_PER_SET {
```
sjdilkes: Is the average number of rotations the best signal for choosing the weighted-sum strategy here?
- `synthesis/greedy.rs:47-57` only counts top-level rotations
- `synthesis/greedy.rs:123` uses that count to decide between the grouped and expanded strategies
- However, measurements and resets are also assigned weights in `backend/weighted_sum/expanded.rs:29-34` and `backend/weighted_sum/grouped.rs:28-34`

This means a graph with large measurement or reset sets could still select the expanded strategy because it contains few rotations. 

My assumption is that this is realistic for actual quantum circuits so not an issue, but, would the number or density of weighted packed positions be a more representative measure?

pg-libs/pg-passes/pg-greedy-synth/src/reducer/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+    ) {
+        assert!(frontier.is_empty());
+        frontier.rebuild(slice);
+        self.cost_backend.update(slice);
+        pool.clear();
+        debug_assert!(pool.is_empty());
+
+        loop {
+            let emissions = frontier.pop_ops();
+            let emitted_gate_count = emissions.iter().map(|item| item.ops.len()).sum::<usize>();
```
sjdilkes: This is only used to check if things are emitted, so could this just be `let emitted_any = !emissions.is_empty()`?

pg-libs/pg-passes/pg-greedy-synth/src/reducer/gate_pool.rs
```diff
@@ -0,0 +1,205 @@
...
+
+    for (index, candidate) in candidates.iter().enumerate() {
+        if candidate.cost < minimum_gate_cost {
+            minimum_gate_cost = candidate.cost;
+            minimum_gate_cost_index = index;
+        }
+        maximum_gate_cost = maximum_gate_cost.max(candidate.cost);
+    }
+    for (index, depth) in depths.iter().copied().enumerate() {
+        if depth <= minimum_depth {
```
sjdilkes: Should this be `<` rather than `<=`, for consistency with the minimum-cost calculation above? Currently equal minimum depths select the last candidate rather than the first

pg-libs/pg-passes/pg-rebase/src/tqe_rebase.rs
```diff
@@ -0,0 +1,299 @@
...
+    pub fn with_allowed_tqes(mut self, allowed_tqes: Vec<GateType>) -> Self {
+        self.allowed_tqes = allowed_tqes.into_iter().collect();
+        self.allowed_tqes.insert(GateType::ZX);
+        self
+    }
+}
+
+impl PGPass for RebaseTQEToZXPass {
+    /// Rebases every disallowed TQE gate and leaves all other operations intact.
+    ///
```
sjdilkes: unnecessary commented out line here

pg-libs/pg-passes/pg-greedy-synth/src/backend/greedy.rs
```diff
@@ -0,0 +1,200 @@
...
+            tableau_weight: 0.0,
+            has_pair_ops: false,
+        }
+    }
+
+    /// Rebuilds the black box weights and trailing tableau weight for the
+    /// current slice.
+    ///
+    /// Each later commuting set is discounted by `ALPHA = 0.588`. This quality
+    /// tuning heuristic is inherited from the original implementation and
```
sjdilkes: "Original" is pytket `GreedyPauliSimp`? maybe add link here

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
```diff
@@ -0,0 +1,822 @@
...
+
+        let packed = match op {
+            Op::Rotation { data } => self.pack_rotation(data, backend),
+            Op::Measure { data } => self.pack_measure(data, backend),
+            Op::Reset { data } => self.pack_reset(data, backend),
+            Op::ConditionalBox { data } => self.pack_conditional(data, backend),
+            Op::BlackBox { data } => self.pack_black_box(data),
+            _ => panic!("unsupported operation in packed slice"),
+        };
+        // TODO: Some downstream code assumes that boxes are stored in standalone
```
sjdilkes, line 454: What's the status of this TODO, could you add an issue?

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
```diff
@@ -0,0 +1,268 @@
...
+            Op::Rotation { .. } => {
+                rotations += 1;
+                current_set_has_rotation = true;
+            }
+            _ => {}
+        }
+    }
+    rotation_sets += usize::from(current_set_has_rotation);
+
+    if rotation_sets == 0 {
```
sjdilkes: Try `rotations.checked_div(rotation_sets).unwrap_or(0)`

sjdilkes (changes requested):
Could we add pg-greedy-synth to the nightly SIMD test job as part of this PR?
The stable workspace tests won’t enable this feature, and the nightly command in `.github/workflows/ci.yml:179` currently tests SIMD only for `pg-bitpacked` and `pg-qm-tableau`. This means the public SIMD synthesiser and its feature-gated tests won’t be compiled in CI.

pg-libs/pg-passes/pg-greedy-synth/src/frontier/mod.rs
yao-cqc: Semantic checks, such as checking for non-identity Pauli strings and valid tableau data currently don't exist anywhere in the flow because they're relatively expensive. However, I plan to add them in the future. This assertion is just a sanity check, but I think it would be better to make the panic more explicit.

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
yao-cqc: PauliGraph currently doesn't support legless ops, so a PauliGraph with 0 qubits must be empty and shouldn't reach this code. Still, I think we should check for this to prevent regressions.

yao-cqc: Added a panic

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
yao-cqc: good point, updated the logic to include resets and measurements

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
yao-cqc: it doesn't, the front sets are only logically dropped. Memory optimisation should be an isolated future investigation.

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
yao-cqc: This is just a reminder for a potential performance improvement. There are similar markers in the other crates, and I plan to group them into fewer issues to avoid flooding the backlog.

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
yao-cqc: Right, good catch. I think this should be part of a broader issue about improving conditionals.

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
yao-cqc: This and the len() issue below will probably go into the same github issue.

pg-libs/pg-passes/pg-greedy-synth/src/reducer/gate_pool.rs
yao-cqc: Yes, this lazy removal is intentional. I think it could become an issue if the circuit is very wide, the front set is very large, and the pool size is also large. In that case, I'd expect the time spent on costing to become a bottleneck first. I suppose I can include this in a broader investigation into GatePool performance.

pg-libs/pg-passes/pg-greedy-synth/src/reducer/mod.rs
yao-cqc: Yes, this is a leftover from a prev refactor. It now uses the outer depth vector.

pg-libs/pg-passes/pg-greedy-synth/src/backend/greedy.rs
yao-cqc: This loop was badly written, it's now more clear. Each set will compound the weight once.

yao-cqc (commented):
@sjdilkes Hopefully I've addressed all your comments. I've also listed the follow-up tasks here, ready to turn into issues once the PR is merged.

## Improve conditional and blackbox handling in greedy synthesis
- Remove the constraint that blackboxes and conditional boxes must occupy standalone sets, so they can share sets with other operations.
- Count operations inside conditionals towards the slice’s operation count to make the lookahead window more accurate.

## Investigate memory usage in PackedPGSlice
- Retired operations are currently only logically removed, so their memory remains allocated.
- Investigate whether this causes meaningful memory overhead in real world usage. This has not yet been shown to be a problem.
- If needed, consider periodically removing retired operations to reclaim memory.

## Investigate and improve GatePool performance
- Benchmark lazy removal of stale candidates, especially with wide circuits, large front sets, and large pools.
- Measure pool growth within each front set and its effect on memory and runtime, including how it compares with the cost of evaluating candidates.
- Test sequential insertion: each worker produces a vector of candidates, then the main thread inserts them into the pool.
- Test multiple pools: assign each worker to a pool, then select the final candidates from the top entries across the pools.

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/greedy.rs
sjdilkes: Great, thanks!

pg-libs/pg-passes/pg-greedy-synth/src/packed_pg_slice/mod.rs
sjdilkes: Great, thanks!

pg-libs/pg-passes/pg-greedy-synth/ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+The pass chooses a backend and weight strategy, then creates a `Reducer`. The
+reducer manages the packed lookahead data, the current frontier, candidate
+scoring and the output graph.
+
+## Entry points
+
+`GreedySynthPass` is the scalar entry point. `GreedySynthSimdPass` is available
+with the `simd` feature and uses portable SIMD operations. Both implement
```
cqc-alec, line 25: Since these do the same thing, ideally they would be one and the same pass, which can detect whether `simd` is available or have a flag to dictate its use. Was this considered?

pg-libs/pg-passes/pg-greedy-synth/ARCHITECTURE.md
```diff
@@ -0,0 +1,133 @@
...
+
+The scalar pass uses `ScalarBackend`. The SIMD pass uses `SimdBackend`. Each
+backend implements both `PackedBackend` and `CostKernel`, so the same backend
+type is used to update packed data and to calculate raw cost changes.
+
+## Main structs
+
+### `Reducer<P, C>`
+
+`Reducer` runs the synthesis. `P` updates the packed data and `C` scores TQE
```
cqc-alec, line 37: Please define "TQE" on first use.

pg-libs/pg-passes/pg-rebase/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "pg-rebase"
+version = "0.1.0"
+edition = { workspace = true }
```
cqc-alec, line 4: This needs to be added to `workspace.members` in `pg-libs/Cargo.toml`. (Similarly for `pg-greedy-synth`.)

pg-libs/pg-passes/pg-greedy-synth/src/reducer/mod.rs
```diff
@@ -0,0 +1,320 @@
...
+                output.add_op(Op::Gate {
+                    data: GateData::new(GateType::BlackBox, data.get_qubits().clone())
+                        .with_data(data.get_content().clone()),
+                });
+            }
+            FrontSetKind::Conditional(data) => {
+                let inner = PauliGraph::new(qubit_depth.len()).with_ops(data.get_ops().clone());
+                let canonical = CanonicalFormPass::new().transform(&inner);
+                let grouped = GroupCommutingOpsPass::new().transform(&canonical);
+                let reduced = self.reduce(&grouped);
```
cqc-alec, line 226: When we recursively call `reduce()` here, should we be updating the high-level depth statistics with the additional depths computed?

pg-libs/pg-passes/pg-greedy-synth/src/synthesis/parallel_mode.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// Use parallel execution when the workload reaches 150,000.
+    Auto,
+}
+
+impl ParallelMode {
+    /// Reports whether the two workload dimensions justify parallel execution.
+    ///
+    /// Auto mode uses parallel costing at a workload product of 150,000. This
+    /// inherited tuning threshold should only change after representative
+    /// benchmarks.
```
cqc-alec, line 17: There are several comments similar to this in the PR, relating to different "magic numbers". I think it would be worth documenting these somewhere for easy reference, ideally with justifications for their choice.

pg-libs/pg-passes/pg-greedy-synth/ARCHITECTURE.md
yao-cqc: Yes, this was considered. Will raised a similar point about the SIMD feature in the QubitMajorTableau crate. The SIMD feature is still experimental, and the current setup lets us run both versions in the same build so we can easily compare them. I’m planning another round of investigation to see how SIMD performs compared to the scalar version. Once that’s done, we’ll have a better idea of what the interface should look like.

pg-libs/pg-passes/pg-greedy-synth/src/reducer/mod.rs
yao-cqc: Makes sense, done.

sjdilkes (approved):
LGTM!

pg-libs/pg-passes/pg-greedy-synth/TUNING.md
```diff
@@ -0,0 +1,25 @@
...
+## Default sizing constants
+
+Defined in [synthesis/greedy.rs](src/synthesis/greedy.rs).
+
+| Parameter             | Value  | Purpose                                                                                                            |
+| --------------------- | ------ | ------------------------------------------------------------------------------------------------------------------ |
+| `DEFAULT_WINDOW_SIZE` | `1280` | Default target number of operations in the costing window.                                                              |
+| `MINIMUM_POOL_SIZE`   | `1000` | Minimum default number of TQE candidates sampled after frontier progress.                                                                        |
+| `MINIMUM_TOP_UP_SIZE` | `200`  | Minimum default number of fresh TQE samples added without frontier progress, replenishing invalidated candidates.                                             |
+| `TQE_SAMPLE_FRACTION` | `0.1`  | Samples roughly 10% of the candidates for a Pauli string with maximum weight. The minimum pool size still applies. |
```
cqc-alec:
```suggestion
| `TQE_SAMPLE_FRACTION` | `0.1`  | Rough proportion of the candidates for a Pauli string with maximum weight to sample. The minimum pool size still applies. |
```

pg-libs/pg-passes/pg-greedy-synth/TUNING.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Internal heuristics
+
+| Parameter | Value | Purpose | Source |
+| --- | --- | --- | --- |
+| `PARALLEL_COSTING_THRESHOLD` | `150_000` | Auto mode uses parallel costing when `work_items * candidates >= 150_000`, to offset parallel overhead. `work_items` is the number of visible operations or tableau generators being costed. | [parallel_mode.rs](src/synthesis/parallel_mode.rs) |
```
cqc-alec:
```suggestion
| `PARALLEL_COSTING_THRESHOLD` | `150_000` | Auto mode uses parallel costing when `work_items * candidates` exceeds this threshold, to offset parallel overhead. `work_items` is the number of visible operations or tableau generators being costed. | [parallel_mode.rs](src/synthesis/parallel_mode.rs) |
```

pg-libs/pg-passes/pg-greedy-synth/TUNING.md
```diff
@@ -0,0 +1,25 @@
...
+| `GROUPED_WEIGHTED_SUM_MIN_OPS_PER_SET` | `64` | Uses `GroupedWeightedSum` when the average set size reaches this threshold. | [greedy.rs](src/synthesis/greedy.rs) |
+| `SPARSE_WEIGHTING_MAX_DENSITY` | `0.25` | `ExpandedSimdWeightedSum` strategy uses sparse iteration below this mask density, avoiding work on zero bits. | [expanded.rs](src/backend/weighted_sum/expanded.rs) |
+| `ALPHA` | `0.588` | Discounts the weight of each successive set, favouring nearby operations. | [greedy.rs](src/backend/greedy.rs) |
```
cqc-alec:
```suggestion
| `ALPHA` | `0.588` | Factor by which to discount the weight of each successive set, favouring nearby operations. | [greedy.rs](src/backend/greedy.rs) |
```

pg-libs/pg-passes/pg-greedy-synth/ARCHITECTURE.md
```diff
@@ -0,0 +1,138 @@
...
+
+## Main structs
+
+### `Reducer<P, C>`
+
+`Reducer` runs the synthesis. `P` updates the packed data and `C` scores
+two qubit entangling (TQE) gate candidates.
+
+> [!NOTE]
+> TQE gates are XX, XY, XZ, YY, YZ and ZZ. Each can be obtained by
```
cqc-alec:
```suggestion
> The TQE gates are XX, XY, XZ, YY, YZ and ZZ. Each can be obtained by
```

cqc-alec (approved):
Some minor rewording suggestions but LGTM!

PR #1970, opened 2026-09-03
feat!: Remove deprecated definitions

aborgna-q:
Remove some deprecated definitions in preparation for `tket-rs 0.22.0`

BREAKING CHANGE: Removed `NormalizeGuppy` alias for `Normalize` pass.
BREAKING CHANGE: Removed legacy `QSystemOp` re-exports in tket-qsystem.
BREAKING CHANGE: Removed `CommandsIterator`

Not a breaking python change:
BEGIN_COMMIT_OVERRIDE
chore: Remove deprecated definitions in rust library (#1970)
END_COMMIT_OVERRIDE


tket/src/serialize/pytket/encoder.rs
```diff
@@ -221,17 +221,9 @@ impl<H: HugrView> PytketEncoderContext<H> {
         .filter(|name| !name.is_empty())
         .cloned();
 
-        // Backwards compatibility only: The global phase parameter
-        // of pytket circuits is now encoded as explicit `tket.global_phase` operations instead.
-        #[expect(deprecated)]
-        let phase = match hugr.get_metadata::<metadata::PytketPhaseExpr>(region) {
-            Some(p) => p.to_string(),
-            None => "0".to_string(),
-        };
```
maximilianruesch, line 230: Since this is encoded into the Python package, is removing the read for this metadata not breaking to the Python package in a bad way?

aborgna-q: We haven't been writing to this this metadata entry in a while.
We moved to exporting standalone `GlobalPhase` ops, but left the metadata read path in case someone had a pytket circuit encoded as a Hugr flying around that needed to be re-exported.

Those hugrs are are unlikely to be be stored, so we should be fine removing this now.

maximilianruesch (approved):

PR #1975, opened 2026-09-04
feat!: Remove deprecated definitions from tket-py

aborgna-q:
drive-by: Fix some incorrect doc examples.
drive-by: Added missing `ExpectedQubitsHint` in `tket.metadata.__all__`.

BREAKING CHANGE: Removed deprecated `tket.passes.NormalizeGuppy`. Use `Normalize` instead.
BREAKING CHANGE: Removed deprecated `tket.metadata.PytketPhaseExpr`.
BREAKING CHANGE: Renamed `tket.passes._QSystemLLVMPass` to `tket.passes.QSystemLLVMPass`.

### DIFF
--- a/tket-py/tket/passes/__init__.py
+++ b/tket-py/tket/passes/__init__.py
@@ -15,7 +15,6 @@ from hugr.passes.composable import (
     implement_pass_run,
 )
 from hugr.passes.scope import GlobalScope, PassScope
-from typing_extensions import deprecated
 
 from tket import _state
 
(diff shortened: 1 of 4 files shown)


tket-py/tket/passes/__init__.py
```diff
@@ -506,7 +499,7 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
 
 
 @dataclass(kw_only=True)
-class _QSystemLLVMPass(ComposablePass):
+class QSystemLLVMPass(ComposablePass):
```
maximilianruesch: Stabilising something is not part of removing deprecated definitions, what is going on here?

aborgna-q: It was part of the breaking cleanup, just to align it with the other qsystem pass.
We don't really expect users to call this from python, so may as well revert the change.

maximilianruesch (approved):

PR #1976, opened 2026-09-04
fix: Include required newer extensions when loading a `CompilationState` from a Python hugr

aborgna-q:
Check that we actually have a compatible extension on the rust side before eliding it from the wire package encoded from a Python hugr.

This includes both `tket`-specific extensions and `std` ones.


tket-py/tket/_state/__init__.py
```diff
@@ -84,17 +84,18 @@ def from_tket1(
     def from_python(hugr: Hugr | Package) -> CompilationState:
         """Convert a python-backed Hugr to a CompilationState."""
         py_extensions = None
-        # Get extensions used by this hugr that are not already in the Rust registry.
+        # Get extensions used by this hugr that cannot be satisfied by the Rust
+        # registry at a compatible version.
         if isinstance(hugr, Hugr):
-            embedded = set(_state.embedded_extensions())
```
maximilianruesch, line 89: Is `embedded_extensions` used anywhere else, or can it be removed?

maximilianruesch (approved):

tket-py/tket/_state/__init__.py
aborgna-q: We use it to check that `tket_exts` is kept in sync with the rust registry. I removed `embedded_extensions` from `_state.__all__`, and also made it check the actual versions in the test.

PR #1988, opened 2026-09-08
chore: Cleanup qis-compiler file organization

aborgna-q:
This PR is purely a cleanup change, splitting the big `lib.rs` file from `qis-compiler` into submodules.


doug-q (approved):
thanks

PR #1996, opened 2026-09-15
feat: Add pauli_exp extension

### LINKED ISSUE #1949: [Feature]: HUGR extension for pauli exponential ops
### Problem Statement

pg-libs et al. work on pauli gadgets/exponentials but recover them from sequences of tket.quantum ops. 

### Proposed Solution

They are ubiquitous and canonical enough we should add explicit ops that can be programmed from guppy, especially in algorithmic standard library implementations. pg-libs should then take them as input.

### Component

Rust library

### Effort Estimate

Medium - Moderate complexity

### Additional Context

_No response_

cqc-alec:
This corresponds to TKET1's [PauliExpBox](https://docs.quantinuum.com/tket/api-docs/optype.html#pytket.circuit.OpType.PauliExpBox).

Implemented with AI assistance.

Closes #1949 .


aborgna-q (approved):
Since this is a 1-to-1 equivalence with pytket's op it should be easy to add an encoder/decoder between both. Can you open an issue?

PR #1997, opened 2026-09-16
fix: pg-libs/pg-tk TKET JSON conversion bug

yao-cqc:
- fix two issues when importing tket json
- use pytket for roundtrip testing instead of hardcoded JSONs


cqc-alec:
Why the switch to using pytket for roundtrip testing? Is it related to the bugfix, or separate?

yao-cqc:
> Why the switch to using pytket for roundtrip testing? Is it related to the bugfix, or separate?

@cqc-alec These bugs were actually discovered by switching to pytket testing.

cqc-alec:
> > Why the switch to using pytket for roundtrip testing? Is it related to the bugfix, or separate?
> 
> @cqc-alec These bugs were actually discovered by switching to pytket testing.

I see. Do you know why they don't show up with JSON testing? (In principle I prefer JSON testing as it doesn't involve spawning Python, so I'm interested if there's a way we can keep that or if it's inherently not sufficient.)

yao-cqc:
> > > Why the switch to using pytket for roundtrip testing? Is it related to the bugfix, or separate?
> > 
> > 
> > @cqc-alec These bugs were actually discovered by switching to pytket testing.
> 
> I see. Do you know why they don't show up with JSON testing? (In principle I prefer JSON testing as it doesn't involve spawning Python, so I'm interested if there's a way we can keep that or if it's inherently not sufficient.)

The old hardcoded JSONs had errors in them, which is why those tests didn’t catch the bugs. Tests that call `compare_unitaries_via_tk` already spawn Python, but that function uses a long-lived worker, whereas this approach starts a new Python process each time.

I can switch back to JSON testing and make sure the data are correct if you’d prefer.

cqc-alec:
> The old hardcoded JSONs had errors in them, which is why those tests didn’t catch the bugs.

Gotcha.
 
> I can switch back to JSON testing and make sure the data are correct if you’d prefer.

Yes please, I would prefer that!

aborgna-q:
Could you use [`tket_json_rs`](https://docs.rs/tket-json-rs/latest/tket_json_rs/index.html) instead of bare json?

yao-cqc:
> Could you use [`tket_json_rs`](https://docs.rs/tket-json-rs/latest/tket_json_rs/index.html) instead of bare json?

@aborgna-q Yes, I totally forgot about this! I should try using `SerialCircuit` as an intermediate conversion target. This seems outside the scope of this PR, though. I can create an issue for it.

cqc-alec (approved):

PR #2012, opened 2026-09-17
chore: Move tket libraries to a subdir

aborgna-q:
Part of #2010: Moves the group of tket libraries into its own subdirectory.

This is just to keep the repo tidy, while we integrate `pg-libs` into the workspace.


DEVELOPMENT.md
```diff
@@ -113,7 +113,8 @@ just test-rust
 just test-python
 ```
 
-Run `just` to see all available commands.
+The `tket` and `pg-libs` have independent justfiles with their own sets of commands.
```
cqc-alec:
```suggestion
The `tket` and `pg-libs` directories have independent justfiles with their own sets of commands.
```

cqc-alec (approved):

PR #2020, opened 2026-09-21
feat: Greedy resynth pass

chloe-quantinuum:
Implements a tket2 pass that takes flat HUGR regions, converts them to pauli graphs, applies the greedy resynth pass, and converts the resulting pauli graph back into a HUGR.

Note: currently the result of greedy resynth does not simplify single qubit gate sequences so it might be useful to add a simple single qubit gate merging pass after resynthesis in the future (`RotationMergingPass` is only for pauli graphs in canonical form)


tket/tket-rs/src/passes/pg_convert.rs
```diff
@@ -0,0 +1,1049 @@
...
+                data: GateData::new(GateType::RZ, vec![qubits[1]]).with_params(vec![-0.25]),
+            },
+            Op::Gate {
+                data: GateData::new(GateType::ZX, vec![qubits[0], qubits[1]]),
+            },
+        ]),
+        OpType::SWAP => Ok(vec![Op::Gate {
+            data: GateData::new(GateType::SWAP, qubits),
+        }]),
+        OpType::Measure => Ok(vec![Op::Gate {
```
sjdilkes, line 257: I don't see Reset in this list - is it supported implicitly in something else? Or is it unsupported in pg-libs? If unsupported, please add explicit handling saying it's not supported

tket/tket-py/tket/passes/__init__.py
```diff
@@ -505,6 +506,51 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
...
+    Converts the circuit into a Pauli Graph and uses GreedyResynth to synthesize the pauli graph.
+    Parameters:
+    - ancilla_budget: The number of ancilla qubits to use.
+    """
+
+    window_size: int | None = None
+    pool_size: int | None = None
+    top_up_size: int | None = None
+    seed: int | None = None
+    parallel_mode: str = "auto"
```
sjdilkes, line 521: Note that the handling for these strings in `greedy_resynth.rs` has capitalised Auto, On, Off while this default str is lowercase auto. I think users are more likely to go lowercase, so I'd change the handling to do on/off instead - though also is there some ENUM esque thing we can use instead of a string?

tket/tket-py/src/passes.rs
```diff
@@ -64,6 +65,12 @@ create_py_exception!(
     "Errors from the modifer resolver pass."
 );
 
+create_py_exception!(
```
sjdilkes, line 68: This error isn't exported so users can't import error (like with TK1PassError), which might make error messages harder to parse

tket/tket-py/tket/passes/__init__.py
```diff
@@ -505,6 +506,51 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
         return program
 
 
+@dataclass
+class GreedyResynthPass(ComposablePass):
+    """Resynthesise a circuit using greedy resynth.
+    Converts the circuit into a Pauli Graph and uses GreedyResynth to synthesize the pauli graph.
+    Parameters:
+    - ancilla_budget: The number of ancilla qubits to use.
```
sjdilkes, line 514: We don't need the ancilla_budget here I think (that's for the layer T-count pass right?)

tket/tket-py/tket/_tket/passes.pyi
```diff
@@ -102,6 +102,20 @@ def tket1_pass(
...
+    scope: PassScope = GlobalScope.PRESERVE_PUBLIC,
+    window_size: int | None = None,
+    pool_size: int | None = None,
+    top_up_size: int | None = None,
+    seed: int | None = None,
+    parallel_mode: str = "auto",
+) -> None:
+    """Applies GreedyResynth to optimize a hugr.
+
+    Parameters:
```
sjdilkes, line 116: this is empty

tket/tket-rs/src/passes/greedy_resynth.rs
```diff
@@ -0,0 +1,327 @@
...
+use hugr::hugr::ValidationError;
+use hugr::{Hugr, Node};
+use pg_canonical_form::CanonicalFormPass;
+use pg_core::{GateType, PGPass};
+use pg_greedy_synth::{GreedySynthPass, ParallelMode};
+use pg_optimise::{GroupCommutingOpsPass, RotationMergingPass};
+use pg_rebase::RebaseTQEToZXPass;
+use std::sync::Arc;
+
+// - `window_size` (`Option<usize>`) - Size of the sliding window for lookahead during synthesis. Default to 1280.
```
sjdilkes, line 28: This needs to be /// not //

tket/tket-rs/src/passes/greedy_resynth.rs
```diff
@@ -0,0 +1,327 @@
...
+    /// Defaults to `max(1000, 0.2 * N^2)` where `N` is the number of qubits.
+    pub fn with_pool_size(mut self, pool_size: usize) -> Self {
+        self.pool_size = Some(pool_size);
+        self
+    }
+
+    /// Sets the number of candidate gates to add after each TQE gate.
+    ///
+    /// Defaults to `max(200, pool_size / N)` where `N` is the number of qubits.
+    pub fn with_top_up_size(mut self, top_op_size: usize) -> Self {
```
sjdilkes, line 83: nit: the parameter is called top_op_size but I think for consistency it should be top_up_size?

sjdilkes (changes requested):
Nice! A few things I've spotted.

tket/tket-py/src/passes.rs
```diff
@@ -230,3 +237,46 @@ fn resolve_modifiers(circ: &mut CompilationState, scope: Option<PyPassScope>) ->
     pass.run(&mut circ.hugr).convert_pyerrs()?;
     Ok(())
 }
+
+fn parse_parallel_mode(s: &str) -> Option<pg_greedy_synth::ParallelMode> {
```
cqc-alec, line 241: It would be better to pass an enum rather than a string.

tket/tket-py/tket/passes/__init__.py
```diff
@@ -505,6 +506,51 @@ def _run_tk(self, program: _state.CompilationState) -> _state.CompilationState:
         return program
 
 
+@dataclass
+class GreedyResynthPass(ComposablePass):
+    """Resynthesise a circuit using greedy resynth.
+    Converts the circuit into a Pauli Graph and uses GreedyResynth to synthesize the pauli graph.
```
cqc-alec, line 512: I don't think this docstring is sufficient. It should describe what `GreedyResynth` does (or link to a description).

tket/tket-rs/src/passes/greedy_resynth.rs
```diff
@@ -0,0 +1,327 @@
+//! Greedy resynthesis of a Clifford + T circuit through a Pauli graph.
+//!
+//! The [`GreedyResynthPass`] resynthesises a circuit by converting it to a
+//! Pauli graph, applying the [`GreedySynthPass`], and converting the result
```
cqc-alec, line 4: This docstring doesn't really tell me anything. It should describe what `GreedySynthPass` does.

tket/tket-rs/src/passes/pg_convert.rs
```diff
@@ -0,0 +1,1049 @@
+//! Conversion utilities between serial circuits and Pauli graphs.
+//!
+//! Provides [`RegisterMap`], [`serial_circuit_to_pauli_graph`] and
+//! [`pauli_graph_to_cmds`] utilities required for passes that resynthesise
```
cqc-alec, line 4:
```suggestion
//! [`pauli_graph_to_cmds`] utilities required for passes that resynthesize
```

tket/tket-rs/src/passes/greedy_resynth.rs
```diff
@@ -0,0 +1,327 @@
...
+                synth_pass = synth_pass.with_window_size(ws);
+            }
+
+            if let Some(ps) = self.pool_size {
+                synth_pass = synth_pass.with_pool_size(ps);
+            }
+
+            if let Some(ts) = self.top_up_size {
+                synth_pass = synth_pass.with_top_up_size(ts);
+            }
```
cqc-alec, line 148: Should check that these values are non-negative and return an error if not.

tket/tket-py/tket/_tket/passes.pyi
cqc-alec, line 1: Please add some tests for the Python wrapper.

tket/tket-rs/src/passes/pg_convert.rs
```diff
@@ -0,0 +1,1049 @@
...
+                GateType::Measure => {
+                    let qubit = register_map.get_qubit_id(args[0])?;
+                    let bit = register_map.get_bit_id(args[1])?;
+                    Ok(vec![Command {
+                        op: Operation::from_optype(OpType::Measure),
+                        args: vec![qubit, bit],
+                        opgroup: None,
+                    }])
+                }
+                _ => Err(ConversionError::UnsupportedGate(
```
cqc-alec, line 558: Note that `GateType::BlackBox` is not covered here. However, `Op::BlackBox` is handled below. I think the two are equivalent and we should combine them. https://github.com/Quantinuum/tket2/issues/2028

PR #2021, opened 2026-09-21
feat(pytket): parse pi as a constant

### LINKED ISSUE #1900: [Feature]: Explicitly handle "pi" when parsing pytket parameters
### Problem Statement

We currently parse pytket parameters using a short pest definition that results in the recursive `PytketParam` enum.
When we find a word terminal, we wrap represent it as a `PytketParam::InputVariable`.

The issue is that pytket treats `pi` as a special case. That is not an input variable but a shorthand for the number.
We should add that explicitly to the parser, rather than manually handling it downstream.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

akshaysoftware:
Fixes #1900 

`pi` was previously parsed as a `PytketParam::InputVariable` and then handled specially in downstream decoder code. This change parses `pi` directly as `PytketParam::Constant(std::f64::consts::PI)`, so it follows the existing constant-loading path instead. I have also removed the special case code in the decoder.


aborgna-q (approved):
Awesome. Thanks!

akshaysoftware:
Hi @aborgna-q it looks like this was removed from the merge queue due to a `ruff` failure on `test_files/modifier_examples/loaded_pytket.py` which I think is present on `main`. Could you re-add this to the queue once `main` is green?

aborgna-q:
Yes, I fixed that issue already so we should be able to retry now

PR #2023, opened 2026-09-22
feat: add pg-libs facade crate

yao-cqc:
Adds pg-libs as a central crate for the Pauli graph libraries, with passes grouped under passes and the crate READMEs included in the API docs.
Could you both take a look and let me know what you think? @cqc-alec @aborgna-q


pg-libs/pg-core/README.md
```diff
@@ -10,6 +10,6 @@ This crate provides:
 
 ## License
 
-This project is licensed under Apache License, Version 2.0 ([LICENSE][] or http://www.apache.org/licenses/LICENSE-2.0).
+This project is licensed under Apache License, Version 2.0 ([LICENSE][] or <http://www.apache.org/licenses/LICENSE-2.0>).
```
cqc-alec, line 13: Should we add this license notice to the other crates' READMEs as well?

pg-libs/pg-converters/pg-tk/src/lib.rs
```diff
@@ -1,5 +1,4 @@
-//! This crate provides conversion functions between `PauliGraph` and serialized TKET circuit format
-//! as well as a function to compare the unitaries of two `PauliGraph`s by calling `pytket`.
+#![doc = include_str!("../README.md")]
```
aborgna-q, line 1: I would recommend against re-using the README for the crate docs.

The README is shown on the landing page of `crates.io`, and normally has more general info on what the crate is about and its usage, installation instructions, links to the docs and repo, badge images, and licence info.

The module doc is geared more towards actual usage, with links to the API etc.

pg-libs/pg-libs/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+repository = { workspace = true }
+description = "A toolkit for constructing, transforming, and synthesizing Pauli graphs"
+readme = "README.md"
+documentation = "https://docs.rs/pg-libs"
+keywords = ["Quantum", "Quantinuum"]
+categories = ["compilers", "science::quantum-computing"]
+
+[features]
+default = []
+simd = ["pg-bitpacked/simd", "pg-qm-tableau/simd", "pg-greedy-synth/simd"]
```
aborgna-q, line 17: Before we set on this. Would it be possible to rename the feature to `unstable_simd`, or pair it with a `nightly` feature (if we plan to offer target-specific stable simd)?

A nightly-only feature is hardly stable.

yao-cqc: Sure, I can rename them to `unstable_simd`

pg-libs/pg-converters/pg-tk/src/lib.rs
yao-cqc: Right, that was the part I was most unsure about. Let's use module docs.

pg-libs/tk-pg-libs/src/lib.rs
```diff
@@ -0,0 +1,48 @@
+#![doc = include_str!("../README.md")]
```
cqc-alec: Is it intentional to leave this as an `include_str`?

cqc-alec (approved):

aborgna-q (approved):
LGTM, aside from Alec's comment

PR #2024, opened 2026-09-22
chore: Unify the pg-libs and tket workspaces

### LINKED ISSUE #2010: [Chore]: Move pg-libs to the root rust workspace

aborgna-q:
Closes #2010.

Things should continue working as usual.

The main changes are to the `just` commands. Running `just test` / `just check` / etc. from the root will run things for both sub-projects. Use `just tket/test` or `just pg-libs/check` to run focused commands.


DEVELOPMENT.md
```diff
@@ -102,20 +113,31 @@ For this you must have `maturin` installed (you can install it with `pip install
...
+just tket/test
+just pg-libs/test
 ```
 
 The `tket` and `pg-libs` directories have independent justfiles with their own sets of commands.
 Run `just tket/help` and `just pg-libs/help` to see all commands.
 
+Pg-libs tests use stable features and include the Python worker integration
+tests. To test its nightly-only SIMD features, run
+`just pg-libs/test-rust-nightly`.
```
cqc-alec, line 139: Note that this fails three tests with the error message: "Install `pytket` in the environment, or set `PG_TK_PYTHON` to a Python interpreter that has it.". Can we include some guidance on how to fix this? I tried setting `PG_TK_PYTHON` to a python interpreter that has pytket installed but got the same errors. @yao-cqc any ideas? (Edit: This applies to both nightly and non-nightly runs.)

cqc-alec (approved):

DEVELOPMENT.md
aborgna-q: Uhm, I don't see the error locally.

The just recipe is doing `uv run --only-group pg-libs cargo +nightly run ...`, and the `pg-libs` group has a `pytket` dependency.

cqc-alec: Weird. Just running
```
uv run --only-group pg-libs cargo test -p pg-bitpacked
```
gives me the three failures.

cqc-alec: But I checked that
```
uv run python -c "import pytket"
```
succeeds...

yao-cqc (approved):
LGTM

PR #2025, opened 2026-09-22
fix: Avoid naming different modified functions the same

nicolaassolini-qntm:
E.g. given `my_func`, the modified version under dagger is: `__modified__my_func[D]`; the modified version under 3 controller (packed as 2,1) and dagger is `__modified__my_func[2.1D]`


tket/tket-rs/src/modifier.rs
```diff
@@ -52,4 +53,16 @@ impl CombinedModifier {
         }
         Ok(())
     }
+
+    fn compact_string(&self) -> String {
```
aborgna-q, line 62: Docs!!!

aborgna-q (approved):

